PR #4, opened 2022-09-09
Daily Data Providers

chrisnowak:
## Overview

This is an investment I made for Symptom Shark - some of the new functionality there I want to contribute to MyDataHelpsUI. 

A common use case for querying device data, especially for the kinds of things that MyDataHelps tends to be used for, is in attempting to summarize the data by day.  

For instance, here is the new "Activity" view in Symptom Shark:
image

However, this is easier said than done, because:
- The user's timezone may shift, meaning a concept like "daily steps" shifts
- Apple Health data is recorded with a full timestamp (including offset), but Fitbit data does not come with an offset - dates/times are all in user local time (this is just the nature of the data)
- The interesting data to extract may be embedded in sub-properties of data points
- Sleep data on Apple Health comes in as time periods which may overlap
- Certain data types make sense to average (for instance, Resting Heart Rates from Apple Health) whereas some make sense to "max" (Apple Health Max Heart Rate) and some make sense to sum (Hourly Steps) to get a single daily value

This creates a new layer for querying daily device data types which abstracts this ^.  There is a function, "queryDailyData", which you can pass a "daily data type", startdate, and enddate and it will return a (promise for a) lookup of values by day.  A daily data type might be "AppleHealthSleep" or "FitbitRestingHeartRate".  It is not necessarily 1:1 with a DeviceDataPointType.

Default daily data types are listed in "query-daily-data.ts".  They can be extended by downstream applications by calling "registerDailyDataType".  Each daily data type has a data provider, and there is also a "checkDailyDataAvailability" function that can be invoked to determine if the data exists for the participant.  For instance, you may not want to show a graph of heart rates if the participant does not have any heart rate data.

The DeviceDataMonthChart component has been refactored and simplified to take advantage of this new functionality.  In particular, there were a lot of options on the DeviceDataMonthChart that could be inferred from the data type (like whether to ignore the offset).

I plan to bring in at least the major parts of the Symptom Shark "activity" view as well, which would use this.

## Security

### Testing
- [X] MyDataHelps iOS app
- [X] MyDataHelps Android app

### Documentation
- [X] If this feature requires a developer doc update, tag @CareEvolution/api-docs.

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/helpers/query-daily-data.ts
```diff
@@ -0,0 +1,93 @@
...
+registerDailyDataProvider("FitbitRestingHeartRate", fitbitRestingHeartRateDataProvider, simpleAvailabilityCheck("Fitbit", ["RestingHeartRate"]));
+registerDailyDataProvider("FitbitSleepMinutes", fitbitTotalSleepMinutesDataProvider, simpleAvailabilityCheck("Fitbit", ["SleepLevelRem", "SleepLevelLight", "SleepLevelDeep", "SleepLevelAsleep"]));
+registerDailyDataProvider("FitbitLightSleepMinutes", fitbitLightSleepMinutesDataProvider, simpleAvailabilityCheck("Fitbit", ["SleepLevelLight"]));
+registerDailyDataProvider("FitbitRemSleepMinutes", fitbitRemSleepMinutesDataProvider, simpleAvailabilityCheck("Fitbit", ["SleepLevelRem"]));
+registerDailyDataProvider("FitbitDeepSleepMinutes", fitbitDeepSleepMinutesDataProvider, simpleAvailabilityCheck("Fitbit", ["SleepLevelDeep"]));
+registerDailyDataProvider("FitbitSpO2", fitbitSpO2DataProvider, simpleAvailabilityCheck("Fitbit", ["SpO2"]));
+registerDailyDataProvider("FitbitSteps", fitbitStepsDataProvider, simpleAvailabilityCheck("Fitbit", ["Steps"]));
+registerDailyDataProvider("GoogleFitSteps", googleFitStepsDataProvider, simpleAvailabilityCheck("GoogleFit", ["Steps"]));
```
ajmenca, line 81: All of the strings here are the "magic" strings that an end user needs to know in order to pass them to the `dailyDataType` parameter of the `DeviceDataMonthChart`.  If you were to just be looking at the storybook, though, is there any way to know what the valid options are?

Also, is there any reason that these can't be an enumeration to add a bit more type safety?

chrisnowak: Hmm, yeah, that caused issues with the dictionary because you can't easily have an enum | string as a dictionary key, but it does seem like something we could do on the method call at least.

The issue is I wanted to be able to extend this downstream...if there is a new device data type or new way to calculate something that comes up...OR if we want to add some study-specific thing (daily mood score etc.) I ideally do not want to *have* to add to MDHUI first

ajmenca: Ah, I see - if someone had a specific way they wanted to handle data, they can currently register their own daily calculator with an arbitrary key and pass that same key to the `DeviceDataMonthChart`.  I certainly see the value in that, even if it feels less "typescript-y".  

So then really my main concern was people not knowing *where* to find these keys for the ones we're supporting out of the box here.  Maybe we could add them as an export in this module?  

```
export const DailyDataTypes = { 
  AppleHealthDistanceWalkingRunning: "AppleHealthDistanceWalkingRunning", 
  AppleHealthFlightsClimbed: "AppleHealthFlightsClimbed", 
  AppleHealthHeartRateRange: "AppleHealthHeartRateRange"...
  // etc
} as const;
```

Its a little verbose, but it at least sort of "publishes" the things we support?

chrisnowak: Yep, agree something like that seems necessary

chrisnowak: Typescript actually supports enums, so I added one for all the daily data types.

PR #10, opened 2022-11-09
Updates for better accessibility

chrisnowak:
## Overview

These are some updates to MyDataHelpsUI to better support 501 compliance and accessibility.

We relied on a fair number of clickable div elements which are not accessible.  This replaces the majority of them with ```<button>``` elements as that seems to be consistent with rule 1 of ARIA use: https://www.w3.org/TR/aria-in-html/#rule1

This enables you to tab and use the keyboard on most elements on the page without needing to use "tabindex" - additionally, if we were to just add "tabindex" to our clickable divs it would require special code to handle the space/enter key.  So replacing with button elements was the easiest path.

This isn't a breaking change unless styles have been overridden downstream.

There are a few other miscellaneous tweaks here as well like adding titles in a few places to input and icon buttons.

## Security

There are no new XSS threats or other risks this would introduce as it is just a change to what elements are being rendreed.

### Testing
- [X] MyDataHelps iOS app
- [X] MyDataHelps Android app

### Documentation
- [X] I have added relevant Storybook updates to this PR as well. (N/A)

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/container/ConnectEhr/ConnectEhr.css
+++ b/src/components/container/ConnectEhr/ConnectEhr.css
@@ -1,6 +1,13 @@
-﻿.mdhui-connect-ehr {
+﻿.mdhui-connect-ehr
+{
     background: #FFF;
     position: relative;
+    width: 100%;
+    text-align: left;
+    border: none;
+    font-size: 1em;
+    line-height: 1.2em;
+    cursor:pointer;
 }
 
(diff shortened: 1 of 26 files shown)


vladtar (approved):
Changes look good to me.

PR #13, opened 2022-12-02
More granular support for limiting ext account widgets and views by provider category.

greinard:
## Overview

This branch updates the EHR connection management widgets such that we can have more granular control over which categories of provider are filtered out.  For the Cue study, we have a need to limit the EHR connection management to Health Plans only.

Previously, a couple of widgets supported excluding device manufacturers while others could be provided with a list of categories.  All the widgets now support the ability to have any combination of Providers, Health Plans, and/or Device Manufacturers.

I believe these changes are all backward compatible with one exception that I will comment on.  I believe that exception was due to a bug that is now fixed.

The diff contains a fair amount of tabs vs. spaces formatting, so I would suggest viewing the diff while ignoring whitespace (add `w=1`) to the URL.

## Security

No security risks have been introduced.  Clients of this library can now further limit the list of external account providers that show up in searches and/or that are present in connection lists and previews.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

The best way to test these changes outside of the apps is with the storybook (`npm run storybook`).  I have added preview states to capture all of the possible filtering scenarios.

### Documentation


src/components/view/ExternalAccountsView/ExternalAccountsView.tsx
```diff
@@ -1,42 +1,64 @@
...
+        <Layout>
+            {props.presentation &&
+            <NavigationBar title={title}
+                           showBackButton={props.presentation == "Push"}
+                           showCloseButton={props.presentation == "Modal"}/>
+            }
+            {!props.presentation &&
+            <StatusBarBackground color="var(--main-bg-color)"/>
+            }
+            <ExternalAccountList previewState={props.preview ? "Default" : undefined} externalAccountProviderCategories={externalAccountProviderCategories} onExternalAccountsLoaded={onExternalAccountsLoaded}/>
```
greinard, line 61: This is the only change that I think is not backward compatible.

Despite having the `excludeDeviceManufacturers`, which could be `false`, the external account list was never being passed the `Device Manufacturer` category, which I think was a bug.

Now the `Device Manufacturer` category will be passed to the external account list if `excludeDeviceManufacturers` is `false`.

src/helpers/strings-es.ts
```diff
@@ -1,56 +1,61 @@
...
+    "distance-traveled": "Distancia Viajada",
+    "google-fit-share": "Toca para compartir datos de condición físic",
+    "devices": "Dispositivos",
+    "last-sync": "Última sincronización",
+    "completed": "Completadas",
+    "ehr-intro": "Sus registros de salud electrónicos son fuentes importantes de información. Pueden ayudar a investigadores hacer nuevos descubrimientos.",
+    "external-accounts-title-divider": " / ",
+    "external-accounts-title-providers": "Proveedores",
+    "external-accounts-title-health-plans": "Planes de Salud",
+    "external-accounts-title-devices": "Dispositivos",
```
greinard, line 46: I wasn't sure what the capitalization rules should be here for `external-accounts-title-health-plans` and `external-accounts-title-devices`.  In the full title, they were not capitalized, but since they can appear by themselves now, I thought they should be.  If there is an issue with having them be capitalized when they might be included as the second or third heading, let me know and I will revisit this.

src/components/container/ExternalAccountsPreview/ExternalAccountsPreview.tsx
```diff
@@ -1,65 +1,91 @@
...
-import { previewAccounts } from './ExternalAccountsPreview.previewdata'
+import {previewAccounts} from './ExternalAccountsPreview.previewdata'
 
 export interface ExternalAccountsPreviewProps {
-	excludeDeviceManufacturers?: boolean
-	applicationUrl: ExternalAccountsApplicationUrl,
-	previewState?: ExternalAccountsPreviewPreviewState
+    excludeProviders?: boolean;
+    excludeHealthPlans?: boolean;
+    excludeDeviceManufacturers?: true;
```
ajmenca, line 11: This looks weird to me?  I would have expected this to also just say `boolean` (because I don't think you can set default values for an interface?) - and also you seem to be checking the truthiness of this below https://github.com/CareEvolution/MyDataHelpsUI/pull/13/files?diff=unified&w=1#diff-c1bad721c4ed87fb7ed25796eb12273532aca486dc0f39e4a124bf4e3d4256e9R36

src/components/view/ExternalAccountsView/ExternalAccountsView.tsx
ajmenca: If this is bug fix, why do you say it's not "backward compatible?"  To me, saying it's not backward compatible implies that end users will have to adjust code for things to work - but in this case if they were trying to use the flag it wouldn't have been working and now it *will* start working without any changes, right?

src/components/container/ExternalAccountsPreview/ExternalAccountsPreview.tsx
greinard: Good catch.  That's a typo.

src/components/view/ExternalAccountsView/ExternalAccountsView.tsx
greinard: I was just saying that it is not backward compatible because it *will* require end users to adjust code if they want to retain the existing behavior.  That is to say, by default, it will now *include* device manufacturers, when previously it did not.

If an end user wants to have the "correct" behavior, given their current config, then nothing needs to be done.

src/components/container/ExternalAccountsPreview/ExternalAccountsPreview.tsx
greinard: Fixed.

greinard:
Going to merge this now, but please continue to comment if there are any further requests for change and I will address them in a follow up PR.

PR #17, opened 2023-01-30
Add Survey Step Components and YouTube Step

ashleyrtate:
## Overview

This adds support for external Web step components and the initial YouTube Web step implementation as a working/useful POC. 

The components include:

- StepMarkdown - supports markdown in other components specifically tailored to use in a survey step (ensuring links open in a new window, for example). This implementation will support the full CommonMark spec implemented in markdown-it and is not suppressed as the implementation within MyDataHelps native Web components. 
- StepTitle - Displays styled step.title with markdown support
- StepText - Displays styled step.text with markdown support
- StepDetailText - Displays step.detailText with markdown support
- StepNextButton - Displays styled "Next" button with support for step.nextButtonText override and default support for English/Spanish localization.
- YouTubeStep - Displays YouTube video with title, text, transcript, and configurable height in pixels. 
  - Video width bleeds to edge of step. 
  - CC default language is matched to MDH language. 
  - Related videos are suppressed to current video channel (fully suppressing all related videos is no longer supported by YouTube)
  - YouTube branding is minimized
  - Autoplay is off
  - Fullscreen is allowed
- ExampleStep - Used as a storybook wrapper around lower level components for correct styling

The building block components will support building new steps that are functionally and stylistically equivalent to existing native survey steps.

### Customized YouTube Step from Storybook
image

## Security

This is primarily a port of survey step functionality and api calls that have already been reviewed and vetted elsewhere. The main risk is displaying unsanitized step content that would allow CSS attacks. This is mitigated by used of UI components that sanitize or prevent user-generated html content by default. Specifically, the markdown-it library used to render all text and support markdown is the same library already used within existing MyDataHelps web components.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

@chrisnowak, @ajmenca, @explunit


chrisnowak:
it looks like the package-lock had way more changes than expected, but unfortunately seems like I can't review on github...did you delete and do a fresh install or something?  Or maybe running old npm?  Might want to revert it and just npm install the packages you are looking for to get the minimum set of changes needed

src/components/step/ExampleStep/ExampleStep.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import '../step.css';
+
+export interface ExampleStepProps {
+    children?: React.ReactNode;
+}
+
+export default function (props: ExampleStepProps) {
```
chrisnowak, line 8: So I think the main goal here seems to be to include the main CSS file for storybook purposes?

I would instead follow the pattern with the other components in this library where I actually have a ```<Layout>``` component that embeds the "global" CSS stuff.  Here I might call it ```<StepLayout>```

Then your YoutubeStep would actually have ```<StepLayout>``` as it's top level dev (and all future steps would) and we don't have a component that's used only for storybook purposes

src/components/step/StepDetailText/StepDetailText.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { StepElementProps } from '../shared';
+import StepMarkdown from '../StepMarkdown/StepMarkdown';
+
+
+export default function (props: StepElementProps) {
+    if (!props.text) {
+      return (<></>)
+    }
+    return (
+      <div id="detail-text" className="step-detail-text" style={stepElementStyle(props)}>
```
chrisnowak, line 13: Does this need an id?  Good chance it could be used multiple times on a page so would not be unique

src/components/step/StepMarkdown/StepMarkdown.tsx
```diff
@@ -0,0 +1,34 @@
+import React, {useEffect} from 'react'
```
chrisnowak, line 1: Minor, don't see use of useEffect

src/components/step/YouTubeStep/YouTubeStep.tsx
```diff
@@ -0,0 +1,83 @@
...
+    // ref player params at https://developers.google.com/youtube/player_parameters
+    let frameHeight = (properties.height ? properties.height : "225") + "px";
+    let ccLanguage = MyDataHelps.getCurrentLanguage(); // used to set default CC language
+    let showRelated = 0; // restricts related videos to current channel only
+    let showControls = 1; // show player controls, probably something we should make configurable
+    let autoplay = 0; // do not autoplay, probably should make this configurable
+    let allowFullscreen = 1; // show fullscreen controls
+    let youTubeBranding = 1; // hide youtube logo
+    return (
+      <div className="step-container">
```
chrisnowak, line 39: (This is where I think this could be a ``<StepLayout>```)

src/components/step/YouTubeStep/YouTubeStep.tsx
```diff
@@ -0,0 +1,83 @@
+import React, { useState, useEffect } from 'react'
+import MyDataHelps, {StepConfiguration} from '@careevolution/mydatahelps-js';
+import StepTitle from '../StepTitle';
+import StepText from '../StepText';
+import StepDetailText from '../StepDetailText'
+import StepNextButton from '../StepNextButton';
+import '../step.css';
+
+export interface YouTubeStepProps {
+    properties?: any;
```
chrisnowak, line 10: I assume these props are entirely for testing...but from what I can see won't actually update the step if you change them (they just set the default value) which might be unexpected for someone using this library.

One thing that I haven't done a great job at (mostly because I learned about the pattern halfway through building this repo) is to separate out presentational vs. container components.  

Given that we want to make things reusable with our outside collaborators, I think it would be worth it to split that out here such that YoutubeStep is *purely* presentational (zero state, just props) and then we have a YoutubeStepContainer component that has the state/useEffect call.

src/components/step/YouTubeStep/YouTubeStep.tsx
```diff
@@ -0,0 +1,83 @@
...
+    styles?: any;
+}
+
+export default function (props: YouTubeStepProps) {
+
+    const [properties, setProperties] = useState<any>(props.properties ?? {});
```
chrisnowak, line 16: Aren't props/styles typed a bit more specifically than any?  Like they are ``` { [key: string]: any }```

chrisnowak:
(before I go further - want to note that this is looking awesome)

src/components/step/YouTubeStep/YouTubeStep.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+    styles: {}
+}
+
+export const YouTubeStepCustomStyling = Template.bind({});
+YouTubeStepCustomStyling.args = {
+    properties: {
+        "title": "_MyDataHelps™_",
+        "text": "Meet Participants Where They **Are**",
+        "nextButtonText": "Forward",
+        "transcript": "Meet participants where they are with _MyDataHelps™_—your one-stop digital health platform for conducting clinical research, clinical trials, and wellness projects! eConsent and enroll participants, collect digital biomarkers, automate survey delivery, and begin analyzing digital phenotype and EHR data in hours. Read more [here](https://careevolution.com/mydatahelps/).",
```
chrisnowak, line 33: Nice 👍

src/components/step/YouTubeStep/YouTubeStep.tsx
```diff
@@ -0,0 +1,83 @@
...
+    return (
+      <div className="step-container">
+           <StepTitle 
+              text={properties.title} 
+              textAlign={styles.titleAlignment}
+              color={styles.titleColor}
+              fontSize={styles.titleFontSize}
+              fontWeight={styles.titleFontWeight}
+            />
+           <StepText
```
chrisnowak, line 47: Something weird here I can't figure out:
image

The step text appears to have some extra height on it.  Looking at chrome browser tools I can't really figure out where it's coming from though (the inner <p> element has the right height)

src/components/step/step.css
```diff
@@ -0,0 +1,54 @@
...
+    background-color: #FFFFFF;
+}
+
+a {
+    color: #5381c3;
+    text-decoration: none;
+}
+
+.step-container {
+    padding: 10px 20px 10px 20px;
```
chrisnowak, line 29: So I'd probably remove the top padding here and instead have it come from a top margin on the step title.  In particular that would enable the video to be fully anchored to the top of the screen if there is no title/text.

chrisnowak: (Removing the left/right padding here and instead adding it to the various child elements I think i'd also be slightly in favor of, so you don't have to do extra stuff to make the YoutubeStep "escape" the frame - might be more stuff like that in the future)

src/components/step/step.css
```diff
@@ -0,0 +1,54 @@
...
+    background-color: #FFFFFF;
+}
+
+@supports (font: -apple-system-body) {
+    html {
+        font: -apple-system-body;
+    }
+}
+
+body,div,dl,dt,dd,ul,ol,li,h1,h2,h3,h4,h5,h6,pre,form,fieldset,input,textarea,p,blockquote,th,td {
```
chrisnowak, line 12: We should ideally isolate this - the CSS is all bundled together, so this will override styles for all components (even not steps).  So I'd nest all the styles in this file under .step-layout if they are necessary.

The html/font stuff is already set up by https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/components/presentational/Layout/Layout.css...which probably should also be fixed to be more isolated now (actually not sure if it's possible to add a class to the <html> element, and I think it might be necessary to apply a style to ```<html>``` achieve dynamic type sizing...not sure)

For now though I'd probably just try to keep things isolated under the class as much as possible so we don't override styles for the rest of the view library downstream for instance

src/components/step/shared.ts
```diff
@@ -0,0 +1,8 @@
+
+export interface StepElementProps {
+    text: string;
+    textAlign?: any;
```
chrisnowak, line 4: 3/5 of these properties aren't used or usable by StepNextButton, no?  I guess I would lean toward separating out NextButtonProps into a separate interface.  But these styles will actually "work" if they are applied separately, right...if someone was using the StepNextButton as a component on it's own and hardcoding the styles?  (So I guess seems fine to leave it...mostly do not want any props that do nothing since that would be confusing downstream for anyone using the component)

src/components/step/step.css
```diff
@@ -0,0 +1,54 @@
+html {
```
chrisnowak, line 1: (I would have this file be attached to my proposed ```<StepLayout>```) component so it's StepLayout.css)

src/components/step/StepText/StepText.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { StepElementProps } from '../shared';
+import StepMarkdown from '../StepMarkdown/StepMarkdown';
+
+
+export default function (props: StepElementProps) {
+    if (!props.text) {
+      return (<></>)
+    }
+    return (
+      <div id="text" className="step-text" style={stepElementStyle(props)}>
```
chrisnowak, line 13: Would remove the "id" unless somehow necessary

src/components/step/StepText/StepText.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import stepElementStyle from '../step-helpers';
+import './StepText.css'
+import { StepElementProps } from '../shared';
+import StepMarkdown from '../StepMarkdown/StepMarkdown';
+
+
+export default function (props: StepElementProps) {
+    if (!props.text) {
+      return (<></>)
```
chrisnowak, line 10: I assume it's functionally equivalent (render nothing) but I usually just return null

rollup.config.js
```diff
@@ -4,12 +4,13 @@ import typescript from "@rollup/plugin-typescript";
...
 import postcss from "rollup-plugin-postcss";
 import image from '@rollup/plugin-image';
+import json from '@rollup/plugin-json'
 const packageJson = require("./package.json");
 import { terser } from "rollup-plugin-terser";
 import peerDepsExternal from 'rollup-plugin-peer-deps-external';
 import analyze from 'rollup-plugin-analyzer';
 
-const limitBytes = 1e6
+const limitBytes = 2e6
```
chrisnowak, line 13: Bundle size increased a lot?

I think it does not matter too much - I spent some time getting it down, but realized if downstream things are using this they would be tree-shaking the library so code they are not using shouldn't be included.  My understanding of such things is still a bit shaky though

src/components/step/StepMarkdown/StepMarkdown.tsx
```diff
@@ -0,0 +1,34 @@
...
+import MarkdownIt from 'markdown-it';
+import parse from 'html-react-parser'
+
+export interface StepMarkdownProps {
+  text: string;
+  inline?: boolean;
+}
+
+export default function (props: StepMarkdownProps) {
```
chrisnowak, line 10: This is definitely fine for now, but as we move toward possibly allowing editing of custom views dynamically (like steps), I can imagine making this more generic (e.g. calling it Markdown) and moving it up to the presentational components

rollup.config.js
```diff
@@ -4,12 +4,13 @@ import typescript from "@rollup/plugin-typescript";
 import dts from "rollup-plugin-dts";
 import postcss from "rollup-plugin-postcss";
 import image from '@rollup/plugin-image';
+import json from '@rollup/plugin-json'
```
chrisnowak, line 7: What's this for?  I don't see any JSON includes or exports here

rollup.config.js
ashleyrtate: I'm not sure exactly where it was. At one point the markdown stuff pushed it over the limit but I'm not sure exactly how much with the latest iteration. Both of the markdown libraries pull in a lot of stuff. I'm going to try starting fresh with the package lock file and I'll check the size again after that. I did reset the package lock a couple of times already and thought it should be clean but this went through a lot of iterations working out the downstream webpack kinks so worth double checking.

chrisnowak:
Done with my initial review - would like to take another pass after comments, but again, overall looking awesome

rollup.config.js
ashleyrtate: markdown-it includes a json file and the rollup build can't handle packing json files into the bundle without this.

rollup.config.js
chrisnowak: I think the real endpoint to check/verify is the bundle size when you visit https://viewlibrary.careevolutionapps.dev/ and https://viewlibrary.careevolutionapps.com/ after this is merged to see if it increased a bunch.  

Also I realize the package-lock.json flog is probably fine assuming it's not due to being on an old npm version or something...wouldn't worry too much about that...I probably should just trust semver and that any package updates that were picked up aren't breaking

rollup.config.js
chrisnowak: Got it, makes sense

src/components/step/shared.ts
ashleyrtate: Yes, it could be used that way to override next button styles in a different context. In this case I was mainly trying to keep the style conversions in one place for all step components.

src/components/step/step.css
chrisnowak: I *think* that literally all the global styles should be defined in https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/components/presentational/Layout/Layout.css

Those that match we can remove presumably (which is good number of them) and I don't think I see any major conflicts between html defined there vs. here

Major issue might be the background color - might look at the Layout component to see how I solved that (there is a way to explicitly set it) and just explicitly set body background color in the StepLayout to #FFFFFF

But yeah, all the CSS is bundled together so have to be careful about global styles since I imagine this is overridding stuff in the view library now

src/components/step/YouTubeStep/YouTubeStep.tsx
ashleyrtate: Yeah, it makes sense generally and I've gone ahead and split this out. Interestingly, the guy who originally popularized this seems to have semi-disavowed it, or at least encouraged people not to over apply:

https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.5hwzcvetl
https://blog.bitsrc.io/implementing-the-container-pattern-using-react-hooks-f490a8492d05

It definitely makes sense if you have any reason to use the presentational bit outside the original/larger context. I'm not sure that will be the case here with this step though. Outside of the storybook framework itself it does not seem we would use the presentational YouTubeStep without its container in the real world. So would be good to understand how much the storybook framework and making things isolated for presentational *testing* is a driver in your thinking about this.

src/components/step/YouTubeStep/YouTubeStep.tsx
ashleyrtate: This was caused by interaction between markdown and the "pre" whitespace styling. Took a while to get rid of it but basically just needed to trim what is coming back from the markdown library.

src/components/step/step.css
ashleyrtate: I've cleaned this up a bit but still a bit stymied on the html background color once you run the bundle in MDHViewLibrary. I can override the background var in the StepLayout.css like this:

```
:root
{
    --main-bg-color: #FFFFFF;
}
```

But that still ends up overriding the same var in the views so you end up with the survey white background there too. Only way I've seen that might work is if the StepLayout component actually did something like this to update the var at runtime:

`document.documentElement.style.setProperty('--main-bg-color', randomColor);`

Any other ideas?

ashleyrtate:
@chrisnowak thanks for the detailed review. This went through a few iterations as I was learning React (starting with class components and refactoring to functional and moving from stand alone into the component library). Quite a lot of churn around the markdown stuff and toolchain issues too. 

Also ran into this problem if you upgrade @rollup/plugin-typescript past 8.3.3 so I've pinned that version and library in package.json:
https://stackoverflow.com/questions/73197460/error-could-not-resolve-entry-module-react-rollup

Except for the background color styling issue noted above it should be in good shape for another look.

src/components/step/step.css
ashleyrtate: The second bit of that may not be working as intended:
`document.body.style.removeProperty('background-color');`

Seems like it may affect only inline styles.

chrisnowak: Gotcha, I think I probably never tested that since I'm not really removing the background color typically once the page is rendered...probably should be a default there anyways or something

src/components/step/YouTubeStep/YouTubeStep.tsx
ashleyrtate: I already split it out, was not a big deal in this case. I just wanted to have a clear rationale in my head for when/why we would implement this pattern or exceptions to it.

chrisnowak:
looks like package-lock has been deleted on this branch now...probably will need to install / re-commit that before merging

src/components/step/YouTubeStep/YouTubeStep.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react'
+import MyDataHelps from '@careevolution/mydatahelps-js';
+import StepLayout from '../StepLayout'
+import StepTitle from '../StepTitle';
+import StepText from '../StepText';
+import StepDetailText from '../StepDetailText'
+import StepNextButton from '../StepNextButton';
+
+export interface YouTubeStepProps {
+    properties: { [key: string]: any };
```
chrisnowak, line 10: I'd have the step take explicit properties.  

e.g. videoId, height, title, text, transcript, and nextButtonText as explicit things here...that makes storybook testing very easy and also potential re-use of this component easy

chrisnowak: (Don't feel super strongly about it though)

chrisnowak:
🚢 looks awesome...this will be fantastic to build our next generation of survey steps on

ashleyrtate:
> looks like package-lock has been deleted on this branch now...probably will need to install / re-commit that before merging

Definitely was not deleted. Looks like it was accidentally upgraded to lockFileVersion 3 again when I was wrestling with rollup the other night. I've reset it to version 2 again.

chrisnowak:
Oh weird, it was showing all red for me before.

ashleyrtate:
I think just because there were like 17k lines of deletions at the top or something

src/components/step/YouTubeStep/YouTubeStep.tsx
ashleyrtate: That does make it more nicely self-documenting so I refactored these out. Styles are even more opaque and in need of documentation...so I started playing with a couple of approaches and may refactor how we override those in a follow on PR.

src/components/step/StepLayout/StepLayout.css
```diff
@@ -0,0 +1,37 @@
+.step-container {
+    background-color: #FFFFFF;
+}
+
+.step-container div,dl,dt,dd,ul,ol,li,h1,h2,h3,h4,h5,h6,pre,form,fieldset,input,textarea,p,blockquote,th,td {
```
chrisnowak, line 5: @ashleyrtate I think this rule is wrong and these are supposed to all be embedded under .step-container right?  So needs to be .step-container div, .step-container dl, .step-container dt, etc.

ashleyrtate: 👍

ashleyrtate: What issue did you see from this? Is it screwing up a view style?

PR #18, opened 2023-02-06
Survey Step Alternative Styling POC

ashleyrtate:
## Overview

Experiments with a different approach to step styling than passing a style blob object through the top level step component down to the low level components. This is a partial implementation to demonstrate the pattern with just StepTitle.

To use this approach a step implementer would add styling support by 

1. Adding a style: StepStyleProps object to their step props definition
2. Adding a StepStyle component within their step component.

This feels more inline with how you would normally override css styling/appearance outside React and makes the style overrides clearer and more self-documenting.

Custom steps not explicitly supported in MyDataHelps should be able to follow a variation of this pattern to override default styling of the low level components without being forced to reimplement them.

If we agree this is the preferred approach it should be easy to apply to the other low level components.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/step/StepStyle/StepStyle.tsx
+++ b/src/components/step/StepStyle/StepStyle.tsx
@@ -1,0 +1,22 @@
+import React from 'react'
+
+export interface StepStyleProps {
+  titleAlignment?: string;
+  titleColor?: string;
+  titleFontSize?: string;
+  titleFontWeight?: string;
+}
+
+export default function (props: StepStyleProps) {
+    return (
+      <style>
+          .step-container .step-title &#123;
+            text-align: {props.titleAlignment};
+            color: {props.titleColor};
+            font-size: {props.titleFontSize};
+            font-weight: {props.titleFontWeight};
+            &#125;
+      </style>  
+    );
+}
+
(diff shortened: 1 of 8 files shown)


ashleyrtate:
@chrisnowak lots of other stuff going on but flagging just in case you missed this one...I expect you already spent some time thinking about style overrides for the views so wanted to see if you had any strong opinions one way or the other.

chrisnowak:
So generally speaking I think what bugs me is that for the most part, parent components should affect their children...not siblings.  And yeah I mean you could make a case that it's the same as if the YoutubeStep embedded the style block directly (affecting it's children), or if you put it in StepLayout...but it just feels off / unexpected when it's like "this thing works as expected, but only when it's on the page with this thing"

I would probably be more of a fan if it was in StepLayout directly since that would feel more hierarchical / obvious to someone building another step in the future - which is probably a thing that will happen soon.  Your parent component might apply some CSS to child components...that is more expected.  

But also honestly I think the current way is fine and doesn't offend me or anything honestly.

ashleyrtate:
Yeah, maybe moving to StepLayout would be clearer. Just feels like the styles are much less obvious/documented relative to the step properties. And passing the style overrides down through the components feels a bit wrong. Styling in general does not seem to have a clearly favored pattern in React (at least in my short time looking around).

PR #20, opened 2023-03-03
Celebration Step

ajmenca:
## Overview

This is an expansion of the work begun in https://github.com/CareEvolution/MyDataHelpsUI/pull/17, providing new templated WebView Step Types.

This PR adds a Celebration Step.  The Celebration Step has a layout which is similar to an Instruction Step, providing parameters for Title, Text, DetailText, IconUrl, and ImageUrl.  When the step loads, there is a brief animation of confetti which falls from the top of the frame.

image

Confetti animation was pulled from https://www.cssscript.com/confetti-falling-animation/ which I made a few small changes to; I updated it to TypeScript, and adjusted how/where it creates the Canvas so that the confetting overlays the entire page.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/step/StepImageIcon/StepImageIcon.css
```diff
@@ -0,0 +1,11 @@
+.step-container .step-image-icon {
```
chrisnowak, line 1: I unfortunately missed this on @ashleyrtate 's original PR, but I think for now at least we should prefix all styles with ```.mdhui- ``` since that's the pattern we've followed so far in this repo

ashleyrtate: Seems pretty safe to just refactor those style names in this PR if you want. I don't think they would be referenced anywhere else yet.

ajmenca: Yeah - pretty focused find/replace.  I've prefixed everything in the new "Step" components.  Can all be verified via the storybook.

src/components/step/StepImage/StepImage.css
```diff
@@ -0,0 +1,10 @@
+.mdhui-step-container .mdhui-step-image {
+    text-align: center;
+}
+.mdhui-step-container .mdhui-step-image img {
+    width: 100%;
+    padding: 0;
+    margin: 0;
+    padding-top: 40px;
```
ashleyrtate, line 8: For the existing step images we don't set padding-bottom. Is this needed to account for styling differences in the other stop components?

ashleyrtate (approved):
Just one small question about image padding, otherwise looks good to me.

src/components/step/StepImage/StepImage.css
ajmenca: Yes; I'd have to look back and do the comparison - but on our normal InstructionStep there's some padding of the container that contains the title/text/detail portion, which is absent in the Step components here.  Seemed slightly easier to just handle it here, rather than have to re-arrange the other CSS and potentially affect the existing YouTube step.

PR #24, opened 2023-03-22
Garmin Connect

zwkohn:
## Overview
https://github.com/CareEvolution/Consumers/issues/11519
This introduces components to connect and show Garmin devices. It is basically a copy of the Fitbit components.

For the DailyDataProviders, I only added Steps and RHR for now.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/container/ConnectGarmin/ConnectGarmin.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+FetchingData.args = { previewState: "fetchingData", title: "Garmin" };
+
+export const NotEnabledDefault = Template.bind({});
+NotEnabledDefault.args = { previewState: "notEnabled", title: "Garmin" };
+
+export const NotEnabledHide = Template.bind({});
+NotEnabledHide.args = { previewState: "notEnabled", title: "Garmin", disabledBehavior: "hide" };
+
+export const NotEnabledDisplayError = Template.bind({});
+NotEnabledDisplayError.args = { previewState: "notEnabled", title: "Garmin", disabledBehavior: "displayError" };
```
ajmenca, line 41: This story is not showing anything?  It shows up the same as NotEnabledHide.  The Fitbit version appears to be doing the same thing.

src/components/container/ConnectGarmin/ConnectGarmin.tsx
ajmenca, line 1: If this is (as it appears to be) essentially a copy/paste of the ConnectFitbit component, it feels like we shouldn't be copy/pasting it.  Is there anything that prevents all of the identical functionality from being contained in a shared component which is customized by props passed in?  It looks like the only *functional* difference between the two widgets is the provider ID that they use?

src/components/container/GarminMonthCharts/GarminMonthCharts.tsx
```diff
@@ -0,0 +1,56 @@
...
+	var year = intervalStart.getFullYear();
+	return (
+		<div style={{ display: visible ? "block" : "none" }}>
+			<DateRangeNavigator intervalType="Month" intervalStart={intervalStart} onIntervalChange={updateInterval} />
+			<DeviceDataMonthChart
+				onDataDetected={() => onDataDetected()}
+				previewState={props.previewState == "connected" ? "WithData" : props.previewState == "notEnabled" || props.previewState == "notConnected" ? "NoData" : undefined }
+				lines={[{
+					showAverage: true,
+					dailyDataType: DailyDataType.GarminSteps,
```
ajmenca, line 36: This again looks like the only thing that differs between this and the Fitbit version?  This should be a configurable option on a shared widget.

ajmenca (changes requested):

src/helpers/strings-en.ts
```diff
@@ -54,7 +57,7 @@
     "external-account-reconnect": "Reconnect",
     "external-account-refresh": "Refresh",
     "external-account-remove": "Remove",
-    "device-data-no-data": "If you have connected Apple Health, Google Fit, or Fitbit, check back later to view your data.",
+    "device-data-no-data": "If you have connected Apple Health, Google Fit, Fitbit, or Garmin, check back later to view your data.",
```
greinard, line 60: This update is missing from the Spanish file.

greinard (commented):
I like Adam's suggestion about a reusable component.  Maybe `ConnectDevice` rather than `ConnectProvider` for the reusable component, since I think this may be a component we'd only ever use for device manufacturers.

src/components/container/GarminMonthCharts/GarminMonthCharts.tsx
zwkohn: I feel like this might be a little different than the other components because each device type might have different data types to display. I only implemented Steps and HR here which do happen to both be implemented in each device, but maybe there are data types that are only in one or the other (Garmin Body Battery for example). 

With that said maybe there is a way to structure this that would support this. thoughts @ajmenca?

ajmenca: Side note: I guess the `previewState` needs to also be defined on the wrapper and passed down to `MonthCharts` too?  But that's a quirk of our storybook setup mostly.

ajmenca: Also, with all this said... I pretty much have this working locally in this form to ensure that I wasn't giving you a bad recommendation 😆 

I can just push it, if you'd like.

zwkohn: Ha.. that is what I was thinking as I was reading through these.

In the immortal words of Salt N Pepa I would say "push it"

ajmenca (approved):
Re-ran the storybook; everything's looking good to me.

PR #37, opened 2023-04-03
Dark Mode support, fewer global styles, various new props

chrisnowak:
## Overview

Originally this was just an attempt to support "dark mode" or a dark color scheme with MyDataHelpsUI components; but turned into including a number of other changes.

Currently MyDataHelpsUI has a couple issues:
- It doesn't have much flexibility in terms of overriding styles; the original intent was that you can always override CSS and that certainly works but isn't the most stable thing to do and breaks "encapsulation"
  - Ideally think we should get to a place where we can use CSS modules to keep CSS encapsulated, and consumers of MyDataHelpsUI should never have to override a .mdhui style
- MyDataHelpsUI had numerous global styles it was putting on the page; these were non-optional and kind of non-obvious in the Layout component.  The default "button" styles were the most offensive

So this does "dark mode" but also tackles a few of those fundamental issues.

Detailed list of changes:
- All color references were turned into CSS variables to support dark mode; many of them were already variables but this was cleaned up and made more consistent
  - These can be overridden in downstream libraries to change the look and feel if desired
- Layout now includes global CSS via the emotion react library's "Global" component; this means that if you do not have a ```<Layout>``` on the page, you don't get any global styles
- Layout supports a few new properties:
  - primaryColor is a shortcut to overriding the main "theme" color for buttons, links, etc. (you could also override this via CSS)
  - noGlobalStyles, when set to true, will not inject any of the global styles on body, links, html, etc. - primarily for things like default font and dynamic text
  - colorScheme can be "light", "dark" or "auto" ("light" by default); and basically sets the CSS variables to that color scheme and also populates a LayoutContext that can be used by downstream components if they need to programmatically know if they are in light / dark color scheme.  "auto" color scheme will pick up the color scheme from the OS (like the css prefers-color-scheme: dark query) 
    - Colors for dark mode were borrowed from iOS 
  - I also deprecated stylesheetPath since that really was a hacky way to conditionally include global CSS, but I've never used it
- colorScheme="auto" was added to all stories and views; so it will impact our out of box views in the next update.  But any other dowstream stuff will opt into it
- The default global "button" styles that unstyled all buttons were removed and turned into an "UnstyledButton" component, inspired by https://mantine.dev/core/unstyled-button/ - this component mostly exists for accessibility reasons so tab completion etc. works
- A "className" was added to a bunch of components, to support downstream libraries being able to set a class on components that they can style without having to know about the CSS of these components.  This was mostly done by walking back from the components I needed in Symptom Shark and making it so I did not have to override any mdhui stuff
  - Face
  - Histogram
  - LoadingIndicator
  - Section
  - SegmentedControl
  - NavigationBar
  - DateRangeNavigator
  - Switch
  - TextBlock
  - TrackerItem
- A boolean "loading" prop was added to Button so you can have a button that shows a loading indicator
- A variant prop was added to Button; right now the only supported options are "default" and "subtle"; subtle basically just hides the background and shows the text in the button color
- A variant prop was added to "LoadingIndicator"; supported options are "default" and "inline" (which makes it display inline)
- A variant prop was added to "DateRangeNavigator"; supported options are "default" and "rounded".  "rounded" makes the navigator appear like Symptom Shark's date range navigator
- I added vscode formatting; sorry for the white-space flog on many files, recommend browsing with w=1

## Security

### Testing
- [X] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation


src/components/container/DeviceDataMonthCharts/DeviceDataMonthCharts.stories.tsx
```diff
@@ -13,7 +13,7 @@ export default {
 } as ComponentMeta<typeof DeviceDataMonthCharts>;
 
 const Template: ComponentStory<typeof DeviceDataMonthCharts> = (args: DeviceDataMonthChartsProps) =>
-	<Layout>
+	<Layout  colorScheme="auto">
```
greinard, line 16: Not a big deal, but looks like most (maybe all) of these have two spaces before the attribute.

ajmenca (commented):
A couple of suggestions from running down the Storybook in Dark Mode.  Overall looks great.

src/components/presentational/Button/Button.tsx
```diff
@@ -9,11 +11,36 @@ export interface ButtonProps {
...
+	}
+	if (props.variant === "subtle") {
+		classes.push("mdhui-button-subtle");
+	}
+
 	return (
-		<button className="mdhui-button" disabled={props.disabled ?? false} onClick={() => props.onClick()}>
+		<button style={{ backgroundColor: props.disabled ? undefined : props.color }}
+			className={classes.join(" ")}
+			disabled={(props.disabled || props.loading) ?? false}
```
greinard, line 32: Do we need the `?? false` here?

src/components/presentational/Layout/Layout.tsx
```diff
@@ -21,12 +41,41 @@ export default function (props: LayoutProps) {
...
 	}
 
+	let colorScheme: "light" | "dark" = "light";
+	if (props.colorScheme === "auto" && window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
+		colorScheme = "dark";
+	} else if (props.colorScheme === "dark" || props.colorScheme === "light") {
+		colorScheme = props.colorScheme;
+	}
+	console.log(props.colorScheme);
+	console.log(colorScheme);
```
greinard, line 51: Some leftover debug logging.

src/components/presentational/SingleExternalAccount/SingleExternalAccount.css
```diff
@@ -17,17 +17,20 @@
...
-    height:100%;
+    border: none;
+    background: none;
+    height: 100%;
 }
 
 .mdhui-single-external-account .external-account-header .external-account-provider-logo {
     height: 24px;
     display: block;
+    background: #FFF;
```
greinard, line 31: Move to a var maybe?

src/components/presentational/WeekCalendar/WeekCalendar.css
```diff
@@ -1,101 +1,88 @@
...
+    background-color: var(--mdhui-color-primary);
+    color: #FFF;
+    font-weight: bold;
+}
 
-    .mdhui-week-calendar .mdhui-week-calendar-day-future > *
-    {
-        opacity: .4;
-    }
+.mdhui-week-calendar .mdhui-week-calendar-day-future>* {
```
greinard, line 86: I think this change is harmless, but just wanted to mention it in case `.class > *` functions differently than `.class>*`.

src/components/presentational/LoadingIndicator/LoadingIndicator.tsx
```diff
@@ -3,10 +3,24 @@ import { faRefresh } from '@fortawesome/free-solid-svg-icons/faRefresh'
 import "./LoadingIndicator.css"
 import '@fortawesome/fontawesome-svg-core/styles.css';
 import { FontAwesomeSvgIcon } from 'react-fontawesome-svg-icon';
+import { ProjectSupport } from '../../container';
```
greinard, line 6: I don't think this is being used.

greinard:
Just a couple of nitpick comments.  Looks good to me overall.  I concur with Adam on the couple hard to read color comments as well.

src/components/presentational/Card/Card.css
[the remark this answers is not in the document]
chrisnowak: This did have extra styling applied; I removed that extra styling.

src/components/presentational/SingleExternalAccount/SingleExternalAccount.css
chrisnowak: This is intentionally hardcoded since this is an image background

src/components/presentational/LoadingIndicator/LoadingIndicator.tsx
chrisnowak: weird, not sure where that came from

PR #73, opened 2023-06-13
Add an AppDownload component to the HomeView.

greinard:
## Overview

This PR adds a new `AppDownload` component to the `HomeView`.

Also added:
- A component for easily rendering localized strings.
- A custom hook to replace some commonly repeated boilerplate initialization code.

## Security

This update is mainly about rendering a couple of mobile app store links when the user is on the web platform.  It should not present any additional security risk. 

The `LocalizedString` component does use `dangerouslySetInnerHTML`, but the values it populates that field with come from the language files.  It is necessary to support HTML formatting in some of the language entries.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation


zwkohn:
This is cool. One thing that comes to mind is something we did for the Uzima project where on non-mobile Web instead of presenting the icons to download the app, we showed a QR code that the user could scan with their phone to download the app.

src/components/container/AppDownload/AppDownload.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { useInitializeView } from '../../../helpers/Initialization';
+import Card from '../../presentational/Card';
+import TextBlock from '../../presentational/TextBlock';
+import googlePlayDownload from '../../../assets/google-play-download.svg';
+import appStoreDownload from '../../../assets/app-store-download.svg';
+import language from "../../../helpers/language";
+import LocalizedString from "../../presentational/LocalizedString";
+
+export interface AppDownloadProps {
+	previewProjectPlatforms?: string[]
```
zwkohn, line 15: maybe some history I am missing or something, but is there a reason for the "preview" prefix?

src/components/container/AppDownload/AppDownload.tsx
```diff
@@ -0,0 +1,66 @@
...
+	if (!projectInfo) {
+		return null;
+	}
+
+	return (
+		<div className="mdhui-app-download">
+			<PlatformSpecificContent platforms={['Web']} previewDevicePlatform={props.previewDevicePlatform}>
+				<Card>
+					<div className="mdhui-app-download-title">
+						<LocalizedString stringName="app-download-title"/>
```
zwkohn, line 43: Would someone want to customize this title? (not sure how that works w/ Localization)

src/helpers/Initialization.tsx
```diff
@@ -0,0 +1,16 @@
+﻿import MyDataHelps from '@careevolution/mydatahelps-js';
+import {useEffect} from 'react';
+
+const debounce = require('lodash.debounce');
+
+export function useInitializeView(initialize: () => void): void {
```
ajmenca, line 6: This is a very handy shortcut; I've copy/pasted this in a ton of places in another application without thinking to make it a custom hook.  Maybe two small suggestions:
1) I didn't instinctively know what the hook was doing where you used it in the AppDownload widget until I came down here.  I know it's a long name, but perhaps `useApplicationDidBecomeVisibleInitialization`?
2) It might be helpful to have a second parameter to the hook which is the array of dependencies that passed to useEffect?

src/components/container/AppDownload/AppDownload.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+import Layout from '../../presentational/Layout'
+import AppDownload, { AppDownloadProps } from './AppDownload';
+
+export default {
+	title: 'Container/AppDownload',
+	component: AppDownload,
+	parameters: {
+		layout: 'fullscreen',
+	}
+} as ComponentMeta<typeof AppDownload>;
```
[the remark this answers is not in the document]
greinard: Fixed.

src/components/container/AppDownload/AppDownload.tsx
greinard: Yes.  It is because those properties are only used to configure the "preview" rendering of this component (and sub components).  Those values are not actually configuration for this component.  They just configure the fake state that is to be used when "previewing" the component in the storybook or during testing.

src/components/presentational/LocalizedString/LocalizedString.tsx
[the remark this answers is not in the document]
chrisnowak: I agree with this^ not sure there is really any reason to make this component..

I think if we find our current localization stuff lacking I would probably also explore just using https://react.i18next.com/

src/components/container/AppDownload/AppDownload.tsx
greinard: Possibly.  I think if we provide that capability, localization would be the responsibility of the overriding application.

I'm not sure if any/many of our current reusable components allow for that kind of customization. If we want to be more flexible about that, maybe we can do something at a global level that won't require modifying each of the individual components to support it.

src/components/presentational/LocalizedString/LocalizedString.tsx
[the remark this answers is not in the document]
greinard: I've removed it for now.  I do like that it makes things a bit more declarative and less javascript-y looking in the HTML you are building.  You can also provide override style classes and it can isolate the calling components from the data structure we currently use for language (and any changes that that may undergo in the future).  I don't feel strongly about keeping it though.

src/helpers/Initialization.tsx
greinard: I'm not opposed to renaming it, but I'm hesitant to rename it something to do with a specific event.  Similar to your suggestion about the array of dependencies, I had considered allowing folks to pass in a collection of events too.  I have a version of this in a couple projects, and one of them registers two different events.

I'll see if I can update it to support both of those customization scenarios.

greinard: I've added the ability to pass in additional events as well as dependencies.

Let me know if you have any other naming suggestions given the latest functionality.

I went with `useInitializeView` because I wanted it to be `initializeView`, but being a hook, it needed the `use` prefix.

greinard:
Planning to merge near the end of the day tomorrow (Tuesday) unless there are further concerns or requests for delay.

ajmenca (approved):

src/helpers/Initialization.tsx
```diff
@@ -0,0 +1,24 @@
+﻿import MyDataHelps, { EventName } from '@careevolution/mydatahelps-js';
+import { DependencyList, useEffect } from 'react';
+
+const debounce = require('lodash.debounce');
```
zwkohn, line 4: this is breaking downstream apps that use vite. I see the following when starting up the MyDatahelpsStarterKit:
Screenshot 2023-06-26 at 3 30 38 PM

zwkohn: this looks to be a bug with vite: https://github.com/vitejs/vite/issues/3409 but I am not sure that has been published yet?

chrisnowak: Oh yeah, don't use require...I had to remove a bunch of require stuff out of here when moving to vite

greinard: Taking a look now.

greinard: https://github.com/CareEvolution/MyDataHelpsUI/pull/78

PR #85, opened 2023-07-13
Fix bugs with garmin sleep calculations

chrisnowak:
## Overview

Garmin has a couple data idiosyncracies that I didn't account for.

It appears that Garmin can produce multiple sleep data points on a given night that overlap, e.g. 
image
This might just be a temporary artifact for the most recent night's sleep until Garmin syncs and removes the shorter sleep period.

To account for this state, however, I am now taking the max value.  

Previously I also missed dividing the value when we summed the values together which caused the sleep values to be massively inflated.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/helpers/daily-data-providers/garmin-sleep.ts
+++ b/src/helpers/daily-data-providers/garmin-sleep.ts
@@ -1,7 +1,7 @@
 import { add, formatISO, parseISO } from "date-fns";
 import queryAllDeviceData from "./query-all-device-data";
 
-function querySleep(property: string, startDate: Date, endDate: Date, sumValues: boolean, divideBy?: number) {
+function querySleep(property: string, startDate: Date, endDate: Date, divideBy?: number) {
     return queryAllDeviceData({
         namespace: "Garmin",
         type: "Sleep",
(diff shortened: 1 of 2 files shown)


src/helpers/daily-data-providers/garmin-sleep.ts
```diff
@@ -24,36 +24,34 @@ function querySleep(property: string, startDate: Date, endDate: Date, sumValues:
                 value = value / divideBy;
             }
 
-            if (!data[dataKey]) {
+            if (!data[dataKey] || data[dataKey] < value) {
```
zwkohn, line 27: trying this out on my data the only way I can get two "Sleep" DDPs for a day is to edit a nights sleep. at which point I see an ENHANCED_FINAL and an AUTO_MANUAL DDP for that day.

With this logic I think the value will be whichever is less, which might be different between property types? (that is if I am drawing a graph w/ LightSleep and REMSleep the Y values might come from the different DDPs across days)



Screenshot 2023-07-13 at 12 37 17 PM

chrisnowak: Oh 🤦 good catch, I was trying to take the max not the min.  I actually got 2 ENHANCED_FINAL readings actually on the same night

PR #93, opened 2023-07-20
Adding NumericCalculationStep

Suvarna-Thejas-CE:
## Overview

This creates a NumericCalculationStep that takes in the calculation string from the Designer and calculates the output.

In order to test the functions, I implemented the jest testing framework and added some unit test cases for the logic of the calculation. 

This should follow a similar model as the celebration/YouTube step in the sense that it will be implemented as a step type in MDHD. 

I did not include a story because this step type only shows a loading screen. It also retrieves data mainly from MDH.js.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


Suvarna-Thejas-CE:
Function has been unit tested, but would love some thoughts on how to test the actual react component. Promise chaining is done correctly, submission happens at the right time, etc.

src/components/step/NumericCalculationStepContainer/NumericCalculationStepContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+            setCalculationResult(calculateValue(calculationString, surveyAnswers).toString());
+            console.log(calculationResult);
+            setCalculationReady(true);
+        })
+    }, []);
+
+    return (
+        <StepLayout>
+            <LoadingIndicator/>
+            <OnVisibleTrigger enabled={calculationReady} onTrigger={() => {MyDataHelps.completeStep(calculationResult)}}/>
```
chrisnowak, line 36: You shouldn't need to have an OnVisibleTrigger here; the use case for OnVisibleTrigger  is basically just to enable infinite scroll type situations.  

You should be able to just call MyDataHelps.completeStep in the useEffect and not worry about having a "calculationReady" property or anything

src/components/step/NumericCalculationStepContainer/NumericCalculationStepContainer.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { useState, useEffect } from 'react'
+import MyDataHelps, {StepConfiguration} from '@careevolution/mydatahelps-js';
+import { calculateValue } from '../../../helpers/calculationStepFunctions/numericCalculationFunctions';
+import StepLayout from '../StepLayout'
+import { LoadingIndicator } from '../../presentational';
+import OnVisibleTrigger from '../../presentational/OnVisibleTrigger/OnVisibleTrigger';
+
+export default function () {
+
+    const [calculationString, setCalculationString] = useState<string>("");
```
chrisnowak, line 10: You shouldn't any state variables at all.  

in useEffect you can just do

```
  MyDataHelps.getStepConfiguration().then(function(config: StepConfiguration){
     MyDataHelps.getCurrentSurveyAnswers().then(function(answers){
         calculateValue
         MyDataHelps.completeStep

     });
  });
```

src/components/step/NumericCalculationStepContainer/NumericCalculationStepContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+          
+          setCalculationString(config.properties.calculation);
+          return new Promise((resolve, reject) => {
+            resolve(config.properties.calculation);
+          })
+        })
+        .then((() => MyDataHelps.getCurrentSurveyAnswers()))
+        .then((surveyAnswers) => {
+            setCalculationResult(calculateValue(calculationString, surveyAnswers).toString());
+            console.log(calculationResult);
```
chrisnowak, line 28: Remove before merging

src/helpers/calculationStepFunctions/numericCalculationFunctions.ts
```diff
@@ -0,0 +1,54 @@
+import { create, all } from 'mathjs'
```
chrisnowak, line 1: oh neat find with mathjs - I like it better than using eval.

So actually...all this string parsing logic should be unnecessary?  Because mathjs supports passing variables?  Like here's one of the examples:

```
let scope = {
    a: 3,
    b: 4
}
math.evaluate('a * b', scope)           // 12
math.evaluate('c = 2.3 + 4.5', scope)   // 6.8
scope.c                                 // 6.8
```

So you could imagine setting up your scope as:
```
let scope = {
    '[Q1]': 3,
    '[Q2]': 4,
    '[F1.Q1]': 5
}
```
from your survey answers.  For form variables (where step id/result id are different) one could just use a dot separator like [F1.Q1].

Or you can set up the scope to call a function to get the answer it looks like: https://mathjs.org/examples/advanced/custom_scope_objects.js.html

The main question/issue is whether mathjs is kind/permissive enough to allow numbers in it's symbols; seems like there is some flexibility there but not sure how much.

src/helpers/calculationStepFunctions/numericCalculationFunctions.ts
```diff
@@ -0,0 +1,54 @@
...
+  const resultIdentifier = splitString.length > 1 ? splitString[1].trim() : undefined;
+
+  try {
+    return (resultIdentifier != undefined ? 
+        parseFloat(surveyAnswers.find((answer) => {
+            return(answer.stepIdentifier == stepIdentifier && answer.resultIdentifier == resultIdentifier);
+        })!.answers[0])
+        : parseFloat(surveyAnswers.find((answer) => {
+            return (answer.stepIdentifier == stepIdentifier);
+        })!.answers[0]));
```
ajmenca, line 49: It's really interesting to see you tackle this because I've just recently been making a generic Calculation WebViewStepTemplate for another project.  It's interesting to compare/contrast how we went about this.  In my case, I knew *specifically* how the questions were going to be formatted (and that all of them were Form Steps), so I didn't try and make my version extremely generic.

Which brings me to my question here, wondering about constraints on how this step could actually be used.  Am I correct in that if the step was a Question Step with the Text Choice Answer Format, and a Text Choice Style of Multiple Choice that you could have multiple answers in the array here?  What's the expectation on how to calculate that type of step?  I could see an argument that it should be the sum of the parseFloat of the answers?

src/helpers/calculationStepFunctions/numericCalculationFunctions.ts
Suvarna-Thejas-CE: Cool. I've been playing with this and have a few thoughts on feasibility.

1. the `.` is the object access operator. So this seems to prevent us from being able to do calculations with stepIds/resultIds that have a `.` in the name (ie. `Form1.Q1.2`). I think we can probably add a little validation message that says when you have calculation step, the calculation string cannot reference a stepIdentifier/resultIdentifier combo that has a `.` in it. 
^How does that sound?

2. When assembling the scope dictionary from the survey answers, does this rule work? if stepIdentifier only appears once in the survey answer list, then use just `{stepIdentifier: value}`, otherwise if it appears more than once use `stepIdentifier: {resultIdentifier1: value, resultIdentifier2: value}`

src/helpers/calculationStepFunctions/numericCalculationFunctions.ts
chrisnowak: Wonder what REDCap does for multiple choice.  If there is an easy way to match them maybe worth it.  Otherwise feels like we should ignore and not try to handle it

src/helpers/calculationStepFunctions/numericCalculationFunctions.ts
chrisnowak: @Suvarna-Thejas-CE maybe dig into REDCap a bit here; what kind of validation / restrictions do they have. 

Even might be worth it if we follow their syntax for the vars; not sure if math.js will allow []  though, e.g. ```[clab_lipidldl_u2] * [clab_lipidldlab]```

For 2...after thinking about it...I think we should just key it by result identifier.  It makes it so much simpler...and I would add validation so that your result identifiers have to be unique (e.g. you can't have Q1 on FormStep1 and Q1 on FormStep2) if you have a numeric calculation step.  

Thoughts?

Suvarna-Thejas-CE: On my preliminary testing, the `[]` do work in mathJS :) so that should work great. Will look into REDCap a bit here. 

I like **not** having restrictions on the step/resultIdentifiers. The '_' is a cool idea to me and feels like a great one from a consumer perspective.

For me, the usability concern for using only resultIdentifiers is that the question step UI doesn't mention "resultIdentifier". Could potentially be fixed with documentation, but would choose not to force customers to rely on reading user guide articles.

src/helpers/calculationStepFunctions/numericCalculationFunctions.ts
Suvarna-Thejas-CE: image

"If possible, avoid" haha. 

Maybe in our docs we say something similar, but explain the behavior if they do use it. Or should we say if length of the results for a step > 1 then throw an error.

src/helpers/calculationStepFunctions/numericCalculationFunctions.ts
chrisnowak: Ah hmm I wasn't suggesting we do anything with underscores actually...was not a separator I was proposing, sorry.  Those were just some redcap vars I copy/pasted.  

Basically just thinking that we have [resultIdentifier] be the syntax; and we enforce unique result identifiers when using this step because in practice they are anyways, and it will be a pain to construct a step identifier / result identifier pair (e.g. [formstepidentifier_resultidentifier] or whatever you were thinking)

In terms of usability, I think that might just be how we describe it.  It is either the step identifier OR the form item identifier if it is a form step basically; so we can avoid the use of "resultIdentifier" if we want (or we can clarify in the survey builder as well).  But it again feels easier to me than the user having to write out [formstepidentifier_resultidentifier] if they can just always do [resultidentifier].  

We don't necessarily have to enforce it via a validation rule.  We can just have [resultidentifier] be the way this works and if they happen to have multiple overlapping resultidentifiers, it just takes the latest survey answer with that resultidentifier.

src/helpers/calculationStepFunctions/numericCalculationFunctions.ts
chrisnowak: Yeah for an MVP the latter might make the most sense...just don't allow results > 1

src/helpers/calculationStepFunctions/numericCalculationFunctions.ts
ajmenca: For what it's worth; using only the result identifier is what I did on my version of this since I knew all my questions were form questions, and the result identifiers were unique.  What stinks about that is that there's not a good way to find the right question that the result identifier is on (there's lots of repeated looping).  I debated, in my version, making a dictionary to do lookups so that I only did the looping once, but didn't take it that far.

That said, I'm also somewhat in favor of only the result identifier being in the eventual UI because, just as an example, the survey I was working on required doing a sum total of 91 variables for its final calculation.  Listing 91 variable names is pretty gnarly in a text input; it's even worse if they're longer because they require the question identifier as well.

Suvarna-Thejas-CE: Cool - yeah. Y'all are convincing me that there is value to using ResultIdentifier. The functional name seems to be "QuestionIdentifier" which, like you said Chris, is defined as the StepIdentifier for Question Steps or FormItemIdentifier for Form Steps. I think this is workable.

It does bring up some questions about taxonomy for me (that I want to document, and maybe get reactions too, but is a bit of scope creep for this PR specifically). 

In the case, where I have repeated questions (i.e, For all the people in your family, specify the height and weight). I would do something that looks like:
```
Form Step: Person1
 -- Form Item: Height
 -- Form Item: Weight
Form Step: Person2
 -- Form Item: Height
 -- Form Item: Weight
Form Step: Person3
 -- Form Item: Height
 -- Form Item: Weight
```
Currently, it's not obvious that there would be a relationship between a form item identifier and a question step identifier. 

I think it makes sense to make that relationship explicit, just through naming changes. 

- Renaming Form Item to Question or Question Item 
- Potentially renaming Form Step to multi-Question step

It would change the taxonomy from:
```
- Question Step
   -- Question
- Form Step
  -- Form Item
     --- Question
  -- Form Item
     --- Question
```

To this model that makes the relationship between multi-question and question steps clear. Functionally, I think it also describes better what a form step is, just a collection of questions. 
 ```
- Question Step
   -- Question Identifier
- Multi-Question Step Identifier
   -- Question Identifier
   -- Question Identifier
```

Comment (chrisnowak) [reply src/helpers/calculationStepFunctions/numericCalculationFunctions.ts:1]: Well...we cannot go renaming things like crazy unfortunately...have to consider the downstream implications given that creates inconsistencies with exported data and the survey answer API.  Have to think of the scope of the change we are making here

Comment (Suvarna-Thejas-CE) [reply src/helpers/calculationStepFunctions/numericCalculationFunctions.ts]: Maybe we can give this another quick review after the latest push

File: src/helpers/calculationStepFunctions/numericCalculationFunctions.test.js [hunk 6]
```diff
@@ -0,0 +1,38 @@
+import { calculateValue } from "./numericCalculationFunctions";
```
ashleyrtate, line 1: Would be great if you could add a note to the project wiki or readme about adding the test framework and how to run the tests.

src/components/step/NumericCalculationStepContainer/NumericCalculationStepContainer.tsx
```diff
@@ -0,0 +1,23 @@
...
+import MyDataHelps, {StepConfiguration} from '@careevolution/mydatahelps-js';
+import { calculateValue } from '../../../helpers/calculationStepFunctions/numericCalculationFunctions';
+import StepLayout from '../StepLayout'
+import { LoadingIndicator } from '../../presentational';
+
+export default function () {
+    
+    useEffect(() => {
+        MyDataHelps.getStepConfiguration().then((config: StepConfiguration) => {
+            if (!config) return; // allows test mode to work
```
chrisnowak, line 11: Is this needed anymore?

src/components/step/NumericCalculationStepContainer/NumericCalculationStepContainer.tsx
```diff
@@ -0,0 +1,23 @@
...
+            MyDataHelps.getCurrentSurveyAnswers().then((surveyAnswers) => {
+                MyDataHelps.completeStep(calculateValue(config.properties.calculation, surveyAnswers).toString());
```
chrisnowak, line 13: Very minor but since calculateValue is the important thing here maybe put it on a separate line above; e.g. 

let value = calculateValue(config.properties.calculation, surveyAnswers).toString();
MyDataHelps.completeStep...

src/helpers/calculationStepFunctions/numericCalculationFunctions.ts
```diff
@@ -0,0 +1,40 @@
...
+      return(error.message);
+    }
+  }
+}
+
+const createScope = (surveyAnswers: CurrentSurveyAnswer[]) => {
+  let answerDict: {[id: string] : number} = {};
+  surveyAnswers.map((answer) => {
+    if (answer.answers.length > 1) {
+      throw new Error(`Calculation Error - Calculation not available for multiple-select text choice questions (${answer.resultIdentifier})`)
```
chrisnowak, line 28: I think we would not want to throw errors for any of these?  We should just skip it and include the variables we can...otherwise any survey with a multi-select text choice or a "." in the resultIdentifier will not work

src/helpers/calculationStepFunctions/numericCalculationFunctions.test.js
```diff
@@ -0,0 +1,38 @@
...
+    test('Calculation includes variable that is not a number', () => {
+      expect(calculateValue(
+        "(([Q1]*[Q1_2])+[Q2])-[Q3]",
+        [{"date":"2023-07-19T11:14:18.377-04:00","stepIdentifier":"Q1","resultIdentifier":"Q1","answers":["Jacob"]},{"date":"2023-07-19T11:14:18.944-04:00","stepIdentifier":"Form1","resultIdentifier":"Q1_2","answers":["4"]},{"date":"2023-07-19T11:14:20.338-04:00","stepIdentifier":"Q2","resultIdentifier":"Q2","answers":["2"]},{"date":"2023-07-19T11:14:20.849-04:00","stepIdentifier":"Q3","resultIdentifier":"Q3","answers":["3"]}]
+      )).toBe('Calculation Error - Answer not a number (Q1)');
+    });
+
+    test('Calculation Passes', () => {
+      expect(calculateValue(
+        "(([Q1]*[Q1_2])+[Q2])-[Q3]",
```
chrisnowak, line 34: Maybe I am missing it but it seems like we did not do anything special for the ```[]``` syntax this to work, no?  So that is all built into Math.js and actually you do not even need the [] in most cases.  Q1 * Q3 works fine as an expression it seems.

However, in my quick test, seems like Q1_2 does not work and you need the brackets for some reason.  Curious if you know why or have figured out what the [] actually does

I think there are other characters that do not work as intended; for instance spaces - [Q1 2] doesn't work.

src/helpers/calculationStepFunctions/numericCalculationFunctions.ts
```diff
@@ -0,0 +1,40 @@
...
+  const config = { }
+  const math = create(all, config)
+
+  try {
+    let answerDict: {[id: string] : number} = createScope(surveyAnswers);
+    try {
+      return(math.evaluate(calcString, answerDict)._data[0].toString());
+    }
+    catch {
+      return("Calculation Error - Unmatched Variable")
```
chrisnowak, line 14: This might not be an unmatched variable but could be any other error...probably just want to surface the error from Math.js

We should probably figure out the right behavior for the numeric step if there is an error.  Looks like you have it just proceeding...bit conflicted if that is the right call...it does maximize participant experience I suppose

chrisnowak: However, if subsequent steps depend on the calculation for branching logic etc...

src/components/step/NumericCalculationStepContainer/NumericCalculationStepContainer.tsx
Suvarna-Thejas-CE: Still need the config in order to get the calculation string.

src/helpers/calculationStepFunctions/numericCalculationFunctions.test.js
Suvarna-Thejas-CE: The [] is the array operator in mathjs. In this case, it doesn't really do anything. I used it because I liked having a "syntax" to designate which is a variable. 

When I was testing the Q1_2 without the [] worked correctly and gave the correct response, but you're right. The [Q1 2] and the Q1 2, will actually calculate but result in an correct calculation. 

Here is a [list of operators ](https://mathjs.org/docs/expressions/syntax.html), so many of these won't work. Doesn't really make sense to error check for all of them. But what is our list of basic ones? dash, space, period? what about parentheses, brackets?

src/helpers/calculationStepFunctions/numericCalculationFunctions.ts
Suvarna-Thejas-CE: I feel like showing more helpful error messages can be a real boon for people that are trying to learn how to use the tool. Giving them an error message about why something failed and a place to look can help them solve problems and know what to look for in the future. 

I have modified how I do the error checking. At a high level:

1. When building the results dictionary from the survey answers, create a list of Identifiers that are flagged for some error. 

2. Parse the calculation string for the variables that are used (this is where having the [] was very very helpful), and compare to the flagged identifiers from step 1.

3. Create an error message that shows these identifiers.

I had to employ this model to catch any errors for Identifiers that had a space followed by a number (i.e. in the case of [Q1] + [Q1 2], Q1 2 was getting replaced by the value of Q1 * 2, but the calculation would succeed successfully, so I wasn't able to catch the error of an identifier with a space if i didn't cross check with the variables used in the calculation string.

PR #94, opened 2023-07-22
Add blank view, view header, variants for buttons, other tweaks

chrisnowak:
## Overview

Some minor additions to support common use cases.

There is now a "ViewHeader" component which can be used for the top of a view.

I also added a "BlankView" with simple simplified properties to support common use cases; it has a title/subtitle and the ability to show back/close buttons.  If back/close buttons are displayed it will use a NavigationBar at the top, otherwise it will use a StatusBarBackground/ViewHeader.

An additional button variant called "light" has been added as well which uses the button color as the text color but with the light grey background.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


explunit:
I don't understand React, but the overall approach seems good to me. Just one possible security thing we might want to address, and I think it would actually be in MDH itself, not this PR:

Since the base64-encoded URL for the custom view tab includes the URL of the custom view and we whitelist the renderer url prefix, it's possible to send a participant a URL which renders a view that was never configured for that study.

image

image

It looks like at this point you do need to know the patientid. Otherwise it fails while trying to render the new view. But I'm not sure if we can count on that behavior.

explunit:
Sorry, commented on wrong PR

chrisnowak:
Will be superceded shortly

PR #99, opened 2023-08-07
Incentives component

zwkohn:
## Overview
This adds an Incentives component to the library
Screenshot 2023-08-07 at 4 12 23 PM

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/presentational/IncentiveBar/IncentiveBar.css
+++ b/src/components/presentational/IncentiveBar/IncentiveBar.css
@@ -1,0 +1,92 @@
+.incentive-bar {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    flex-shrink: 0;
+	margin: 0 10px;
+	width: 100%;
+	max-width: 500px;
+	--star_size: 16px;
+	--star_border_width: 2px;
+	--green: #00A862;
+	--lt_green: #BFE9D8;
+}
+
+.incentive-bar-earned {
+    position: absolute;
+    height: calc( var(--star_size) + 4px);
+	top:0;
+}
+
+.incentive-bar .incentive-bar-background {
+    flex-shrink: 0;
+    border-radius: 500px;
+	border-width: var(--star_border_width) ;
+	border-style: solid;
+    background: #fff;
+	overflow: hidden;
+	width: 100%;
+	position: relative;
+	height:calc( var(--star_size) + 4px);
+}
+
+.incentive-bar .incentive-text{
+	font-size: 0.65em;
+	line-height: inherit !important;
+}
+
+.incentive-bar .reward-earned {
+	color: #fff;
+	line-height: inherit !important;
+}
+
+.incentive-bar .reward-waiting {
+	background-color: #fff;
+	line-height: inherit !important;
+}
+
+.incentive-bar .incentive-stars {
+	display: flex;
+	justify-content: space-between;
+	width: 100%;
+	position: relative;
+	z-index: 2;
+	height: var(--star_size);
+}
+
+.incentive-bar .incentive-star {
+	display: none;
+}
+	
+@media (min-width: 350px) {
+	.incentive-bar {
+		--star_size: 28px;
+	}
+	.incentive-bar .incentive-star {
+		border-radius: 50%;
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		font-weight: 600;
+		width:var(--star_size);
+		text-align: center;
+		border-radius: 50%;
+		border-width: var(--star_border_width);
+		border-style: solid;
+		vertical-align: top;
+		line-height: inherit !important;
+		height: var(--star_size);
+		margin: -1px 0;
+		position: relative;
+		top: 1px;
+	}
+	.incentive-bar .incentive-star span {
+		font-size: 13px;
+		display: flex;
+		align-items: center;
+		justify-content: center;
+	}
+	.incentive-bar .incentive-star.incentive-text span {
+		font-size: 12px;
+	}
+}
--- a/src/components/presentational/IncentiveBar/IncentiveBar.stories.tsx
+++ b/src/components/presentational/IncentiveBar/IncentiveBar.stories.tsx
@@ -1,0 +1,25 @@
+import React from "react";
+import IncentiveBar, { IncentiveBarProps } from "./IncentiveBar";
+import { ComponentMeta, ComponentStory, Meta, StoryFn } from "@storybook/react";
+import Layout from "../Layout";
+
+export default {
+	title: "Presentational/IncentiveBar",
+	component: IncentiveBar,
+	parameters: {
+		layout: 'auto',
+	}
+} as Meta<typeof IncentiveBar>;
+
+const Template: StoryFn<typeof IncentiveBar> = (args: IncentiveBarProps) => <Layout colorScheme="auto"><IncentiveBar {...args} /></Layout>;
+
+export const Default = Template.bind({});
+Default.args = {
+	primaryColorString: "#00A862",
+	secondaryColorString: "#BFE9D8",
+	increments: 10,
+	totalAvailable: 100,
+	earned: 30,
+	showDetail: true,
+
+} as IncentiveBarProps;
--- a/src/components/presentational/IncentiveBar/IncentiveBar.tsx
+++ b/src/components/presentational/IncentiveBar/IncentiveBar.tsx
@@ -1,0 +1,84 @@
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import "./IncentiveBar.css";
+import React, { useEffect, useState } from "react";
+import { faStar } from '@fortawesome/free-solid-svg-icons';
+
+export interface IncentiveBarProps {
+	primaryColorString?: string;
+	secondaryColorString?: string;
+	increments: number;
+	totalAvailable: number;
+	earned: number;
+	showDetail: boolean;
+}
+
+export default function IncentiveBar(props: IncentiveBarProps) {
+	const [mainColor, setMainColor] = useState<string>('#00A862');
+	const [secondaryColor, setSecondaryColor] = useState<string>('#80D4B0');
+	const [calculatedWidth, setCalculatedWidth] = useState<string>("0px");
+	const containerRef = React.createRef<HTMLDivElement>();
+
+	function getStarStyle(earned: number, threshold: number) {
+		return {
+			color: earned >= threshold ? '#fff' : mainColor,
+			backgroundColor: earned >= threshold ? mainColor : '#fff',
+			borderColor: secondaryColor
+		}
+	}
+
+	useEffect(() => {
+		if (props.primaryColorString) {
+			setMainColor(props.primaryColorString);
+		}
+		if (props.secondaryColorString) {
+			setSecondaryColor(props.secondaryColorString);
+		}
+	}, [props.primaryColorString, props.secondaryColorString]);
+
+	useEffect(() => {
+		if (props.totalAvailable % props.increments != 0) {
+			console.error("IncentiveBar increments must be a factor of totalAvailable");
+		}
+		const totalStars = props.totalAvailable / props.increments + 1;
+		const earned = props.earned / props.increments;
+		const starWidthStr = containerRef != null ? window.getComputedStyle(containerRef!.current!).getPropertyValue('--star_size') : "0px";
+		const starWidth = parseInt(starWidthStr.substring(0, starWidthStr.length - 2));
+		console.log("starWidth", starWidthStr);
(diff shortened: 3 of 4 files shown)


chrisnowak:
Few thoughts:
- I think we want to name this a lot more narrowly; e.g. "IncentivesProgressBar"; or honestly just "ProgressBar" since it may not always be in the context of incentives but could work for goal setting etc.
- We should do a bit of a review of other incentives progress bars we have done for other studies...e.g. PROGRESS...may suggest different options or look/feel.  PROGRESS may be the gold standard here in some ways
image - I would see if we can track a bit more to that look / feel
- Would suggest we do not have an "increment" or anything...but rather we should leave this as a presentational component and have:
  - background (string that does into CSS background prop; could be color or gradient)
  - fill (string that does into CSS background prop of the fill layer; could be color or gradient)
  - fillPercent (how wide the fill bar is): 0 to 100
  - steps[] e.g. the placement of the various icons...each of these would have
    - stepPercent: number 0 to 100 for where the step should be placed on the progress bar
    - a react element for the thing that should be displayed on the step...could be an icon or something...
image

Then I think maybe another component for "IncentiveProgressBarStepIcon" that has some options like a $$ amount that you can use in the "step" array...

ajmenca:
A couple of reference points from recent work on AoU Rewards.  They use two variations of this kind of idea; one shows progress "in" level progression, the other shows progress across the entire reward program.

image
image

image
image

- I think that for both the AoU and Progress designs, having the right side of the fill bar be curved looks really nice, I think we should integrate that.
- I think both of these designs also animate the bar when the percentage changes; that's definitely a feature we should have. There's also some subtle animations that the AoU Bar has; if you hit 100%, the bar animates to the top, and *then* the star icon changes from gray to orange.  This is also a really nice effect that looks polished.
- In your screenshots, it looks like some things might need to get adjusted left/right by a pixel or so.  There's some aliasing between the circle of the "100" and the blue border, and also in the second screenshot there's some white pixels leaking through on the left side of the completion bar.

bxh:
AOU clearly has more than one kind of progress bar in mind.

Screenshot 2023-08-15 160615

Suvarna-Thejas-CE:
I like the idea of making this just a pure progress bar. I created an issue #162 that could be a wrapper around this component that has labels/text that make it useful for a specific use case. 

This one I think would be good with the parameters that Chris suggested above.

src/components/presentational/ProgressBar/ProgressBar.css
```diff
@@ -0,0 +1,50 @@
+.progress-bar {
```
chrisnowak, line 1: Should prefix all CSS classes with .mdhui-, so this should be .mdhui-progress-bar

I would also nest all classes in this file under that - so for the fill it shuold be ```.mdhui-progress-bar .mdhui-progress-bar-fill```

chrisnowak: (alternative to consider at some point is using CSS modules; but so far we haven't taken that leap and CSS is still being overridden in places)

src/components/presentational/ProgressBar/ProgressBar.css
```diff
@@ -0,0 +1,50 @@
...
+	display:inline-block;
+	margin: 10px 0px;
```
chrisnowak, line 3: Would remove the margin on this

src/components/presentational/ProgressBar/ProgressBar.css
```diff
@@ -0,0 +1,50 @@
...
+	width: 100%;
+	max-width: 500px;
```
chrisnowak, line 5: Why a max-width?  Will look awkward on desktop so I would remove

src/components/presentational/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,41 @@
+import "./ProgressBar.css";
+import React, { ReactElement, useEffect } from "react";
+
+export interface ProgressBarProps {
+	background?: string;
+	fill?: string;
+	fillPercent: number;
+	steps?: [
+		{
+			stepPercent: number;
```
chrisnowak, line 10: "step" prefix here seems unnecessary since this will be in a "steps" array; so would just go with percent/icon

src/components/presentational/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,41 @@
+import "./ProgressBar.css";
+import React, { ReactElement, useEffect } from "react";
+
+export interface ProgressBarProps {
+	background?: string;
```
chrisnowak, line 5: We use color suffix for color properties, so would use backgroundColor and fillColor.

Also would make both background/fill a ColorDefinition and not a string - that was probably introduced after this PR but see the button component for an example (ColorDefinition allows specifying 2 colors, one for light mode and one for dark mode)

src/components/presentational/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,41 @@
...
+	fillPercent: number;
+	steps?: [
+		{
+			stepPercent: number;
+			stepIcon: ReactElement;
+		}
+	]
+}
+
+ProgressBar.defaultProps = {
```
chrisnowak, line 16: What the heck how have I never heard about "defaultProps" before now

src/components/presentational/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,41 @@
...
+	background: "#d3d3d3",
```
chrisnowak, line 17: Would have this be the closest CSS backround color var from the globalCss file

src/components/presentational/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,41 @@
+import "./ProgressBar.css";
+import React, { ReactElement, useEffect } from "react";
+
+export interface ProgressBarProps {
+	background?: string;
+	fill?: string;
+	fillPercent: number;
```
chrisnowak, line 7: I think we can add this component to the view builder just by itself, but that requires a couple additional props.

We need to add the "innerRef" prop so this can be dragged/dropped, and I also would support a "defaultMargin" prop that we would set to true for the view builder which would add the default 16px margin to this (see the link for example)

src/components/presentational/ProgressBar/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+export default {
+	title: "Presentational/ProgressBar",
+	component: ProgressBar,
+	parameters: {
+		layout: 'auto',
+	},
+	argTypes: {
+		background: { control: 'color', description: "The background color of the progress bar" },
```
chrisnowak, line 16: Documentation!  Awesome.  Need to start doing that...

src/components/presentational/ProgressBar/ProgressBar.tsx
chrisnowak: Oh hmm, i'm realizing that fill might actually be a gradient...yet let me think about that for a sec

src/components/presentational/ProgressBar/ProgressBar.css
```diff
@@ -0,0 +1,50 @@
...
+	position: absolute;
+	z-index: 2;
+	display:inline-block;
+}
+
+.progress-bar .step-icon {
+	position: absolute;
+	transform: translate(-100%, 0)
```
chrisnowak, line 43: This does make things a bit awkward if you have a stop near the beginning of the progress bar - it will get clipped out...there's no way to have a stop at 0 reliably I think (since steps are based on percent).

That being said, this seems like the best and most reasonable solution to this and, as demonstrated by the stories, works pretty well.  I don't think having a stop right at the beginning makes a ton of sense

chrisnowak: and you could always fix that by having the element at the beginning have a left margin or something

src/components/presentational/ProgressBar/ProgressBar.css
```diff
@@ -0,0 +1,50 @@
...
+}
+
+.progress-bar .step-icon span {
```
chrisnowak, line 46: I wouldn't assume a span will be used as the child element...

So there is actually some cool stuff you did in the story here for the various little star icons and how they are colored etc.

I would not just waste that in a story but would actually invest in a separate component called "ProgressBarStep"...it could be exported under the progress bar...such that we have some out of box progress bar steps that can be placed on this (but someone could also come up with their own to supply to this).

and the props could be like:
- borderColor?: ColorDefinition
- backgroundColor: ColorDefinition
- children  (this is where you'd put in the font awesome icon for the star, or the text 75 or whatnot)
- height (based on this it would auto-calculate the top margin to center itself vertically, e.g. for the large star)


e.g. for the big star at the end your step would be like:

```
steps: [{
		stepPercent: 100,
		stepIcon: <ProgressBarStep borderColor="yellow" backgroundColor="orange" height="24px">
<FontAwesomeIcon icon={faStar} style={{ color: "yellow" }} />
</ProgressBarStep>
	}],
```

Basically to have to limit the amount of work someone has to do to get it to look like the storybook, while still allowing some flexibility

src/components/presentational/ProgressBar/ProgressBar.css
```diff
@@ -0,0 +1,51 @@
...
+	--icon_border_width: 2px;
+}
+
+.mdhui-progress-bar .mdhui-progress-bar-default-margin {
+	margin: 16px;
+}
+
+.mdhui-progress-bar .mdhui-progress-bar-background {
+	border-radius: 500px;
+	border-width: var(--icon_border_width) ;
```
chrisnowak, line 15: Huh...did you just add a border to this?  I didn't think there was a border before

src/components/presentational/ProgressBar/ProgressBar.css
```diff
@@ -0,0 +1,51 @@
+.mdhui-progress-bar {
+	display: block;
+	width: 100%;
+	position: relative;
+	--icon_size: 24px;
+	--icon_border_width: 2px;
+}
+
+.mdhui-progress-bar .mdhui-progress-bar-default-margin {
```
chrisnowak, line 9: Looks like this isn't working because you have a space between these and this class is applied to the progress bar itself.

Also, you will need to adjust the width like we do for buttons so it's ```width:calc(100% - 32px)```, otherwise it overflows

src/components/presentational/ProgressBar/ProgressBar.css
chrisnowak: i would probably drop the border because it makes it look a bit awkward when stuff overlaps it:
Screenshot 2023-11-30 at 2 04 36 PM

src/components/presentational/ProgressBar/ProgressBar.stories.css
```diff
@@ -0,0 +1,3 @@
+.sb-story {
```
chrisnowak, line 1: What does this CSS do?

src/components/presentational/ProgressBar/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+	},
+	argTypes: {
+		backgroundColor: { control: 'object', description: "The background color of the progress bar" },
+		fillColor: { control: 'object', description: "The fill color of the progress bar" },
+		fillPercent: { control: 'number', description: "The percent of the progress bar that is filled" },
+		steps: { control: 'object', description: "An array of steps to display on the progress bar. A step specifies the position on the progress bar as a percent and a React element to display at that position." },
+	}
+} as Meta<typeof ProgressBar>;
+
+const Template: StoryFn<typeof ProgressBar> = (args: ProgressBarProps) => <Layout colorScheme="auto"><ProgressBar {...args} /></Layout>;
```
chrisnowak, line 25: Minor, maybe put "defaultMargin" on this so it's spaced out from the edges

chrisnowak (approved):
Looks good!

src/components/presentational/ProgressBar/ProgressBar.stories.css
zwkohn: Trying to get some padding in the storybook component. I replaced this with an extra div in the Template
```
const Template: StoryFn<typeof ProgressBar> = (args: ProgressBarProps) => <Layout colorScheme="auto"><div style={{padding: "35px 0"}}><ProgressBar {...args} /></div></Layout>;
```

PR #100, opened 2023-08-09
Add ConnectEhrStep

bxh:
## Overview

This PR adds the ConnectEhrStep, as specified in https://github.com/CareEvolution/Consumers/issues/11603.

https://github.com/CareEvolution/MyDataHelpsUI/assets/3732224/8b9c3742-d8b0-4a2b-9363-a84b719eaf40

A few notes about the implementation.

- In the initial state, the next button is disabled.
- In order to proceed, a participant needs to select a provider using ProviderSearch component, which initiates EHR connection process.
- Internally, there is a polling process that begins once a provider is selected, which checks for whether an external account from that provider exists.
- When such provider account is found, the polling process stops and the participant is allowed to proceed to the next step.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/step/ConnectEhrStep/ConnectEhrStep.tsx
```diff
@@ -0,0 +1,54 @@
...
+            />
+            <StepText
+                text={props.text}
+                textAlign={props.styles.textAlignment}
+                color={props.styles.textColor}
+                fontSize={props.styles.textFontSize}
+                fontWeight={props.styles.textFontWeight}
+            />
+            <ProviderSearch
+                previewState={props.previewState ? "Default" : undefined}
```
ashleyrtate, line 37: Could you add some doc around this previewState property? What is it doing and what are the allowed values. Is it just whether the search is executed to preview the list of providers on component load?

ashleyrtate (approved):
Generally looks good except for the question about previewState.

src/components/step/ConnectEhrStep/ConnectEhrStep.tsx
bxh: It allows Storybook to display ConnectEhrStep with the default provider search preview state. Would it be clearer if we use the type of `ProviderSearchPreviewState` instead?

bxh:
I've updated the static screenshot with a live recording. In the meantime, I've also made sure that we open EHR page in a new window, similar to #104. It was the experience the Scripps team specifically requested, and I think it makes sense. 

Besides, I've also positioned the next button at the bottom so that it is not kept pushing down by the provider search component, which supports infinite scroll.

Let me know what you think. @ashleyrtate @chrisnowak

src/components/step/ConnectEhrStep/ConnectEhrStep.tsx
ashleyrtate: Yes, I think that would be clearer naming. But do you mean this property is only used because of the storybook? A comment explaining the use and reason would be good.

bxh: Yes, it only exists because of the storybook. I will add a comment.

bxh: Anything else?

ashleyrtate (approved):

src/components/container/ProviderSearch/ProviderSearch.tsx
```diff
@@ -93,9 +95,13 @@ export default function (props: ProviderSearchProps) {
         debouncedPerformSearch(event.target.value);
     }
 
-    function connectToProvider(providerID: number) {
-        if (!(linkedExternalAccounts[providerID] && linkedExternalAccounts[providerID].status != 'unauthorized')) {
-            MyDataHelps.connectExternalAccount(providerID);
+    function connectToProvider(provider: ExternalAccountProvider) {
+        const providerID = provider.id;
+        if (!props.previewState && !(linkedExternalAccounts[providerID] && linkedExternalAccounts[providerID].status != 'unauthorized')) {
```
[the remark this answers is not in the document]
bxh: You can find another `!` if you go one level above. 

It allows users connecting to a provider if they don't already have an account _or_ their account is marked as unauthorized.

PR #97, opened 2023-07-27
Updates to survey task list look / feel

chrisnowak:
## Overview

This updates the look / feel of the survey tasks / survey task list a bit based on some recent designs and adds some more options.  

The individual survey tasks no longer have the checkmark on the left side; instead there is a button to start/resume that is more explicit and user friendly:
image
Start changes to "resume" once progress has been saved for a task.  Completed tasks display the checkmark:
image

Additionally, there are now a few variants/options for SurveyTaskList.

- "noCard" is the default to preserve backwards compatibility; basically it is the same as today
- "singleCard" means that the task list will be embedded in a card
- "multiCard" expands the task list so each task has it's own card, and this is now what is used by default on the survey tasks view to give it more breathing room:
image

## Security

REMINDER: All file contents are public.

- [X] I have ensured no secure credentials or sensitive information remain in code, metadata, comments, etc.
  - [X] Please verify that you double checked that .storybook/preview.js does not contain your participant access key details. 
  - [X] There are no temporary testing changes committed such as API base URLs, access tokens, print/log statements, etc.
- [X] These changes do not introduce any security risks, or any such risks have been properly mitigated.

Describe briefly what security risks you considered, why they don't apply, or how they've been mitigated.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app
- [X] [MyDataHelps website](mydatahelps.org)

### Documentation
- [X] I have added relevant Storybook updates to this PR as well.

Consider "Squash and merge" as needed to keep the commit history reasonable on `main`.

## Reviewers

Assign to the appropriate reviewer(s). Minimally, a second set of eyes is needed ensure no non-public information is published. Consider also including:
- Subject-matter experts
- Style/editing reviewers
- Others requested by the content owner

PR #105, opened 2023-08-29
Survey Task List Updates, Date Range Coordinator, New Charts + more

chrisnowak:
Supercedes both https://github.com/CareEvolution/MyDataHelpsUI/pull/97 and https://github.com/CareEvolution/MyDataHelpsUI/pull/94

This makes numerous changes to MyDataHelpsUI to support some new projects; since the previous PRs haven't gotten reviews yet, I figured I would wrap this into one large PR.

### Survey Task List Updates

This updates the look / feel of the survey tasks / survey task list a bit based on some recent designs and adds some more options.  

The individual survey tasks no longer have the checkmark on the left side; instead there is a button to start/resume that is more explicit and user friendly:
image
Start changes to "resume" once progress has been saved for a task.  Completed tasks display the checkmark:
image

Additionally, there are now a few variants/options for SurveyTaskList.

- "noCard" is the default to preserve backwards compatibility; basically it is the same as today
- "singleCard" means that the task list will be embedded in a card
- "multiCard" expands the task list so each task has it's own card, and this is now what is used by default on the survey tasks view to give it more breathing room:
image

### Blank View / View Header / Button Variants

There is now a "ViewHeader" component which can be used for the top of a view.

I also added a "BlankView" with simple simplified properties to support common use cases; it has a title/subtitle and the ability to show back/close buttons.  If back/close buttons are displayed it will use a NavigationBar at the top, otherwise it will use a StatusBarBackground/ViewHeader.

An additional button variant called "light" has been added as well which uses the button color as the text color but with the light grey background.

### Daily Data Chart

Screenshot 2023-08-29 at 10 46 34 AM

There is now a new "Daily Data Chart" which should be a simpler and more robust replacement for the current "DeviceDataMonthChart".  

You can now do bar, area, or line charts.  It has both a weekly and monthly view.  It works both inside a date range coordinator or outside.  It will display most recent data by default, and has specific options for when the week starts (e.g. is it always sunday, 6 days ago, 7 days ago)

It allows for both a value formatter (for the tooltip) and a value converter (for instance to convert minutes of sleep to hours for the chart) and some color options for bars.

### Date Range Coordinator 
There is also a new "Date Range Coordinator" which works in tandem with the daily data charts and potential future time-based components.

Date Range Coordinator allows for children and exposes a DateRangeContext to it's children.  This makes it easier and more natural to just do something like:

```
<DateRangeCoordinator intervalType="Month">
    <DailyDataChart title="Steps"
	dailyDataType="DailyDataType.Steps"
	chartType="Line" />
</DateRangeCoordinator>
```

Without having to write the connecting code to pass props around.

It supports a rounded variant and similar options for week/month or when the week starts on by default.

### Language is now a function
This touched a lot of files but isn't really that big of a change.  Previously, the "language" object was imported on page render based on the language in the query string.  

That doesn't work if you ever want to change the language of your view at run time.  Admittedly this is not common but I have a use case where it is necessary.  Ultimately it was mostly changing a lot of ```language["key"]``` to ```language("key")```

## Security

### Testing
- [X] MyDataHelps iOS app
- [X] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/container/ExternalAccountsLoadingIndicator/ExternalAccountsLoadingIndicator.tsx
```diff
@@ -57,6 +57,10 @@ export default function (props: ExternalAccountsLoadingIndicatorProps) {
     }
 
     function initialize() {
+        if (props.previewState) {
+            setIsWeb(true);
+            return;
```
greinard, line 62: Should this call `refresh()` before returning?

chrisnowak: Yep, totally should...I just realized that myself

src/components/presentational/Action/Action.css
```diff
@@ -1,4 +1,5 @@
-﻿.mdhui-action {
+﻿.mdhui-action,
+.mdhui-unstyled-button.mdhui-action {
```
greinard, line 2: I'm curious why the additional class was needed here?  Was `.mdhui-action` not specific enough to have precedence over something else?

src/components/presentational/Button/Button.tsx
```diff
@@ -14,20 +14,29 @@ export interface ButtonProps {
...
+
+	let backgroundColor = props.color;
+	let textColor = "#FFF";
 	if (props.variant === "subtle") {
-		classes.push("mdhui-button-subtle");
+		textColor = props.color ? props.color : "var(--mdhui-color-primary)";
+		backgroundColor = "transparent";
+	}
+
+	if (props.variant === "light") {
```
greinard, line 33: Minor:  Could use `else if` here.

src/components/presentational/Action/Action.css
chrisnowak: This might have been a local dev artifact where it wasn't working right for me.  The precedence as it is right now depends on CSS being imported in a specific order - mdhui-action has to be imported after mdhui-unstyled-button, which should happen due to the order of imports.  I have mostly seen that be the case but saw some situations where it wasn't; so just for safety added the more specific selector to make sure this works.

src/helpers/get-interval-start.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export type WeekStartsOn = "Sunday" | "Monday" | "Tuesday" | "Wednesday" | "Thursday" | "Friday" | "Saturday" | "6DaysAgo" | "7DaysAgo";
+
+export function getWeekStart(weekStartsOn?: WeekStartsOn) {
+    if (!weekStartsOn) {
+        weekStartsOn = "Sunday";
+    }
+    let today = new Date();
+    if (weekStartsOn === "6DaysAgo" || weekStartsOn === "7DaysAgo") {
+        let today = new Date();
```
greinard, line 21: I think you can remove this line.

src/components/container/DailyDataChart/DailyDataChart.tsx
```diff
@@ -0,0 +1,270 @@
...
+    barColor?: string
+}
+
+export interface AreaChartOptions {
+    lineColor?: string
+    areaColor?: string
+}
+
+function getDefaultIntervalStart(intervalType: "Week" | "Month", weekStartsOn?: WeekStartsOn) {
+    let intervalStart = new Date();
```
greinard, line 41: No need to initialize `intervalStart` here.

src/components/container/DailyDataChart/DailyDataChart.tsx
```diff
@@ -0,0 +1,270 @@
...
+    return intervalStart;
+}
+
+export default function DailyDataChart(props: DailyDataChartProps) {
+    let [currentData, setCurrentData] = useState<DailyDataQueryResult | null>(null);
+    let [hasAnyData, setHasAnyData] = useState(false);
+
+    const dateRangeContext = useContext<DateRangeContext | null>(DateRangeContext);
+    let intervalType = props.intervalType || "Month";
+    let intervalStart: Date = new Date();
```
greinard, line 56: No need to initialize `intervalStart` here.

src/components/container/DailyDataChart/DailyDataChart.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+    const DayTick = ({ x, y, stroke, payload }: any) => {
+        var value = payload.value;
+        if (intervalType == "Month") {
+            let currentDate = new Date(intervalStart.getFullYear(), intervalStart.getMonth(), value);
+            return <text className={isToday(currentDate) ? "today" : ""} fill="var(--mdhui-text-color-2)" x={x} y={y + 15} textAnchor="middle" fontSize="12">{value}</text>;
+        } else {
+            let currentDate = intervalStart;
+            let dayOfWeek = format(currentDate, "E").substr(0, 1);
+            let month = currentDate.getMonth();
```
greinard, line 158: I don't think `month` gets used anywhere.

src/components/container/DailyDataChart/DailyDataChart.tsx
```diff
@@ -0,0 +1,270 @@
...
+    };
+
+    const DayTick = ({ x, y, stroke, payload }: any) => {
+        var value = payload.value;
+        if (intervalType == "Month") {
+            let currentDate = new Date(intervalStart.getFullYear(), intervalStart.getMonth(), value);
+            return <text className={isToday(currentDate) ? "today" : ""} fill="var(--mdhui-text-color-2)" x={x} y={y + 15} textAnchor="middle" fontSize="12">{value}</text>;
+        } else {
+            let currentDate = intervalStart;
+            let dayOfWeek = format(currentDate, "E").substr(0, 1);
```
greinard, line 157: This substring length doesn't match the one below.

Also, I think you can use `let dayOfWeek = format(currentDate, "EEEEEE");` and drop the substring calls.

Further, do we need to initialize `dayOfWeek` here?  Won't it always get set in the loop below?

src/components/container/DailyDataChart/DailyDataChart.tsx
```diff
@@ -0,0 +1,270 @@
...
+                if (domainMin == "Auto") {
+                    domain = ["auto", "auto"];
+                } else if (domainMin != undefined) {
+                    domain = [domainMin, "auto"];
+                }
+            }
+        }
+
+        return <>
+            <defs>
```
greinard, line 201: Could these defs be moved down with their respective charts since they aren't really applicable globally.  One or the other or both is not relevant for every chart at the moment.

greinard (approved):
The code looks good to me.  I just had a few minor nitpicks and questions.

I didn't notice any issues while playing around with the storybook either.

The only thing that I think would be slightly more visually appealing to me would be if the survey task buttons (i.e. start/resume) were of equal width.  I don't feel too strongly about it though.

Feel free to :ship: when ready.

src/components/container/DailyDataChart/DailyDataChart.tsx
chrisnowak: Huh...no idea what that was for.

src/components/container/DailyDataChart/DailyDataChart.tsx
chrisnowak: ah nice catch.  Yeah will just initialize it to empty string so typescript doesn't complain

src/components/container/DailyDataChart/DailyDataChart.tsx
chrisnowak: Ah yep, not sure why I did it this way

chrisnowak:
Thanks @greinard !

PR #113, opened 2023-09-18
Support "innerRef" property on most components

chrisnowak:
## Overview

This adds "innerRef" to the props of most of the components (basically anywhere it was convenient to do so).

In general I think this is useful for attaching additional behaviors to these components.  The most pressing reason, however, had to do with a situation I wasn't able to cleanly work around.

Currently to attach a ref to a "DailyDataChart" or something you would have to just wrap it inside a div. 

Cards are also supposed to hide themselves when they are empty.

However, if you have a situation like this:
```
<Card>
    <div ref="myRef">
        <DailyDataChart hideIfNoData ... />
    </div>
    <div ref="myRef2">
        <DailyDataChart hideIfNoData ... />
    </div>
</Card>
```

The Card is not actually empty if both charts are hidden; it still has 2 divs in it.

I attempted to fix this with CSS using the "has" pseudo-selector to detect if a card ONLY has empty elements; however unfortunately a number of browsers do not support that.

So the most elegant solution I can think of is to support innerRefs such that we can do:

```
<Card>
    <DailyDataChart  innerRef="myRef" hideIfNoData ... />
    <DailyDataChart  innerRef="myRe2" hideIfNoData ... />
</Card>
```

## Security

### Testing
- [x] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/container/ConnectDevice/ConnectDevice.tsx
+++ b/src/components/container/ConnectDevice/ConnectDevice.tsx
@@ -19,6 +19,7 @@ export interface ConnectDeviceProps {
 	previewState?: ConnectDevicePreviewState,
 	disabledBehavior?: 'hide' | 'displayError',
 	dataCollectionProperty:	string 
+	innerRef?: React.Ref<HTMLDivElement>
 }
 
 export type ConnectDevicePreviewState = ExternalAccountStatus | "notConnected" | "notEnabled";
@@ -112,7 +113,7 @@ export default function (props: ConnectDeviceProps) {
 	}
 
 	return (
-		<div className="mdhui-connect-device">
+		<div className="mdhui-connect-device" ref={props.innerRef}>
 			{props.title &&
 				<CardTitle title={props.title} />
 			}
--- a/src/components/container/ConnectDevice/ConnectDevice.tsx
+++ b/src/components/container/ConnectDevice/ConnectDevice.tsx
@@ -103,7 +103,7 @@ export default function (props: ConnectDeviceProps) {
 	if (!deviceEnabled) {
 		if (props.disabledBehavior == 'displayError' && !loading) {
 			return (
-				<div className="mdhui-connect-device">
+				<div className="mdhui-connect-device" ref={props.innerRef}>
 					<div className="content">{props.title} is not enabled for this project.</div>
 				</div>
 			);
--- a/src/components/container/ConnectEhr/ConnectEhr.tsx
+++ b/src/components/container/ConnectEhr/ConnectEhr.tsx
@@ -15,6 +15,7 @@ export interface ConnectEhrProps {
 	previewState?: ConnectEhrPreviewState,
 	disabledBehavior?: 'hide' | 'displayError'
 	bottomBorder?: boolean
+	innerRef?: React.Ref<HTMLButtonElement>
 }
 
 export type ConnectEhrApplicationUrl = "preview" | string;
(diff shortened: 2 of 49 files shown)


src/components/container/ConnectDevice/ConnectDevice.tsx
```diff
@@ -112,7 +113,7 @@ export default function (props: ConnectDeviceProps) {
 	}
 
 	return (
-		<div className="mdhui-connect-device">
+		<div className="mdhui-connect-device" ref={props.innerRef}>
```
greinard, line 116: I can't comment high enough in the diff, but line `106` probably needs the `ref` as well.

src/components/container/ConnectEhr/ConnectEhr.tsx
```diff
@@ -95,7 +96,7 @@ export default function (props: ConnectEhrProps) {
 	}
 
 	return (
-		<Action bottomBorder={props.bottomBorder} title={language('connect-ehr-title-prefix') + language('connect-ehr-title-providers') + language('connect-ehr-title-divider') + language('connect-ehr-title-health-plans')} className="mdhui-connect-ehr" onClick={() => connectToEhr()}>
+		<Action innerRef={props.innerRef} bottomBorder={props.bottomBorder} title={language('connect-ehr-title-prefix') + language('connect-ehr-title-providers') + language('connect-ehr-title-divider') + language('connect-ehr-title-health-plans')} className="mdhui-connect-ehr" onClick={() => connectToEhr()}>
```
greinard, line 99: Similar here.  I can't comment high enough.  Line `85` may need the `ref`.

src/components/container/HealthPreviewSection/HealthPreviewSection.tsx
```diff
@@ -98,7 +99,7 @@ export default function (props: HealthPreviewSectionProps) {
         return null;
     }
 
-    return <Action indicatorPosition={props.indicatorPosition} bottomBorder indicatorValue={model.Count} className="mdhui-health-preview-section" title={getTitle()} titleIcon={<img className="mdhui-health-preview-icon" src={getIconUrl()} />} onClick={() => props.onClick()}>
+    return <Action innerRef={props.innerRef} indicatorPosition={props.indicatorPosition} bottomBorder indicatorValue={model.Count} className="mdhui-health-preview-section" title={getTitle()} titleIcon={<img className="mdhui-health-preview-icon" src={getIconUrl()} />} onClick={() => props.onClick()}>
```
greinard, line 102: Line `95` may need the `ref`.

src/components/container/LabResultsSummary/LabResultsSummary.tsx
```diff
@@ -87,6 +88,7 @@ export default function (props: LabResultsSummaryProps) {
     }
 
     return <Action
+        innerRef={props.innerRef}
```
greinard, line 91: Line `78` may need the `ref`.

chrisnowak: Good catch on all of these.  Did end up having to wrap most of these in a <div> since the ref can't be both an HTMLDivElement ref and HTMLButtonElement ref - but that seems to work fine.

greinard (approved):

PR #104, opened 2023-08-14
ConnectDeviceAccountStep

bxh:
## Overview

This PR adds ConnectDeviceAccountStep.
- Allow a participant to connect to their Omron/Fitbit/Garmin account.
- Advance to the next survey step if the participant has already been connected.

### Garmin connection demo
https://github.com/CareEvolution/MyDataHelpsUI/assets/3732224/b69459d3-7c19-4cc8-959e-90eb8fb53391

### Automated advance demo
The test subject has already been connected to Garmin.
https://github.com/CareEvolution/MyDataHelpsUI/assets/3732224/e5635d65-0e87-44a2-a070-8ba3dfeacdea

## Security

REMINDER: All file contents are public.

- [x] I have ensured no secure credentials or sensitive information remain in code, metadata, comments, etc.
  - [x] Please verify that you double checked that .storybook/preview.js does not contain your participant access key details. 
  - [x] There are no temporary testing changes committed such as API base URLs, access tokens, print/log statements, etc.
- [x] These changes do not introduce any security risks, or any such risks have been properly mitigated.

Describe briefly what security risks you considered, why they don't apply, or how they've been mitigated.

All code changes are sandboxed as other survey steps are. Interaction with the server is handled exclusively by MDH.js.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app
- [ ] [MyDataHelps website](mydatahelps.org)

### Documentation
- [x] I have added relevant Storybook updates to this PR as well.
- [x] If this feature requires a developer doc update, tag @CareEvolution/api-docs.

Consider "Squash and merge" as needed to keep the commit history reasonable on `main`.

## Reviewers

Assign to the appropriate reviewer(s). Minimally, a second set of eyes is needed ensure no non-public information is published. Consider also including:
- Subject-matter experts
- Style/editing reviewers
- Others requested by the content owner

PR #114, opened 2023-09-18
ConnectDeviceAccountStep: Replace 'providerName' parameter with 'providerID'.

bxh:
## Overview

This PR changes the ConnectDeviceAccountStep to use an input of `providerName` to `providerID`.

While we test #104 in QA, I realized an implicit dependency about `providerName`, which is that the FHIR provider must have its name in its key word list. @DKaczmar tried to use Fitbit Staging provider but couldn't connect to the provider. The reason was that "Fitbit Staging" was not in the keyword list.

I picked `providerName` over `providerID` at the time of writing #104 because I thought the former would be more intuitive for users. But given the additional maintenance overhead, I think now it makes more sense to just use a `providerID` for the ConnectDeviceAccountStep, which actually was the way to go in the long covid (AOU) app. 

Screenshot 2023-09-18 150720

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
+++ b/src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
@@ -7,10 +7,11 @@ export default function () {
     const [text, setText] = useState<string>();
     const [styles, setStyles] = useState<any>({});
     const [deviceType, setDeviceType] = useState<string>("");
-    const [providerID, setProviderID] = useState<number>(0);
+    const [providerID, setProviderID] = useState<number>();
     const [connected, setConnected] = useState<boolean>();
 
     useEffect(() => {
+        // Get the step configuration from MyDataHelps.
         MyDataHelps.getStepConfiguration().then(function (
             config: StepConfiguration
         ) {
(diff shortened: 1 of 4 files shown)


src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
```diff
@@ -2,16 +2,12 @@ import React, { useState, useEffect } from "react";
...
-}
-
-export default function (props: ConnectDeviceAccountStepContainerProps) {
+export default function () {
     const [title, setTitle] = useState<string>();
     const [text, setText] = useState<string>();
     const [styles, setStyles] = useState<any>({});
     const [deviceType, setDeviceType] = useState<string>("");
-    const [providerName, setProviderName] = useState<string>(props.providerName ?? "");
+    const [providerID, setProviderID] = useState<number>(0);
```
chrisnowak, line 10: Would not use "0" as default, would use null to be more explicit

chrisnowak (approved):

PR #115, opened 2023-09-19
Connect Device Account Step: Remove polling and set providerID

bxh:
## Overview

This PR removes polling as it turns out to be unnecessary. Also set `providerID` based on `deviceType` and host name.

## Security

In the case of misconfigured providerID, a participant will just be shown a wrong yet secure provider login website.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
```diff
@@ -36,7 +36,7 @@ export default function () {
 
     useEffect(() => {
         // Start polling for connected status after a providerID is selected.
-        if (!providerID || connected) return;
+        if (!providerID || connected || !MyDataHelps.token) return;
```
ashleyrtate, line 39: Not sure I understand why this is needed. Doesn't getExternalAccounts() set the token?

chrisnowak: But yeah, as @ashleyrtate pointed out, I think the token actually would get set by the first call to getExternalAccounts() (because that is the first API request).

Actually, backing up, what happens if they visit this step and they've already connected Fitbit?  (I can't test it in QA right now since it seems to be complaining about missing providerID)

chrisnowak: We probably want to check if they have already connected the device and auto-proceed if so; so I think when this step loads you would need to call getExternalAccounts immediately...

bxh: 1. A _MyDataHelpsDesigner_ user add a ConnectDeviceAccountStep to a survey.
2. They set a providerID.
3. At this time, the rendered ConnectDeviceAccountStep step starts polling, which always fail due to lack of token.

This PR aims to address this issue.

ashleyrtate: I would set a "designMode" flag or something to make that explicit.

bxh: @chrisnowak @ashleyrtate ^

chrisnowak: I thought it should only start polling after the button is clicked?

bxh: No, this is the ConnectDeviceAccountStep.

> When the user visits this step it would check to see if they have the account connected; if so it would automatically proceed.

https://github.com/CareEvolution/Consumers/issues/11603

chrisnowak: Ah sorry I didn't notice this in the original PR - there was another issue I also was wondering about the other day:
- Seems like we should only check once when the step is loaded vs. always poll, right?
- Also...curious which / what platform is forcing us to poll anyways?  ConnectExternalAccount should return a promise so you should know when it is complete, no?

bxh: Polling is needed as we wait for the participant to connect to the device account. But you are right that this process can start **after** the participant clicks the button. In the initialization phase, we can just make one request.

bxh: (I'm working on the change)

chrisnowak: @bxh why is it needed?  We get a promise back when they click the button that resolves when they connect the account

bxh: Clicking the button only invokes `.connectExternalAccount`. That does not mean the participant is instantaneously connected, right? They still need to proceed to the third party website to sign in themselves.

chrisnowak: Right...after they do the connection the promise returned by connectExternalAccount resolves with a status (or fails if they cancel I think)

chrisnowak: e.g. you should be able to do something like
```
MyDataHelps.connectExternalAccount(1).then(() => {
   //stuff that should happen after connection is complete
});
```

bxh: Oh, I didn't realize that `connectExternalAccount` resolves _after_ a connection is complete.

bxh:
@chrisnowak I've refactored ConnectDeviceAccountStep based on your suggestion.

1. Removed polling. 
2. Moved all `MyDataHelps` stuff to the container component.

A follow up question, how should we handle errors when `connectExternalAccount` fails?

bxh:
On a related note, I'll remove polling from ConnectEhrStep using another PR.

src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
```diff
@@ -2,13 +2,26 @@ import React, { useState, useEffect } from "react";
...
-    const [providerID, setProviderID] = useState<number>();
-    const [connected, setConnected] = useState<boolean>();
+    const [deviceType, setDeviceType] = useState<string>(props.deviceType ?? "");
+    const [providerID, setProviderID] = useState<number>(props.providerID ?? 0);
+
+    function onConnect() {
+        MyDataHelps.connectExternalAccount(providerID, {
+            openNewWindow: true,
+        }).then(function() {
+            console.log("Completing step. Connected to provider ID", providerID);
```
chrisnowak, line 21: Remove console log before merging

src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
```diff
@@ -2,13 +2,26 @@ import React, { useState, useEffect } from "react";
...
+            MyDataHelps.completeStep("")
```
chrisnowak, line 22: Maybe put the providerID in here

src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
```diff
@@ -28,44 +41,36 @@ export default function () {
...
-            const accounts = await MyDataHelps.getExternalAccounts();
-            const containProvider = accounts.some(
-                (acc) => acc.provider.id === providerID
-            );
-            if (!connected && containProvider) {
-                setConnected(containProvider);
+        MyDataHelps.getExternalAccounts().then(function(accounts) {
+            const connected = accounts.some((acc) => acc.provider.id === providerID);
+            if (connected) {
+                console.log("Completing step. Already connected to provider ID", providerID);
```
chrisnowak, line 58: remove before merging

src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
```diff
@@ -28,44 +41,36 @@ export default function () {
...
+                MyDataHelps.completeStep("");
```
chrisnowak, line 59: Would put providerID in here actually

src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
```diff
@@ -2,13 +2,26 @@ import React, { useState, useEffect } from "react";
...
     const [title, setTitle] = useState<string>();
     const [text, setText] = useState<string>();
     const [styles, setStyles] = useState<any>({});
-    const [deviceType, setDeviceType] = useState<string>("");
-    const [providerID, setProviderID] = useState<number>();
-    const [connected, setConnected] = useState<boolean>();
+    const [deviceType, setDeviceType] = useState<string>(props.deviceType ?? "");
+    const [providerID, setProviderID] = useState<number>(props.providerID ?? 0);
+
+    function onConnect() {
```
chrisnowak, line 17: Why is this moved to the container?  

I think the idea with the container is that it should just be thin (no stories) and just get the config / styles and render the step...ultimately this pattern is a bit wonky though and I am regretting it in favor of having some sort of "previewState" or "designerMode" flag like the others.  But...maybe neither here nor there for now

chrisnowak:
Seems reasonable other than the comments I made...I think most of my issues are around the container pattern which is probably due for a refactor anyways.  🚢

chrisnowak:
Oh, but same comment as on other PR, I would make sure to test the connectExternalAccount promise flows on iOS/Android/Web to ensure it works consistently for the various states.  Obviously if the user cancels the connection nothing should happen so want to make sure we aren't proceeding in that case with the code as written or if there is some special handling it requires to detect that case

src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
bxh: I was under the impression that the containers are for behavior, whereas the UI components are just for presentation.

bxh:
@chrisnowak Can you approve the pull request?

src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
```diff
@@ -2,13 +2,34 @@ import React, { useState, useEffect } from "react";
...
+        MyDataHelps.connectExternalAccount(providerID, {
+            openNewWindow: true,
+        }).then(function () {
+            MyDataHelps.completeStep(providerID);
+        });
+    }
+
+    function isProd() {
+        return (
+            window.location.host.endsWith(".com") ||
```
chrisnowak, line 29: This isn't the right logic; you need to check MyDataHelps.baseUrl like we do here: https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/components/container/ConnectFitbit/ConnectFitbit.tsx#L18

Note that MyDataHelps.baseUrl won't be populated until you have a token - so you should probably invoke getExternalAccounts first

src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
```diff
@@ -2,13 +2,34 @@ import React, { useState, useEffect } from "react";
...
     const [title, setTitle] = useState<string>();
     const [text, setText] = useState<string>();
     const [styles, setStyles] = useState<any>({});
-    const [deviceType, setDeviceType] = useState<string>("");
-    const [providerID, setProviderID] = useState<number>();
-    const [connected, setConnected] = useState<boolean>();
+    const [deviceType, setDeviceType] = useState<string>(
+        props.deviceType ?? ""
+    );
+    const [providerID, setProviderID] = useState<number>(props.providerID ?? 0);
```
chrisnowak, line 17: Would drop the state param and prop since it's not needed, you can just get the providerID with a method

src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
```diff
@@ -25,47 +46,57 @@ export default function () {
...
+                            config.properties.deviceType
+                    );
             }
-            setProviderID(config.properties.providerID);
 
             setStyles(config.styles ?? {});
         });
     }, []);
 
     useEffect(() => {
```
chrisnowak, line 76: These separate useEffects aren't necessary anymore; would just chain getStepConfiguration() into getExternalAccounts()

src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
bxh: I'm including both https://mdhorg.ce.dev and https://mydatahelps.dev, since both of them are in use.

src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
```diff
@@ -2,20 +2,75 @@ import React, { useState, useEffect } from "react";
...
+
+    function isTest() {
+        return (
+            !MyDataHelps.baseUrl ||
+            MyDataHelps.baseUrl.startsWith("https://mdhorg.ce.dev") ||
+            MyDataHelps.baseUrl.startsWith("https://mydatahelps.dev")
+        );
+    }
+
+    function convertToProviderID(deviceType: string) {
```
chrisnowak, line 35: Will probably refactor this to use the new methods in https://github.com/CareEvolution/MyDataHelpsUI/pull/118 so this can be centralized

chrisnowak (approved):

src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
bxh: Fixed.

PR #122, opened 2023-10-11
Changes to SingleSurveyTask and SurveyTaskList

reneefromhold:
## Overview

Modified the SurveyTaskList and SingleSurveyTask to 
- Update the checkmark on completed tasks to match design referenced earlier
- Fixed Incomplete survey list not sorting Never Due surveys to the bottom of the list

## Security

- [ x] I have ensured no secure credentials or sensitive information remain in code, metadata, comments, etc.
  - [x ] Please verify that you double checked that .storybook/preview.js does not contain your participant access key details. 
  - [ x] There are no temporary testing changes committed such as API base URLs, access tokens, print/log statements, etc.
- [x ] These changes do not introduce any security risks, or any such risks have been properly mitigated.

This is an existing component. The functionality changed was mainly styles and a bug fix.

### Testing
- [x ] MyDataHelps iOS app
- [ TO DO  ] MyDataHelps Android app 
- [x ] [MyDataHelps website](mydatahelps.org)

### Documentation
- [x ] If this feature requires a developer doc update, tag @CareEvolution/api-docs.

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/container/SurveyTaskList/SurveyTaskList.css
```diff
@@ -1,3 +1,7 @@
+.mdhui-survey-task-list{
```
chrisnowak, line 1: This isn't a view but a component; it should not have a background

src/components/container/SurveyTaskList/SurveyTaskList.css
```diff
@@ -12,6 +16,10 @@
     padding-bottom: 0;
 }
 
+.mdhui-card-title {
+    color: rgba(60, 60, 67, 0.60)!important;
```
chrisnowak, line 20: Few issues with this:
- It will override the card title style for the entire library
- We should use variables for colors
- We should absolutely avoid using !important
- We should just make this a customization option

src/components/container/SurveyTaskList/SurveyTaskList.css
chrisnowak: I have a PR open so you can customize view background colors which should allow you to achieve this in the view builder

src/components/container/SurveyTaskList/SurveyTaskList.css
chrisnowak: I can tackle this as part of the color customization work i'm doing since i'm about to touch this

src/components/presentational/SingleSurveyTask/SingleSurveyTask.css
```diff
@@ -25,6 +25,7 @@
     {
         font-size: .88em;
         font-weight: bold;
+        width: 180px;
```
chrisnowak, line 28: Definitely wouldn't hardcode widths like this

src/components/presentational/SingleSurveyTask/SingleSurveyTask.css
```diff
@@ -62,3 +63,13 @@
     {
         cursor: pointer;
     }
+
+.button{
```
chrisnowak, line 67: I would leave the button width inconsistent like it is today; we have now had it designed both ways so unless it was really intended that button width be consistent here we should settle on one approach and I inclined to do the one that does not require changes.  I think the different width actually serves as a nice visual indicator that this row is different.  So I would revert this change

src/components/presentational/SingleSurveyTask/SingleSurveyTask.css
```diff
@@ -62,3 +63,13 @@
...
+    width: 90px;
+}
+
+.taskDetail {
```
chrisnowak, line 71: (I assume this change isn't needed if the button width stuff stays as-is)

PR #124, opened 2023-10-13
Add sticky option for date range navigator, fix Symptom Shark Calendar

chrisnowak:
## Overview

The symptom shark calendar view currently looks like this - the date range navigator is covering the filter buttons so you can only see a sliver of the bottom of them:
image.

Additionally, as you scrolled down, content was scrolling behind the status bar because the "top" was set to the safe area vs. the padding-top

This is a quick fix for this issue - since this is used in the device data pages I couldn't just adjust the margin / padding / top - so this basically makes those non-sticky (which is fine...I think it's a bit awkward they were)

While an existing issue, I would like to refactor this a bit more later to use portals so that there is a single stacked "sticky" context at the top of the page.  Right now this isn't sticky if you already have a navigation bar at the top, for instance (see the symptom detail view for an example).  I think having a portal with a single sticky spot at the top of the page would work though.

## Security

### Testing
- [x] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/presentational/DateRangeNavigator/DateRangeNavigator.css
+++ b/src/components/presentational/DateRangeNavigator/DateRangeNavigator.css
@@ -16,7 +16,7 @@
     position: sticky;
     position: -webkit-sticky;
     top: 0;
-    padding-top: calc(16px + env(safe-area-inset-top));
+    padding-top: calc(16px + env(safe-area-inset-top, 0px));
 }
 
 .mdhui-date-range-navigator .navigator-button {
--- a/src/components/presentational/DateRangeNavigator/DateRangeNavigator.css
+++ b/src/components/presentational/DateRangeNavigator/DateRangeNavigator.css
@@ -7,11 +7,16 @@
     text-align: center;
     font-weight: bold;
     border-bottom: solid 1px var(--mdhui-border-color-0);
-    position: sticky;
-    position: -webkit-sticky;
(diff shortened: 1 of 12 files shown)


src/components/presentational/DateRangeNavigator/DateRangeNavigator.css
```diff
@@ -7,11 +7,16 @@
     text-align: center;
     font-weight: bold;
     border-bottom: solid 1px var(--mdhui-border-color-0);
-    position: sticky;
-    position: -webkit-sticky;
-    top: env(safe-area-inset-top, 0px);
     background: var(--mdhui-background-color-0);
     z-index: 1;
+    position: relative;
```
bxh, line 12: Can you explain this line? It seems unnecessary to me.

bxh (approved):

src/components/presentational/DateRangeNavigator/DateRangeNavigator.css
chrisnowak: The left/right icons are absolutely positioned.  When I removed position:sticky I needed to add position:relative in order for them to continue to be absolutely positioned to this element.

bxh: Makes sense.

PR #148, opened 2023-11-09
Blood Pressure Visualization

reneefromhold:
## Overview

New Presentational Component - Dumbbell Chart
The dumbbell chart requires input parameters 1) axis info and 2)an array of dumbbells (the data)

New Container Component - Blood Pressure Visualization
The Blood Pressure Visualization displays a weeks worth of blood pressure data (week M-Sun) and uses color to draw attention to out of range readings. This visualization supports multiple entries per day. The long term goal is to display data from device, ehr, and survey, but for right now only survey data is supported. The input parameters are source (currently only Survey is supported), and survey name, step name, and result ids of systolic, diastolic, and date of entry.

## Security

  - [x ] Please verify that you double checked that .storybook/preview.js does not contain your participant access key details. 
  - [x ] There are no temporary testing changes committed such as API base URLs, access tokens, print/log statements, etc.
- [ x] These changes do not introduce any security risks, or any such risks have been properly mitigated.

The Dumbbell Chart is for displaying data provided to it. It is a presentational component
 The blood pressure visualization relies on the MyDataHelps api to authenticate ppt data. It will not work outside of storybook without a participant token. This behavior is identical to other components in MyDataHelpsUI

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation
- [x ] If this feature requires a developer doc update, tag @CareEvolution/api-docs.

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,257 @@
...
+                <div className="bpTitle">Blood Pressure</div>
+                <div>
+                    <div className="pagerRow">
+                        <DateRangeNavigator intervalType="Week" intervalStart={startOfWeek} onIntervalChange={pageWeek}></DateRangeNavigator>
+                    </div>
+                    <DumbbellChart dumbbells={weeklyDataForViz} axis={axis} ></DumbbellChart>
+                </div>
+                <div className="buffer"></div>
+                <div>
+                    <Button onClick={() => mdh.startSurvey(props.surveyName ?? "")}><span className="buttonCaption">Log Blood Pressure</span><img src={plus}/></Button>
```
reneefromhold, line 252: The bottom left border radius is different from the rest of the button curves and I'm not sure why. 
1) I question if the button should be part of this component
2) Is debouce needed here? I understood that it would get rid of blinking, but I still see blinking.

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,257 @@
...
+        return <LoadingIndicator/>;
+    }
+    else
+    {
+        return (
+            <div className="bloodPressureVisualization">
+                <div className="bpTitle">Blood Pressure</div>
+                <div>
+                    <div className="pagerRow">
+                        <DateRangeNavigator intervalType="Week" intervalStart={startOfWeek} onIntervalChange={pageWeek}></DateRangeNavigator>
```
reneefromhold, line 246: There is a border bottom on this that looks like a yaxis line, Any way to remove this?

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,257 @@
...
+import React from "react";
+import { Button, DateRangeNavigator, LoadingIndicator } from "../../presentational";
+import DumbbellChart from "../../presentational/DumbbellChart";
+import { Dumbbell, ClosedInterval, Axis, DataPoint, DumbbellClass } from "../../presentational/DumbbellChart/DumbbellChart";
+import { previewSurveyData } from './BloodPressureVisualization.previewdata'
+import { addDays, format, isEqual, isMonday, nextSunday, parseISO, previousMonday, startOfDay } from "date-fns";
+
+export type DataFaucet = "PreviewSurvey" | "Survey";
+
+export interface BloodPressureVisualizationProps {
```
reneefromhold, line 15: This same concept will be needed for the Blood Pressure Metrics component. Is there a best practice for sharing these? I didn't see existing components with this same situation

src/components/container/BloodPressureVisualization/BloodPressureVisualization.css
```diff
@@ -0,0 +1,41 @@
+@import url('https://fonts.googleapis.com/css2?family=Inter&display=swap');
+
+body{
```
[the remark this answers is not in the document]
chrisnowak: (I would remove this)

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+interface IBloodPressureDumbbell {
+    date: Date,
+    dumbbell : Dumbbell,
+    averageSystolic : number,
+    averageDiastolic : number
+};
+
+export default function (props : BloodPressureVisualizationProps) {
+    const mdh = MyDataHelps;
```
chrisnowak, line 39: Would just leave this as "MyDataHelps", you don't need to assign it to another var here - would be inconsistent with the rest of this

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,257 @@
...
+};
+
+export interface IBloodPressureDataPoint {
+    dateLabel: string,
+    date: Date,
+    systolic : number,
+    diastolic : number
+};
+
+interface IBloodPressureDumbbell {
```
chrisnowak, line 31: I would not prefix interfaces with "I" - that is not something you do in typescript.  Interface should just be "BloodPressureDummbell"

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
chrisnowak: I don't think that should be a separate component?  Seems like it should be a part of this one?

chrisnowak: Maybe not understanding what you are asking though

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+    function groupSurveyAnswersByResults(answers : SurveyAnswer[]){
+        let bpDataPoints : IBloodPressureDataPoint[] = [];
+
+        if (answers.length > 0){
+            let resultIds = [...new Set(answers.map(a => a.surveyResultID ))];
+            resultIds.forEach( (resultId) => {
+                var resultsForSubmission = answers.filter(a => a.surveyResultID == resultId);
+                var bpLogDateResults = resultsForSubmission.find( r => r.resultIdentifier == props.dateOfResultId);
+                var bpLogDate =bpLogDateResults?.answers[0];
```
chrisnowak, line 101: Minor - autoformat

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,257 @@
...
+                const dataPoint : DataPoint = {dataSet1: diastolicInterval, dataSet2: systolicInterval};
+                var db : Dumbbell = {dataPoint  :dataPoint, xValue : logDay, class : assignClass(systolicAverage, diastolicAverage)};
+                dbs.push({date: entriesForDate[0].date, dumbbell: db, averageSystolic : systolicAverage, averageDiastolic : diastolicAverage});
+            });
+        }
+
+        return dbs;
+    }
+
+    async function getSurveyAnswers(surveyName: string, stepId : string[], resultId : string[]): Promise<SurveyAnswer[]> {
```
chrisnowak, line 146: This is in the context of a component that has props that declare the step / result identifiers...you do not need to pass them around as parameters.  Just use the props

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
chrisnowak: Why do you need a specific "plus" icon here?  Does the font awesome plus icon not work?  Not a big deal either way but seems unnecessary to bring in another svg

src/components/presentational/DumbbellChart/DumbbellChart.tsx
```diff
@@ -0,0 +1,96 @@
+import "./DumbbellChart.css";
+import line from './horizontalLine.svg';
```
chrisnowak, line 2: definitely would NOT use an image just for a horizontal line?  Either SVG or a div with a border?

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,257 @@
...
+    if (loadingData) 
+    {
+        return <LoadingIndicator/>;
+    }
+    else
+    {
+        return (
+            <div className="bloodPressureVisualization">
+                <div className="bpTitle">Blood Pressure</div>
+                <div>
```
chrisnowak, line 244: Remove unnecesary div wrapper?

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
chrisnowak: This actually seems correct to have a bottom border?

Screenshot 2023-11-14 at 10 23 05 AM

That being said, the DateRangeNavigator should potentially have another variant if you want to achieve the look/feel you want here with the different blue buttons for left/right

src/components/presentational/DumbbellChart/DumbbellChart.tsx
```diff
@@ -0,0 +1,96 @@
...
+export interface Dumbbell {
+    dataPoint? : DataPoint;
+    xValue : string,
+    class? : DumbbellClass 
+}
+
+export interface Axis {
+    yRange : ClosedInterval;
+    yIncrement : number;
+    xIncrement : number;
```
chrisnowak, line 25: This is something that should be figured out automatically from the data and should not be specified in px...your chart needs to fill the screen and not just be anchored to the left side on large screens.  Would look at the other charts in this library and how they space out dynamically as you stretch the screen - basically x-axis things should be specified in % and not in px

This is a good example: https://ui.mydatahelps.org/?path=/story/symptomshark-presentational-overallexperiencechart--default

Screenshot 2023-11-14 at 10 27 44 AM

src/components/presentational/DumbbellChart/DumbbellChart.css
```diff
@@ -0,0 +1,104 @@
...
+ }
+
+ .dumbbell{
+    bottom: 0px;
+    height: 100%;
+    position: absolute;
+ }
+
+.inRange {
+    background-color: #6495ED;
```
chrisnowak, line 97: Let's use ```var(--mdhui-color-primary)``` instead of a new color

src/components/presentational/DumbbellChart/DumbbellChart.css
```diff
@@ -0,0 +1,104 @@
...
+    border-color:#6495ED;
+}
+
+.outOfRange {
+    background-color: #EF3724;
```
chrisnowak, line 102: Let's use ```var(--mdhui-color-danger)``` instead of a new color

src/components/presentational/DumbbellChart/DumbbellChart.tsx
```diff
@@ -0,0 +1,96 @@
...
+        for (var i = 0; i < (increments + 1); i++){
+            var key = `yAxis${i+1}`;
+            var yLabel = (i == 0) ? "" : yValue;
+            yAxis.push(<div className="yAxisTick" style={style} key={key}><img src={line}></img><div className="axisText">{yLabel}</div></div>);
+            yValue = (yValue + props.axis.yIncrement);
+            bottom = (i + 1) * props.axis.yIncrement;
+            style = {"bottom": `${(bottom/_range) * 100}%`};
+        }
+
+        setYAxisLines(yAxis);
```
chrisnowak, line 68: Definitely would not store react nodes in state...you should just build this dynamically on render.  e.g. return the yAxis from this method and call buildYAxis() below

src/components/presentational/DumbbellChart/DumbbellChart.tsx
```diff
@@ -0,0 +1,96 @@
...
+}
+
+export default function (props : DumbBellChartProps) {
+    const mdh = MyDataHelps;
+    const _minRange = props.axis.yRange.values[0];
+    const _maxRange = props.axis.yRange.values[1];
+    const _range = _maxRange - _minRange;
+
+    const [yAxisLines, setYAxisLines] = useState<any[]>([]);
+    const [chartDumbbells, setChartDumbbells] = useState<any[]>([]);
```
chrisnowak, line 45: This presentational component should need zero state...if you need a perf optimization (which you shouldn't) you could useMemo or something, but otherwise there's no reason to put stuff in state here

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,257 @@
...
+import { Button, DateRangeNavigator, LoadingIndicator } from "../../presentational";
+import DumbbellChart from "../../presentational/DumbbellChart";
+import { Dumbbell, ClosedInterval, Axis, DataPoint, DumbbellClass } from "../../presentational/DumbbellChart/DumbbellChart";
+import { previewSurveyData } from './BloodPressureVisualization.previewdata'
+import { addDays, format, isEqual, isMonday, nextSunday, parseISO, previousMonday, startOfDay } from "date-fns";
+
+export type DataFaucet = "PreviewSurvey" | "Survey";
+
+export interface BloodPressureVisualizationProps {
+    source : DataFaucet,
```
chrisnowak, line 16: Let's follow convention and just use "previewState" like other components to put in some preview data.  That needs to be independent of what data sources this is configured to use in reality anyways

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,257 @@
...
+    surveyName?: string,
+    stepIdentifier? : string,
+    dateOfResultId? : string,
+    systolicResultId? : string,
+    diastolicResultId? : string
```
chrisnowak, line 21: I would make this a lot less tied to surveys or any particular data source; instead I think this viz should take a set of "BloodPressureDataProviders".  There would just be 1 BloodPressureDataProvider for now - "SurveyBloodPressureDataProvider" - but it would enable us to add more in the future for Apple Health etc.

The Viz would just invoke the data provider; so all the querying of survey results would move out of this component and into the "SurveyBloodPressureDataProvider"

See for instance how I did this with "Daily data" providers: https://github.com/CareEvolution/MyDataHelpsUI/tree/main/src/helpers/daily-data-providers
https://github.com/CareEvolution/MyDataHelpsUI/pull/4

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+        if ((avgSystolic >=90 && avgSystolic <= 119) && (avgDiastolic >= 60 && avgDiastolic <= 79)){
+            return DumbbellClass.inRange;
+        }
+
+        return DumbbellClass.outOfRange;
+    }
+
+    useEffect(() => {
+        let debouncedInitialize = debounce(initialize, 500);
```
chrisnowak, line 225: Debouncing should not be necessary at all...

chrisnowak (changes requested):
Did a very initial surface-level pass but haven't yet gotten into the meat of the charts; I suspect there are more comments I will have there.  But I think plenty to address in the meantime

src/components/container/BloodPressureVisualization/BloodPressureVisualization.css
```diff
@@ -0,0 +1,41 @@
...
+    background: var(--bg-secondary, #FFFFFF)!important;
+}
+
+.bloodPressureVisualization {
+    width:  300px;
+    height: 100%;
```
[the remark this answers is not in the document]
reneefromhold: @chrisnowak When the component was larger than this, it did not display properly on android. I can change to 100%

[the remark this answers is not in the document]
chrisnowak: I meant the height:100%...I do not know what that accomplishes since I didn't see a fixed height on a container

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,303 @@
...
+import { FontAwesomeSvgIcon } from 'react-fontawesome-svg-icon';
+import { faPlus } from '@fortawesome/free-solid-svg-icons';
+import { BloodPressureDataParameters, BloodPressureDataPoint, BloodPressureDataProvider } from "../../../helpers/blood-pressure-data-providers/query-blood-pressure";
+
+export enum Category {"Low", "Normal", "Elevated", "Stage 1", "Stage 2", "Crisis", "Unknown"};
+export type BloodPressurePreviewState = "WithData" | "NoData" | "Loading" | "Live";
+
+export interface BloodPressureVisualizationProps {
+    previewState? : BloodPressurePreviewState,
+    previewDataProvider : BloodPressureDataProvider,
```
chrisnowak, line 18: Actually wouldn't have a previewDataProvider here but rather just have the component itself determine the data provider based on the previewState.  

Otherwise in the view builder we will have to supply this as well which seems unnecessary

chrisnowak: Basically I think "previewState" means we do not even invoke a data provider but instead just populate whatever data makes sense to render that previewState

src/helpers/blood-pressure-data-providers/query-blood-pressure.ts
```diff
@@ -0,0 +1,26 @@
+import surveyBloodPressureDataProvider, { BloodPressureDataParameters as SurveyBloodPressureDataParameters } from "./survey-blood-pressure-data-provider";
+
+export interface BloodPressureDataPoint {
+    dateLabel: string,
```
chrisnowak, line 4: Remove dateLabel here and just compute it when rendering

src/helpers/blood-pressure-data-providers/query-blood-pressure.ts
```diff
@@ -0,0 +1,26 @@
...
+    date: Date,
+    systolic : number,
+    diastolic : number
+};
+
+export type BloodPressureDataSource = "Survey";
+
+export interface BloodPressureDataParameters {
```
chrisnowak, line 12: See comment on structure; I think we can remove this class

src/helpers/blood-pressure-data-providers/survey-blood-pressure-data-provider.ts
```diff
@@ -0,0 +1,81 @@
+import MyDataHelps, { Guid, SurveyAnswer, SurveyAnswersPage, SurveyAnswersQuery } from "@careevolution/mydatahelps-js";
+import { BloodPressureDataPoint } from "./query-blood-pressure";
+import { format, parseISO, startOfDay, toDate } from "date-fns";
+
+export interface BloodPressureDataParameters {
+    surveyName: string,
+    dateOfResultId? : string,
```
chrisnowak, line 7: These should be suffixed "ResultIdentifier" for clarity, e.g. "systolicResultIdentifier"

chrisnowak: Also doesn't make a ton of sense that this is called "dateOfResultIdentifier", I would shorten to "dateResultIdentifier"

src/helpers/blood-pressure-data-providers/survey-blood-pressure-data-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+        let dataPage = await getSurveyDataPage(surveyName, resultId);
+        let allData = dataPage.surveyAnswers;
+        while (dataPage.nextPageID) {
+            dataPage = await getSurveyDataPage(surveyName, resultId, dataPage.nextPageID);
+            allData = allData.concat(dataPage.surveyAnswers);
+        }
+        return allData;
+    }
+
+    async function getSurveyDataPage(surveyName: string, resultId : string[], pageID?: Guid): Promise<SurveyAnswersPage> {
```
chrisnowak, line 68: You can skip taking surveyName / resultId as parameters because this is nested inside a function which already has them as parameters

chrisnowak: Would also use "resultIdentifier" instead of "resultId" everywhere

src/helpers/blood-pressure-data-providers/survey-blood-pressure-data-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+        let bpDataPoints : BloodPressureDataPoint[] = [];
+
+        if (answers.length > 0){
+            let resultIds = [...new Set(answers.map(a => a.surveyResultID ))];
+            resultIds.forEach( (resultId) => {
+                var resultsForSubmission = answers.filter(a => a.surveyResultID == resultId);
+                var bpLogDateResults = resultsForSubmission.find( r => r.resultIdentifier == props.dateOfResultId);
+                var bpLogDate =bpLogDateResults?.answers[0];
+                var bpSystolicResults = resultsForSubmission.find( r => r.resultIdentifier == props.systolicResultId);
+                var bpSystolic = bpSystolicResults?.answers[0];
```
chrisnowak, line 36: Good to guard against answers being an empty array, empty string, or invalid (e.g. non-numeric for systolic / diastolic, non-date for date)

src/helpers/blood-pressure-data-providers/survey-blood-pressure-data-provider.ts
```diff
@@ -0,0 +1,81 @@
+import MyDataHelps, { Guid, SurveyAnswer, SurveyAnswersPage, SurveyAnswersQuery } from "@careevolution/mydatahelps-js";
+import { BloodPressureDataPoint } from "./query-blood-pressure";
+import { format, parseISO, startOfDay, toDate } from "date-fns";
+
+export interface BloodPressureDataParameters {
+    surveyName: string,
+    dateOfResultId? : string,
+    systolicResultId? : string,
+    diastolicResultId? : string
```
chrisnowak, line 9: Why are these optional?  Seems like they should be required

src/helpers/blood-pressure-data-providers/survey-blood-pressure-data-provider.ts
chrisnowak: e.g. probably ignore/skip the result if it's actually invalid

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,303 @@
...
+        diastolicWeeklyAvg = Math.round(diastolicWeeklyAvg);
+        let diastolicWeeklyAvgAlert = getDiastolicCategory(diastolicWeeklyAvg);
+
+        let systolicReadings = dbs.map(db => db.averageSystolic);
+        let systolicWeeklyAvg = systolicReadings.reduce((a, b) => a + b) / systolicReadings.length;
+        systolicWeeklyAvg = Math.round(systolicWeeklyAvg);
+        let systolicWeeklyAvgAlert = getSystolicCategory(systolicWeeklyAvg);
+
+        let bpMetrics : BloodPressureMetrics = {
+            averageSystolic : systolicWeeklyAvg,
```
chrisnowak, line 138: Aren't you averaging an average here?  Is that what is intended?  e.g. What if someone took 20 readings on day 1 and then 1 reading on day 2...is it supposed to be the average of all 21 readings

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+        if (!loadingData){
+            setLoadingData(true);
+
+            if (props.previewState && props.previewState == "Loading"){
+                return;
+            }
+            else 
+            {
+                props.previewDataProvider(props.dataParameters).then((bloodPressureDataPoints) => {
```
chrisnowak, line 67: I think I am confused by the data provider setup here...why is this called the preview data provider, is it not supplying real data in a real use case?

In any case, I think as mentioned before I think the props can just have a SurveyBloodPressureDataParameters and we can just call the SurveyBloodPressureDataProvider from here

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,303 @@
...
+}
+
+export default function (props : BloodPressureVisualizationProps) {
+    const _minDiastolic = 0;
+    const _maxSystolic = 250; 
+    const yInterval : ClosedInterval = {values: [_minDiastolic, _maxSystolic]};
+    const axis : Axis = {yRange : yInterval, yIncrement : 50, xIncrement : (85/7)};
+
+    const [loadingData, setLoadingData] = useState(false);
+    const [bpData, setBpData] = useState<BloodPressureDumbbell[]>([]);
```
chrisnowak, line 52: So thinking about what you actually need for state...you don't need to be holding onto this much state.

I would have 2 state variables:

```
const [bloodPressureData, setBloodPressureData] = useState<BloodPressureDataPoint[] | null>(null);
const [startOfWeek, setStartOfWeek] = useState<Date>(initializeWeekPager());
```

You can compute everything else based on those 2 state variables.

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,303 @@
...
+            averageSystolicAlertClass : systolicWeeklyAvgAlert == Category.Normal ? "mdhui-blood-pressure-metric-normal" : "mdhui-blood-pressure-metric-not-normal",
+            averageDiastolic : diastolicWeeklyAvg,
+            averageDiastolicAlert: Category[diastolicWeeklyAvgAlert],
+            averageDiastolicAlertClass : diastolicWeeklyAvgAlert == Category.Normal ? "mdhui-blood-pressure-metric-normal" : "mdhui-blood-pressure-metric-not-normal",
+            minDiastolicAlertClass : "",
+            maxSystolicAlertClass : ""
+        };
+
+        let diastolicLows = dbs.map(db => db.dumbbell.dataPoint ? db.dumbbell.dataPoint.dataSet1.values[0] : Infinity);
+        let diastolicLow = diastolicLows ? Math.min(...diastolicLows) : Infinity;
```
chrisnowak, line 149: Why "infinity" as a default here and not null or undefined?

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
chrisnowak: (Loading can be inferred by a null bloodPressureData)

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        if (diastolicValue > 120){
+            return Category.Crisis;
+        }
+
+        return Category.Unknown;
+    }
+    
+    function initializeWeekPager() : Date {
```
chrisnowak, line 253: I think you can drop this and just call "getWeekStart" from here: https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/helpers/get-interval-start.ts#L15 with "Monday"

Though I would consider making the "weekStartsOn" a prop like this: https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/components/presentational/DateRangeCoordinator/DateRangeCoordinator.tsx#L9

src/components/presentational/DumbbellChart/DumbbellChart.css
```diff
@@ -0,0 +1,103 @@
+.mdhui-dumbbell-visualization {
+    padding-top: 25px;
```
chrisnowak, line 2: This padding-top seems wrong based on the design.  The 250 should be right below the date range navigator.

Just removing it does cause a double border so you could also add margin-top:-1px or make the 250 tick just not have a border on it or something

src/components/presentational/DumbbellChart/DumbbellChart.css
```diff
@@ -0,0 +1,103 @@
...
+.mdhui-dumbbell-axis-text{
+    color: #BFBFBF;
+    text-align: center;
+    font-family: 'Inter';
+    font-size: 10px;
+    font-style: normal;
+    font-weight: 700;
+    line-height: normal;
+    display: flex;
+    justify-content: right;
```
chrisnowak, line 39: In the design there is some space between the right side:
Screenshot 2023-11-27 at 4 45 48 PM

That is missing here:
Screenshot 2023-11-27 at 4 45 00 PM

chrisnowak:
Have to sign off for the day but still a good number of things to tackle here; will aim to complete review tomorrow

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
```diff
@@ -0,0 +1,303 @@
...
+import { faPlus } from '@fortawesome/free-solid-svg-icons';
+import { BloodPressureDataParameters, BloodPressureDataPoint, BloodPressureDataProvider } from "../../../helpers/blood-pressure-data-providers/query-blood-pressure";
+
+export enum Category {"Low", "Normal", "Elevated", "Stage 1", "Stage 2", "Crisis", "Unknown"};
+export type BloodPressurePreviewState = "WithData" | "NoData" | "Loading" | "Live";
+
+export interface BloodPressureVisualizationProps {
+    previewState? : BloodPressurePreviewState,
+    previewDataProvider : BloodPressureDataProvider,
+    dataParameters : BloodPressureDataParameters
```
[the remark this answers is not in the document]
reneefromhold: @chrisnowak Just to be clear, is there a preference to directly call the survey specific data provider I'd created, or keep using the common entry point https://github.com/CareEvolution/MyDataHelpsUI/blob/renee/BP/src/helpers/blood-pressure-data-providers/query-blood-pressure.ts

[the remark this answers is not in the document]
chrisnowak: Yeah i'd just call the survey specific data provider.  It is good to keep the fetching of the data separate from the display still, but I think the common entry point doesn't make a lot of sense right now

src/components/container/BloodPressureVisualization/BloodPressureVisualization.css
```diff
@@ -0,0 +1,79 @@
...
+    padding-bottom: 10px;
+    display: flex;
+    flex-direction: column;
+    width: 100%;
+}
+
+.mdhui-blood-pressure-metrics-rows{
+    display: flex;
+    flex-direction: row;
+    background-color: #FFFFFF;
```
chrisnowak, line 22: This shouldn't need a background color and if it did, would use the variable for this so it works in dark mode

src/components/container/BloodPressureVisualization/BloodPressureVisualization.css
```diff
@@ -0,0 +1,79 @@
...
+    height: 50%;
+}
+
+.mdhui-blood-pressure-metrics-block{
+    display: flex;
+    flex-direction: column;
+    padding: 16px;
+    border-radius: 6px;
+    border-color:#EDEDED;
```
chrisnowak, line 31: Would use a border color variable from the globals for this - again so it works in dark mode

src/components/presentational/DumbbellChart/DumbbellChart.css
```diff
@@ -0,0 +1,103 @@
...
+}
+
+.mdhui-dumbbell-pill{
+    position: absolute;
+    width: 9px;
+    border-radius: 25px;
+    border-style: solid;
+    border-width: 2px;
+    z-index: 1;
+    background-color: #FBFCFE;
```
chrisnowak, line 50: Should this be var(--mdhui-color-primary);

src/components/presentational/DumbbellChart/Dumbbell.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    function getStyle(height : any, bottom : any){
+       return {"height": height, "bottom": bottom};
+    }
+
+    return (
+        <div className="mdhui-dumbbell" style={leftAsPercent} >
+            {props.dumbbell?.dataPoint && 
+                <>
+                {dataSet2OutOfRange && <div className="mdhui-dumbbell-out-of-range-icon mdhui-dumbbell-top"></div>}
```
chrisnowak, line 60: Curious why there is special logic for when something is "out of range"...I think the chart should cover all reasonable ranges for blood pressure no?   Seems like we would just want to have the lines be cut off at the top?

src/components/presentational/DumbbellChart/DumbbellChart.css
```diff
@@ -0,0 +1,103 @@
...
+.mdhui-dumbbell-y-axis{
+    position: absolute;
+    flex-flow: nowrap;
+    display: flex;
+    flex-direction: column;
+    justify-content:flex-end;
+    width: 100%;
+    border-top-width: 1px;
+    border-top-style: solid;
+    border-top-color: #E7EFFF;
```
chrisnowak, line 22: Should be a border color from the global vars...would almost never hardcode colors

src/components/presentational/DumbbellChart/DumbbellChart.css
```diff
@@ -0,0 +1,103 @@
...
+}
+
+.mdhui-dumbbell-x-axis{
+    position: absolute;
+    bottom: -20px;
+}
+
+.mdhui-dumbbell-axis-text{
+    color: #BFBFBF;
```
chrisnowak, line 31: Text color should be from vars

src/components/container/BloodPressureVisualization/BloodPressureVisualization.css
```diff
@@ -0,0 +1,79 @@
...
+    width: 50%;
+}
+
+.mdhui-blood-pressure-metric-description{
+    font-size: 12px;
+    font-weight: 500;
+    line-height: 15px;
+    letter-spacing: 0em;
+    text-align: left;
+    color: #585858;
```
chrisnowak, line 43: Use one of the text color vars

src/components/container/BloodPressureVisualization/BloodPressureVisualization.css
```diff
@@ -0,0 +1,79 @@
...
+    color: var(--mdhui-color-danger);
+}
+
+.mdhui-blood-pressure-metric-value{
+    font-size: 24px;
+    font-weight: 600;
+    line-height: 29px;
+    letter-spacing: 0em;
+    text-align: left;
+    color: #000000;
```
chrisnowak, line 70: Use one of the text color vars

src/components/container/BloodPressureVisualization/BloodPressureVisualization.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+export default {
+	title: "Container/BloodPressureVisualization",
+	component: BloodPressureVisualization,
+	parameters: {
+		layout: 'fullscreen',
+	}
+} as ComponentMeta<typeof BloodPressureVisualization>;
+
+const Template: ComponentStory<typeof BloodPressureVisualization> = (args: BloodPressureVisualizationProps) =>
+	<Layout>
```
chrisnowak, line 18: I would add colorScheme="auto" to this story and make sure it works in dark mode (you can change your computer to dark mode fairly easily)...in particular will ensure we use the various CSS variables for text / border colors since those are set up for light/dark mode

src/components/presentational/DumbbellChart/Dumbbell.tsx
reneefromhold: We iterated over a few ways to present out of range, and this was the design that was settled on.

src/helpers/blood-pressure-data-providers/query-blood-pressure.ts
reneefromhold: I did remove this as a field, and am doing the format not on the dumbbell render, but on the creation of the xaxis text for the dumbbell

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
reneefromhold: Yes, averaging the daily averages where a daily average exist.
In the example you provided, (Day1Average + Day2Average)/totalDaysWhereDataExists

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
reneefromhold: I am having a tough time getting the weeklyData and weeklyMetrics out of state.

chrisnowak: I would think about it this way...you should be able to create a separate standalone method which takes in 2 parameters (startOfWeek and bloodPressureData) - and it should be able to return weeklyData/weeklyMetrics.  Unless there is some other dependency i'm overlooking?

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
reneefromhold: I preferred to stick with the number data type vs any when using undefined.
I ended up removing a lot of this.

chrisnowak: it isn't "any" though, it would just be ```number | undefined``` or ```number | null```

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
reneefromhold: I was able to reduce like suggested. Thank you for this feedback

src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx
reneefromhold: Math.min and Math.max were not happy being supplied number | undefined. 
I ended up filtering out the non numeric values, which helped keep the array purely numeric.

src/components/presentational/DumbbellChart/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from "./DumbbellChart";
```
chrisnowak, line 1: You'll also need to add exports to the presentation/index.ts for this to actually be exported from the library

src/components/container/BloodPressureVisualization/index.ts
```diff
@@ -0,0 +1 @@
+export {default} from "./BloodPressureVisualization";
```
chrisnowak, line 1: You'll also need to add exports to the container/index.ts for this to actually be exported from the library

src/helpers/blood-pressure-data-providers/survey-blood-pressure-data-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    return groupSurveyAnswersByResults(sortedAnswers);
+
+
+    function groupSurveyAnswersByResults(answers : SurveyAnswer[]){
+        let bpDataPoints : BloodPressureDataPoint[] = [];
+
+        if (answers.length > 0){
+            let resultIds = [...new Set(answers.map(a => a.surveyResultID ))];
```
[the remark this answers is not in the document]
chrisnowak: Minor, but would still be nice to rename this var

chrisnowak (approved):
Overall looks good - just a few final comments, but this seems good to go now

chrisnowak:
Per the original design, I think it would make sense to still include the "250" tick mark, but don't feel strongly about it.  Otherwise the latest changes / storybook looks good

PR #156, opened 2023-11-16
Fix baseUrl race condition in providerID helpers

ajmenca:
## Overview

There is a race condition with `MyDataHelps.baseUrl` not being set yet when `getFitbitProviderID` or `getGarminProviderID` are called through either the `ConnectFitbit` or `ConnectGarmin` widgets when use on the `FitbitView` or `GarminView`.  This is because no other `MyDataHelps` API call happens before `getFitbitProviderID` or `getGarminProviderID`, so `baseUrl` isn't set because no access token has been parsed yet.

The solution I'm providing here is the most straight forward.  This fix targets the views directly affected by this race condition, which does leave open the possibility for this to crop up again.  I'm certainly open to other ideas for fixing this; the only other idea I had was to push the `connect` call into the providerID getters directly - which would require them to return a `Promise`, which has more downstream effects and a lot more thread pulling than seemed necessary at this point in time.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/view/FitbitView/FitbitView.tsx
+++ b/src/components/view/FitbitView/FitbitView.tsx
@@ -1,8 +1,9 @@
-import React from 'react'
-import { Layout, Card, StatusBarBackground, ConnectFitbit, FitbitDevices, FitbitMonthCharts } from "../.."
+import React, { useEffect, useState } from 'react'
+import { Layout, Card, StatusBarBackground, ConnectFitbit, FitbitDevices, FitbitMonthCharts, LoadingIndicator } from "../.."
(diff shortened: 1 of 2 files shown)


chrisnowak (approved):
Originally I thought that maybe we should do this on a component level, but upon reflection this feels like the right place since this is a pretty narrow issue.  The view builder for instance populates a context with ppt info before rendering any components so it wouldn't be an issue there, and it feels wrong to make the individual components care about this

PR #161, opened 2023-11-27
Add a simple calendar implementation that supports streak highlighting.

greinard:
## Overview

This branch adds a simple calendar implementation that supports streak highlighting.

`SimpleCalendar` wraps the existing `Calendar` component, providing a default and state-based day rendering implementation.

It works by using a state configuration mapping (which maps a given state to a style and streak highlighting settings) as well as a function to compute the state for a given day.

image

## Security

- No new security risks have been introduced.  This update just provides a default implementation for calendar day rendering that allows for some state based styling customization.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation


src/components/presentational/SimpleCalendar/SimpleCalendar.css
```diff
@@ -0,0 +1,35 @@
+.mdhui-simple-calendar {
+    background: #fff;
```
chrisnowak, line 2: Would remove background; it is up to the card or container what the background is typically...and if we really need a contrast we would want to use the css background vars anyways so it works in dark mode

src/components/presentational/SimpleCalendar/SimpleCalendar.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export interface SimpleCalendarProps {
+    intervalStart?: Date;
+    stateConfiguration: SimpleCalendarStateConfiguration;
+    computeStateForDay: (date: Date) => string;
+    onDayClicked: (date: Date) => void;
+    innerRef?: React.Ref<HTMLDivElement>;
+}
+
+export default function (props: SimpleCalendarProps) {
```
chrisnowak, line 17: I would go with "StreakCalendar" I think.  I think the novelty here is in streaks...you could theoretically use this without streaks...but it would not be massive benefit necessarily...the reason that we would build this as opposed to some sort of standalone "CalendarDay" component is to wrap the logic of streaks

src/components/presentational/SimpleCalendar/SimpleCalendar.tsx
```diff
@@ -0,0 +1,68 @@
...
+            } else if (currentDayState === nextDayState && date.getDay() < 6 && date.getDate() < getLastDayOfMonth(date)) {
+                dayClasses.push('mdhui-simple-calendar-day-streak-right');
+            }
+
+            if (currentDayStateConfiguration?.streakColor) {
+                dayStyle = {'--mdhui-simple-calendar-streak-color': resolveColor(layoutContext.colorScheme, currentDayStateConfiguration?.streakColor)} as CSSProperties
+            }
+        }
+
+        return <div className={dayClasses.join(' ')} style={dayStyle} onClick={() => props.onDayClicked(date)}>
```
chrisnowak, line 60: This could be a separate component for "CalendarDay" that takes props for "streakLeft", "streakRight", "day",  "backgroundColor", "streakColor"

greinard:
@chrisnowak - I ended up reworking this a bit after seeing your suggestions.  I created a new `CalendarDay` component, but I went a step further and just moved all of the streaking logic into that.

We were going to need project specific calendar impls to define the state/styling config anyways, so it was already a bit awkward to have an additional "SomeKindOfCalendar" component in the hierarchy.  With this update, the project specific calendar impls can just define their state/styling configuration and state computing function and then use the new `CalendarDay` component to render the days, like so:

```
const renderDay = (year: number, month: number, day?: number): React.JSX.Element => {
	return <CalendarDay
		year={year}
		month={month}
		day={day}
		stateConfiguration={stateConfiguration}
		computeStateForDay={computeStateForDay}
		onClick={onDayClicked}
	/>;
};
```

In additional to some standard stories to show how the `CalendarDay` component renders on its own, I have also added a story that shows how to use it with the existing `Calendar` component.

src/components/presentational/SimpleCalendar/SimpleCalendar.css
greinard: This is gone now as part of the PR revisions.

src/components/presentational/CalendarDay/CalendarDay.stories.tsx
```diff
@@ -0,0 +1,254 @@
...
+    parameters: {layout: 'fullscreen'}
+};
+
+const render = (args: CalendarDayProps) => <Layout colorScheme="auto">
+    <div style={{width: '60px', margin: '0 auto'}}>
+        <CalendarDay {...args} />
+    </div>
+</Layout>;
+
+const now = parseISO('2023-11-29T21:41:43.678Z');
```
reneefromhold, line 18: nit: use today

src/components/presentational/CalendarDay/CalendarDay.tsx
```diff
@@ -0,0 +1,62 @@
...
+    let currentDayState = props.computeStateForDay(date);
+    let nextDayState = props.computeStateForDay(add(new Date(date), {days: 1}));
+
+    let currentDayStateConfiguration = props.stateConfiguration[currentDayState];
+
+    let dayClasses = ['mdhui-calendar-day'];
+    let dayStyle: CSSProperties | undefined;
+    if (currentDayStateConfiguration?.streak) {
+        if (currentDayState === previousDayState && currentDayState === nextDayState) {
+            if (date.getDate() > 1 && date.getDay() > 0 && date.getDay() < 6 && date.getDate() < getLastDayOfMonth(date)) {
```
reneefromhold, line 42: nit, this would read easier if there were variables for isFirstDayOfMonth, isLastDayOfWeek etc instead of date.getDate() > 1, date.getDay() < 6

src/components/presentational/DateRangeCoordinator/index.ts
```diff
@@ -1 +1 @@
-export { default } from "./DateRangeCoordinator";
\ No newline at end of file
+export { default, DateRangeContext } from "./DateRangeCoordinator";
```
reneefromhold, line 1: Nice add!

reneefromhold (approved):

src/components/presentational/CalendarDay/CalendarDay.stories.tsx
greinard: I renamed it to `testDate`.  It doesn't really represent "now" or "today".  It used to be `new Date()`, which worked for "now", but I had to set it to a fixed date.  Otherwise, depending on which day of the week/month it was, the story renderings would be different.  Now they are static.

src/components/presentational/CalendarDay/CalendarDay.tsx
greinard: Good call.  I reworked the logic a bit to simplify it and pulled out a couple of helper methods and more human readable variables.

chrisnowak (approved):
Looks good and the changes make a ton of sense

PR #163, opened 2023-11-30
Prevent duplicate task survey launches.

### LINKED ISSUE #139: Survey tasks can be re-clicked after being completed when the list refresh is slow or delayed.
We have seen a few instances where folks appear to have been able to re-click a survey task even after that task has been completed due to a delay with refreshing of the task list.  This delay could be due to either network or app latency, but it allows a user to re-launch the survey when they should not be able to.

I suggest we disable clicking on a survey task list item when it has been clicked.  Launching a survey always results in a survey task list refresh when the survey gets completed or canceled, so it should be safe to leave the item "disabled" until the list has a chance to refresh.

We may want to provide some sort of visual indication on the disabled item (a tiny spinner, perhaps) to communicate that something is going on in the background during any refresh latency.  Otherwise, the user may think the UI is just being unresponsive to clicks for no reason.

greinard:
## Overview

Fixes #139

This update is meant to prevent duplicate survey launches for tasks which have not yet had a chance to be cleared from the task list.  This can happen if there is some latency between the completion of a survey task and the refreshing of the task list.  This change effectively makes the `SingleSurveyTask` a single use widget.  That is to say, once it has been clicked, it will not be clickable again until the survey task list has been refreshed.  This should already happen any time a survey gets completed or canceled.

## Security

- No new security risks.  This update just prevents double launching of a survey task.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@careevolution/mydatahelps-ui",
-  "version": "2.9.1-Issue139.1",
+  "version": "2.9.1-Issue139.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@careevolution/mydatahelps-ui",
-      "version": "2.9.1-Issue139.1",
+      "version": "2.9.1-Issue139.2",
       "license": "MIT",
       "dependencies": {
         "@emotion/react": "11.11.1",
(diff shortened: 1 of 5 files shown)


greinard:
Hmm.  During some further testing, I'm not seeing the survey task list refresh as I had expected.  Digging into that further now.

greinard:
I think I'm going to need to take a different approach to fix this.  Closing this PR.  I'll open another with the updated solution.

PR #164, opened 2023-11-30
Prevent duplicate task survey launches.

### LINKED ISSUE #139: Survey tasks can be re-clicked after being completed when the list refresh is slow or delayed.

greinard:
## Overview

Fixes #139 

Here is another approach at fixing the issue with duplicate task survey launches.  As described previously, there are scenarios where a delay in refreshing the task list can allow a participant to launch a task survey more than once, even when the task has already been completed.

In order for refreshes to work correctly, management of the "active surveys" had to be moved up to the list itself.  This also meant that I had to move the click handling up to the list as well.  The single survey task no longer launches the survey directly. Instead, it has a click handler property and a flag for whether the task's survey is currently active (so that the spinner can be displayed).

The combination of these changes results in the clicked survey task displaying a spinner while the survey is active and then that spinner gets removed (and clicking gets re-enabled) once the task list has had a chance to refresh.

image

## Security

- No new security implications.  This change just prevents relaunching of completed task surveys.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

### DIFF
--- a/src/components/container/SurveyTaskList/SurveyTaskList.tsx
+++ b/src/components/container/SurveyTaskList/SurveyTaskList.tsx
@@ -7,6 +7,7 @@ import { previewCompleteTasks, previewIncompleteTasks } from './SurveyTaskList.p
 import language from '../../../helpers/language'
 import { ColorDefinition, resolveColor } from '../../../helpers/colors'
 import { ButtonVariant } from '../../presentational/Button/Button'
+import { useInitializeView } from '../../../helpers/Initialization';
 
 export interface SurveyTaskListProps {
 	status: SurveyTaskStatus,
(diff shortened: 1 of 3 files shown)


src/components/container/SurveyTaskList/SurveyTaskList.tsx
```diff
@@ -28,19 +29,30 @@ export type SurveyTaskListListPreviewState = "IncompleteTasks" | "CompleteTasks"
 
 export default function (props: SurveyTaskListProps) {
 	const [loading, setLoading] = useState(true);
-	const [tasks, setTasks] = useState<SurveyTask[] | null>(null);
+	const [tasks, setTasksInner] = useState<SurveyTask[] | null>(null);
+	const [activeSurveys, setActiveSurveys] = useState<string[]>([]);
```
chrisnowak, line 33: You should never be able to have more than 1 active survey...you can't launch them on top of each other

chrisnowak: Oh...but I guess in a very very slow loading scenario I see how this could happen

src/components/container/SurveyTaskList/SurveyTaskList.tsx
```diff
@@ -28,19 +29,30 @@ export type SurveyTaskListListPreviewState = "IncompleteTasks" | "CompleteTasks"
...
-		initialize()
-		MyDataHelps.on("applicationDidBecomeVisible", initialize);
-		return () => {
-			MyDataHelps.off("applicationDidBecomeVisible", initialize);
+	const setTasks = (tasks: SurveyTask[]): void => {
+		setTasksInner(tasks);
+		setActiveSurveys([]);
+	};
+
+	useInitializeView(initialize, [], [props.previewState]);
```
chrisnowak, line 41: Oh wow, I totally missed that you added useInitializeView.  I actually added a very similar hook which I should probably get rid of, since it is just less good: https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/helpers/useMyDataHelps.ts

src/components/container/SurveyTaskList/SurveyTaskList.tsx
```diff
@@ -28,19 +29,30 @@ export type SurveyTaskListListPreviewState = "IncompleteTasks" | "CompleteTasks"
...
+
+	const isSurveyActive = (task: SurveyTask): boolean => {
+		return activeSurveys.includes(task.surveyName);
+	};
+
+	const onTaskClicked = (task: SurveyTask) => {
+		if (!loading && !isSurveyActive(task)) {
```
chrisnowak, line 48: I am not sure that the check for "loading" does anything actually, does it?  Was probably something I erroneously added in the first version of this.  There would be no tasks if this was loading so nothing to click...

chrisnowak (approved):
Few comments but looks good overall

src/components/container/SurveyTaskList/SurveyTaskList.tsx
greinard: I removed the `!loading` check from the task click handler since `loading` will only be `true` when there are no tasks yet.

PR #166, opened 2023-12-07
Add Calorie queries for Garmin and Apple Health

ajmenca:
## Overview

This PR expands the available predefined data queries with Calorie queries for Apple Health and Garmin (to bring them closer in line with available Fitbit queries).

  - Garmin exposes two calorie counts in the daily summary.  Active and Basal Metabolic Rate (BMR).  The query for BMR Calories is probably not interesting, as it is (mostly) an unchanging precalculated value based on the user's vital statistics.  I added a 3rd query for Total Calories that will add the active and bmr values together, as this total value is more similar to the existing Fitbit Calorie query which does appear to include both values.
  - Apple Health appears to not have a "Base" calorie representation in the daily summary that we're storing.  So there's only an Active calorie query here.  This will be a discrepancy in available representation across devices.

This also fixes a dependency error in the GarminView widget; the LoadingIndicator was not being included properly and was resulting in an error in the Storybook because the widget doesn't compile correctly.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/helpers/daily-data-providers/query-garmin-total-daily-values.ts
```diff
@@ -0,0 +1,46 @@
...
+				return Math.round(parseFloat(d.properties[type]));
+		}
+
+		result.forEach((d) => {
+			types.forEach(type => {
+				if (!d.startDate) { return; }
+				if (!d.properties) { return; }
+				if (!d.properties[type]) { return; }
+				//Garmin average stress level is -2 if you are not wearing the device
+				if (parseInt(d.properties[type]) <= 0) { return; }
```
chrisnowak, line 29: Should this be parseFloat?  I guess I dunno what happens if you call parseInt on a float value, presumably since it's JS it'll just parse the int part of it?

src/components/view/GarminView/GarminView.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react'
-import { Layout, Card, StatusBarBackground, ConnectGarmin, GarminDevices, GarminMonthCharts } from "../.."
+import { Layout, Card, StatusBarBackground, ConnectGarmin, GarminDevices, GarminMonthCharts, LoadingIndicator } from "../.."
```
chrisnowak, line 2: I don't see a LoadingIndicator in this view?

chrisnowak (approved):

src/components/view/GarminView/GarminView.tsx
ajmenca: Its down on Line 41.

src/helpers/daily-data-providers/query-garmin-total-daily-values.ts
ajmenca: I brought this part over from the single daily value version of it, so was just keeping consistent here.  But yes, it will just parse the integer portion.  Amsuingly you can do something as strange as `parseInt("4.foo")` and you'll get the value `4` out of it.

PR #167, opened 2023-12-08
Make some adjustments to the NavigationBar.

greinard:
## Overview

This branch updates the `NavigationBar` in a few ways.
- Adds support for a "save" button (and associated click handler).
- Adds some additional button styling and positioning properties.
- Adds a property to set the background color.
- Adjusts the "compressed" variant a tad bit (I moved the title higher up so it sits between the buttons).
image

These changes are in support of some upcoming designs I have been working on.

## Security

- No new security risks.  Just updating the rendering and capabilities of the navigation bar a little bit.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

### DIFF
--- a/src/components/presentational/NavigationBar/NavigationBar.tsx
+++ b/src/components/presentational/NavigationBar/NavigationBar.tsx
@@ -1,49 +1,57 @@
-import React, { useRef, useLayoutEffect, useContext } from 'react'
-import "./NavigationBar.css"
-import MyDataHelps from "@careevolution/mydatahelps-js"
+import React, { CSSProperties, useContext, useLayoutEffect, useRef } from 'react'
+import './NavigationBar.css'
+import MyDataHelps from '@careevolution/mydatahelps-js'
 import '@fortawesome/fontawesome-svg-core/styles.css';
 import { faChevronLeft } from '@fortawesome/free-solid-svg-icons/faChevronLeft'
-import language from "../../../helpers/language"
+import language from '../../../helpers/language'
 import { FontAwesomeSvgIcon } from 'react-fontawesome-svg-icon'
 import { ColorDefinition, resolveColor } from '../../../helpers/colors';
 import { LayoutContext } from '../Layout';
+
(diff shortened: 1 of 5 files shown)


src/components/presentational/NavigationBar/NavigationBar.tsx
```diff
@@ -1,88 +1,123 @@
...
+	backButtonBackgroundColor?: ColorDefinition;
 	showCloseButton?: boolean;
-	children?: React.ReactNode;
 	closeButtonText?: string;
-	backButtonText?: string;
+	closeButtonColor?: ColorDefinition;
+	closeButtonBackgroundColor?: ColorDefinition;
+	closeButtonPosition?: NavigationBarButtonPosition;
+	showSaveButton?: boolean;
+	saveButtonText?: string;
```
chrisnowak, line 26: This feels like a ton of additional properties here which complicate usage of this, when really it is just that you have 2 buttons you want to define right?  

Why not just have a "navigationBarLeft" and "navigationBarRight" props that are ReactNodes...and when specified they basically replace the left (back) / right (close) buttons...they would not have a handler or anything by default, you would specify that in the node.

I think that would serve advanced use cases a lot better, otherwise we will end up having to continually customize this as we come up with new button ideas...like what if we want 2 buttons on the right etc.

greinard: That sounds good to me.  I'll make the adjustment.

I may not "replace" the existing buttons, though, since it may be a little odd to have a "showBackButton" property that doesn't actually do anything when set to `true`.

Instead, as a user of the component, when using the custom left/right nav bar nodes, I would just leave the built-in buttons hidden.

greinard: I have made the suggested changes.  Please take another look when you have a moment.

chrisnowak (approved):

PR #168, opened 2023-12-08
Add a new ValueSelector component.

greinard:
## Overview

This branch adds a new `ValueSelector` component that can be used to present options for single or multiple selection.

There are two variants, buttons (the default) and checkboxes.

You can customize the colors.

There is also a setting to prevent an empty selection (i.e. requiring that at least one value remains selected).

## Security

- No new security risk.  This just adds a new UI component that can be used to provide users with value selection.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation


src/components/presentational/ValueSelector/ValueSelector.tsx
```diff
@@ -0,0 +1,100 @@
+import React, { CSSProperties, useContext, useEffect, useState } from 'react';
+import './ValueSelector.css'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCircle } from '@fortawesome/free-regular-svg-icons';
+import { faCheckCircle } from '@fortawesome/free-solid-svg-icons';
+import { ColorDefinition, resolveColor } from '../../../helpers/colors';
+import { LayoutContext } from '../Layout';
+
+export interface ValueSelectorProps {
+	backgroundColor?: ColorDefinition;
```
chrisnowak, line 10: Quick thing that jumped out at me - I have mostly left component backgrounds up to the container, since Card allows you to define a background color.  Is there a reason this should have a separate background color?

greinard: We can probably drop that.  I had put it there because the default colors (white buttons) requires a gray background to work.  I added the override property because I figured if I was setting a default color in CSS, there should be a way to override it.

greinard: I removed the background color and override property.  It now relies on the containing element for its background color.

src/components/presentational/ValueSelector/ValueSelector.css
```diff
@@ -0,0 +1,77 @@
...
+    --mdhui-value-selector-button-value-selected-background-color: var(--mdhui-color-primary);
+    --mdhui-value-selector-button-value-selected-text-color: #fff;
+    --mdhui-value-selector-checkbox-color: var(--mdhui-text-color-4);
+    --mdhui-value-selector-checkbox-selected-color: var(--mdhui-color-primary);
+    padding: 16px;
+}
+
+.mdhui-value-selector-title {
+    font-weight: bold;
+    font-size: 18px;
```
chrisnowak, line 15: Would always use ems ...and I would stick to sizes consistent with other components even if it is a little off from the design...so that we are consistent.  I think for instance our default text size (e.g. 1em) ends up being 17px not 18px but 12px is commonly used

chrisnowak: (using ems lets iOS dynamic text work across everything though, e.g. if someone bumps up their desired font size)

src/components/presentational/ValueSelector/ValueSelector.css
```diff
@@ -0,0 +1,76 @@
+.mdhui-value-selector {
+    --mdhui-value-selector-title-color: var(--mdhui-text-color-1);
+    --mdhui-value-selector-subtitle-color: var(--mdhui-text-color-3);
+    --mdhui-value-selector-value-background-color: #fff;
+    --mdhui-value-selector-value-text-color: var(--mdhui-text-color-1);
+    --mdhui-value-selector-button-value-selected-background-color: var(--mdhui-color-primary);
+    --mdhui-value-selector-button-value-selected-text-color: #fff;
```
chrisnowak, line 7: This will break it in dark mode, since the other mdhui colors adjust automatically, so this will end up as white on white

src/components/presentational/ValueSelector/ValueSelector.tsx
```diff
@@ -0,0 +1,98 @@
...
+	variant?: 'default' | 'checkboxes';
+	multiSelect?: boolean;
+	preventEmptySelections?: boolean;
+	innerRef?: React.Ref<HTMLDivElement>;
+}
+
+export default function (props: ValueSelectorProps) {
+	const layoutContext = useContext(LayoutContext);
+
+	const [selectedValues, setSelectedValues] = useState<string[]>([]);
```
chrisnowak, line 32: Why have a state variable here?  Seems like this is pretty similar to the SegmentedControl where it would make sense to be stateless?  Especially since selectedValues is a parameter.

Whatever parent component you use is presumably gonna need to save the value anyways, no?

src/components/presentational/ValueSelector/ValueSelector.css
chrisnowak: Overall I would just test and make sure the default colors work nicely in dark mode...understand that all the colors ColorDefinition would let you configure it look nice in dark mode but I think it should look good without any of that

chrisnowak: so just here for the #FFF values I would probably use the CSS var for that as well

chrisnowak (approved):
Looks great!

src/components/presentational/ValueSelector/ValueSelector.css
greinard: I modified `--mdhui-value-selector-value-background-color` to use a variable, but I left `--mdhui-value-selector-button-value-selected-text-color` as `#fff`.  I don't see a variable for white text at the moment.  I could add one, but in this case, I'd actually want it to remain white since the primary color does not adjust for light vs dark mode.

src/components/presentational/ValueSelector/ValueSelector.stories.tsx
```diff
@@ -0,0 +1,226 @@
...
+import { Card, Layout } from '../../presentational';
+import ValueSelector, { ValueSelectorProps } from './ValueSelector'
+
+export default {
+	title: 'Presentational/ValueSelector',
+	component: ValueSelector,
+	parameters: {layout: 'fullscreen'}
+};
+
+let render = (cardColor: ColorDefinition, args: ValueSelectorProps) => <Layout colorScheme="auto" bodyBackgroundColor="#fff">
```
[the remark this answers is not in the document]
greinard: Done.

src/components/presentational/ValueSelector/ValueSelector.css
chrisnowak: Right, yes...same thing for normal mdhui-buttons

src/components/presentational/ValueSelector/ValueSelector.tsx
greinard: I think I added it because it felt weird for the component to not be functional in the story book.  However, you are correct, there should always be a containing component that is managing that state.

I've removed the state from the component and updated the story rendering to maintain that state so that component remains functional in the storybook.

PR #171, opened 2023-12-11
Add support for applying an alpha value to resolved colors.

greinard:
## Overview

This branch updates the `resolveColor` helper function to accept an optional alpha value to be applied to resolved colors.

I have updated the `CalendarDay` streak highlighting stories to demostrate how this can be utilized.

## Security

- No new security risks.  This change just allows for applying optional alpha transparency to resolved colors.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

### DIFF
--- a/src/components/presentational/CalendarDay/CalendarDay.stories.tsx
+++ b/src/components/presentational/CalendarDay/CalendarDay.stories.tsx
@@ -2,6 +2,7 @@ import React, { useContext } from 'react';
 import CalendarDay, { CalendarDayProps, CalendarDayStateConfiguration } from './CalendarDay';
 import { Calendar, Card, DateRangeContext, DateRangeCoordinator, Layout } from '../../presentational';
 import { add, formatISO, isSameDay, parseISO } from 'date-fns';
+import { resolveColor } from "../../../helpers/colors";
 
 export default {
     title: 'Presentational/CalendarDay',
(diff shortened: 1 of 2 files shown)


chrisnowak:
@greinard not sure I get this.  ColorDefinition is just 2 strings that get applied to CSS so you should be able to just supply an RGBA value?   It shouldn't need to be separate?

greinard:
The reason I had for this functionality is to be able to highlight things or apply backgrounds to things based on a specified color without needing to have additional specific color properties for each of those things.

In this example, "Symptoms", "Impacts", and "Triggers" each have a color (purple, orange, black).  The edit button uses the primary color (blue).  Using this updated `resolveColor` function, I am able to apply the text color using the specified colors, but highlight them with partially transparent backgrounds (based on the same color), without needing to directly specify those background colors.  If the color of one of these things changes, such as the primary color, the background of the button automatically adjusts.

image

greinard:
Looking at it a bit more.  I can probably accomplish what I am trying to do with some adjusted CSS rules.  Let me take a stab at that real quick, and if that works, I'll close out this PR.

greinard:
I was able to get this working with some adjusted CSS rules.  I do have a feeling that this functionality may come up as a useful thing to have in the future, but for the time being, I'm going to close this PR.

PR #173, opened 2023-12-15
Severity calendar

reneefromhold:
## Overview

A calendar that supports logging of severity and showing days and streaks where it was mild, moderate, or severe. 
The first use case is for showing swelling severity/streaks

Changes outside of component files
-making the CalendarDay component's onclick event optional 
-adding get-survey-answers to the helper folder to reduce duplication of code that gets all surveyanswers at once.

## Security

- No ppt information is exposed in the code changes
- The component requires configuration as described in the Readme for connecting to ppt information
- 
### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/container/SwellingCalendar/SwellingCalendar.previewdata.tsx
```diff
@@ -0,0 +1,226 @@
+import { SurveyAnswer } from "@careevolution/mydatahelps-js";
+import { addDays, startOfDay } from "date-fns";
+
+var today = new Date();
+
+export var previewSwellingData : SurveyAnswer[] = [
```
greinard, line 6: Minor: Use 'const' instead of `var`.

src/components/container/SwellingCalendar/SwellingCalendar.previewdata.tsx
```diff
@@ -0,0 +1,226 @@
...
+        surveyVersion: 1,
+        surveyName: "Survey 1",
+        surveyDisplayName: "test",
+        date: startOfDay(addDays(today, -8)).toISOString(),
+        stepIdentifier: "1",
+        resultIdentifier: "swellingDate",
+        answers: [startOfDay(addDays(today, -8)).toISOString()],
+        insertedDate: Date.now.toString()
+    }
+]
```
greinard, line 226: Minor:  Missing semi-colon.

src/components/container/SwellingCalendar/SwellingCalendar.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+import SwellingCalendar from "./SwellingCalendar";
+import { SurveyResultDataParameters } from "../../../helpers/daily-data-providers";
+
+export default {
+    title: "Container/SwellingCalendar",
+    component: SwellingCalendar,
+    parameters: {
+        layout: 'fullscreen',
+    }
+} as ComponentMeta<typeof SwellingCalendar>;
```
greinard, line 14: This is using the deprecated version of story setup.  It should be converted to using the newer pattern.   See the `CalendarDay` stories for an example.

https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/components/presentational/CalendarDay/CalendarDay.stories.tsx

src/components/container/SwellingCalendar/SwellingCalendar.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+import Card from "../../presentational/Card"
+import Layout from "../../presentational/Layout"
+import SwellingCalendar from "./SwellingCalendar";
+import { SurveyResultDataParameters } from "../../../helpers/daily-data-providers";
+
+export default {
+    title: "Container/SwellingCalendar",
+    component: SwellingCalendar,
+    parameters: {
+        layout: 'fullscreen',
```
greinard, line 12: Minor: `'` (single quotes) here is inconsistent with the rest of the file.  Also, the extra `,` is not needed at the end.

src/components/container/SwellingCalendar/SwellingCalendar.tsx
```diff
@@ -0,0 +1,177 @@
...
+    const surveyAnswerQuery : SurveyAnswersQuery = {
+        surveyName : props.surveyName,
+        resultIdentifier : [props.dateRecordedResultIdentifier, props.severityResultIdentifier]
+    }
+    const [datePagerMonth, setMonth] = useState<Date>(startOfMonth(new Date()));
+    const [data, setData] = useState<Map<string, SwellingDay> | undefined>();
+
+    async function initialize(){
+        if (!data){
+            if (props.previewState && props.previewState == "NoData")
```
greinard, line 36: `props.previewState == "NoData"` should be enough here.  You don't need the initial `props.previewState &&` part.  Also, while I believe Typescript handles it okay, I still prefer to use `===` when performing exact comparisons to avoid any ambiguity.

src/components/container/SwellingCalendar/SwellingCalendar.tsx
```diff
@@ -0,0 +1,177 @@
...
+            {
+                transformToCalendarData([]);
+            } else {
+                if (props.previewState && props.previewState == "WithData"){
```
greinard, line 40: Minor:  This can be `else if` instead of being separate.  Again here, the `props.previewState && ` part isn't needed.

src/components/container/SwellingCalendar/SwellingCalendar.tsx
```diff
@@ -0,0 +1,177 @@
...
+                background: 'var(--mdhui-color-danger)'
+            },
+            streak: true,
+            streakColor: '#FCCCC7'
+        },
+    };
+
+    const computeStateForDay = (date: Date): string => {
+        let key : string = startOfDay(date).toDateString();
+        return (data && data?.get(key)?.severity) ?? "";
```
greinard, line 79: Minor:  You don't need the `data &&` part here, which will also let you get rid of the extra set or parenthesis.

src/components/container/SwellingCalendar/SwellingCalendar.tsx
```diff
@@ -0,0 +1,177 @@
...
+        swellingEntries.forEach( swellingEntry => {
+            let key : string = startOfDay(parseISO(swellingEntry.dateRecorded)).toDateString();
+            let exists = calendarData.get(key);
+            if (exists){
+                if (parseISO(swellingEntry.dateRecorded) > parseISO(exists.dateRecorded)){
+                    exists.severity = swellingEntry.severity;
+                    calendarData.set(key, exists);
+                }
+            }
+            else {
```
greinard, line 95: Please run and auto-format on this file.  I've noticed a couple of things here and there that might be corrected by that.

src/components/container/SwellingCalendar/SwellingCalendar.tsx
```diff
@@ -0,0 +1,177 @@
...
+        var swellingEntries : Map<Guid, SwellingDay> = createResultsMap(results);
+
+        var calendarData : Map<string, SwellingDay> = new Map<string, SwellingDay>();
+        swellingEntries.forEach( swellingEntry => {
+            let key : string = startOfDay(parseISO(swellingEntry.dateRecorded)).toDateString();
+            let exists = calendarData.get(key);
+            if (exists){
+                if (parseISO(swellingEntry.dateRecorded) > parseISO(exists.dateRecorded)){
+                    exists.severity = swellingEntry.severity;
+                    calendarData.set(key, exists);
```
greinard, line 92: I'm not sure, but I don't think you need to set it again here.  The updated `exists` object will still be in the map with that key.

src/components/container/SwellingCalendar/SwellingCalendar.tsx
```diff
@@ -0,0 +1,177 @@
...
+        />;
+    };
+
+    function pageMonth(newStart: Date) {
+        setMonth(startOfMonth(newStart));
+    } 
+
+
+    useEffect(() => {
+        initialize();
```
greinard, line 158: There is a helper hook you can use to get the same functionality as this `useEffect` setup.

```
useInitializeView(initialize);
```

However, I'm also not sure this is going to do anything meaningful for you.  I may be misunderstanding the logic, but I noticed that the `initialize` function only loads data if `data` has not been set yet.  It looks like `data` will be set in all cases after the first execution of `initialize()`.  Or am I missing something there?

src/components/container/SwellingCalendar/SwellingCalendar.tsx
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }, []);
+
+
+    if (!data){
+        return <LoadingIndicator/>
+    }
+    else {
+        return <>
+            <DateRangeNavigator intervalType={"Month"} intervalStart={datePagerMonth} onIntervalChange={pageMonth}></DateRangeNavigator>
```
greinard, line 172: The `DateRangeNavigator` should not be part of this component.  The `DateRangeNavigator` will already be provided by a wrapping `DateRangeCoordinator`.  You'll just want to reference a `DateRangeContext` in this component to get the correct dates to use.

```
const dateRangeContext = useContext(DateRangeContext);
```


Here is the Asthma Control Calendar (not in PR yet) to peek at for an example:  https://github.com/CareEvolution/MyDataHelpsUI/blob/AsthmaUI/src/components/asthma/components/AsthmaControlCalendar/AsthmaControlCalendar.tsx

You can also check out the Symptom Shark calendar for another example: https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/components/symptom-shark/presentational/Calendar/Calendar.tsx

src/components/container/SwellingCalendar/SwellingCalendar.tsx
```diff
@@ -0,0 +1,177 @@
...
+            <Calendar className="mdhui-simple-calendar" year={datePagerMonth.getFullYear()} month={datePagerMonth.getMonth()} dayRenderer={renderDay}/>
+            <Button defaultMargin onClick={() => MyDataHelps.startSurvey(props.surveyName)}>{language("log-swelling")} <FontAwesomeSvgIcon icon={faPlus} /></Button>
```
greinard, line 174: I don't think this button should be in here either.  You should be able to add a button below the calendar component when building views or other components that contain this calendar component.

src/helpers/get-survey-answers.ts
```diff
@@ -0,0 +1,33 @@
...
+export default async function(props : SurveyAnswersQuery) : Promise<SurveyAnswer[]> {
+
+    const results = await getSurveyAnswers();
+    var sortedResults = (results).sort((a, b) => {
+            if (parseISO(a.date) > parseISO(b.date)) { return -1; }
+            if (parseISO(a.date) < parseISO(b.date)) { return 1; }
+            return 0;
+    });
+
+    return sortedResults;
```
greinard, line 13: Just a nitpick, but it was a bit confusing to me at first to have the return up here in the middle of the function.  I think it would read easier if the functions below were defined above (or as separate functions outside this function), with the return being the last thing that happens.  Just a personal preference, though.

src/components/container/index.ts
```diff
@@ -27,4 +27,5 @@ export { default as ProviderSearch } from "./ProviderSearch"
 export { default as RelativeActivityToday } from "./RelativeActivityToday"
 export { default as RestingHeartRateCalendar } from "./RestingHeartRateCalendar"
 export { default as SurveyTaskList } from "./SurveyTaskList"
+export {default as SwellingCalendar} from "./SwellingCalendar"
```
greinard, line 30: Minor: Formatting is off here.

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     // Default
-    "target": "es5",
+    "target": "es2015",
```
greinard, line 4: Not disagreeing with this change, but just wondering what you ran into that needed it?

greinard (changes requested):
It may be worth moving this component into a subfolder of components (like `symptom-shark` and soon `asthma`) to distinguish this from general reusable components since it has a fairly specific domain and purpose.

tsconfig.json
greinard: Nevermind.  I see your explanation in the PR description.  😃

reneefromhold: @greinard I reverted this. I was using Set initially but moved away from it when moving to use Map.

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,149 @@
+import { useContext, useState } from "react";
+import { Guid, SurveyAnswer, SurveyAnswersQuery } from "@careevolution/mydatahelps-js";
+import React from "react";
+import surveyResults from "../../../helpers/get-survey-answers";
+import Calendar from "../../presentational/Calendar/Calendar";
+import { parseISO, startOfDay, startOfMonth } from "date-fns";
+import { CalendarDay, CalendarDayStateConfiguration, DateRangeContext, LoadingIndicator } from "../../presentational";
+import { previewSeverityData } from "./SeverityCalendar.previewdata";
+import { useInitializeView } from "../../../helpers/Initialization";
+export type SeverityCalendarPreviewState = "WithData" | "NoData" | "Loading" | "Live";
```
chrisnowak, line 10: You shouldn't have a "Live" preview state - not having a preview state defined is "Live" (it is the default use of the component)

src/helpers/get-survey-answers.ts
```diff
@@ -0,0 +1,33 @@
...
+        if (pageID) {
+            queryParameters.pageID = pageID;
+        }
+
+        return MyDataHelps.querySurveyAnswers(queryParameters);
+    }
+
+    const results = await getSurveyAnswers();
+    var sortedResults = (results).sort((a, b) => {
+            if (parseISO(a.date) > parseISO(b.date)) { return -1; }
```
chrisnowak, line 27: Sorting should be unnecessary; answers are already sorted by date via the API...it also seems out of place in this function

src/helpers/get-survey-answers.ts
```diff
@@ -0,0 +1,33 @@
+import MyDataHelps, { Guid, SurveyAnswer, SurveyAnswersPage, SurveyAnswersQuery } from "@careevolution/mydatahelps-js";
+import {parseISO} from "date-fns";
+
+export default async function(props : SurveyAnswersQuery) : Promise<SurveyAnswer[]> {
```
chrisnowak, line 4: I would export this function as "queryAllSurveyAnswers"...that is what distinguishes it from the SDK querySurveyAnswers method...it is getting *all* the answers

src/components/presentational/CalendarDay/CalendarDay.tsx
```diff
@@ -69,7 +69,7 @@ export default function (props: CalendarDayProps) {
         }
     }
 
-    return <div ref={props.innerRef} className={dayClasses.join(' ')} style={dayStyle} onClick={() => props.onClick(date)}>
+    return <div ref={props.innerRef} className={dayClasses.join(' ')} style={dayStyle} onClick={() => props.onClick ? props.onClick(date) : void(date)}>
```
chrisnowak, line 72: I missed this in the original review of CalendarDay @greinard 

For accessibility, we should not attach onClick to div but instead have a child ```<UnstyledButton>``` element that the onClick is attached to, so that tab completion etc. works appropriately

When there is no "onClick" we would not include the button.

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,149 @@
...
+            }
+        });
+
+        setData(calendarData);
+    }
+
+    function createResultsMap(results: SurveyAnswer[]) {
+        var groupedByResult = new Map<Guid, SeverityDay>();
+
+        results.forEach((r) => {
```
chrisnowak, line 97: I might do this a bit differently:
- First do a pass to group the survey answers by surveyResultID.  So you have an {[key:Guid]: SurveyAnswer[]}
- Then loop through that collection and:
  - If there's no severity answer all, skip this group
  - If there's no date answer at all, use the date from the severity answer (to allow for a situation where you aren't "dating" the logs" but are always logging for today)

In particular want to make sure this is relatively safe and if you misconfigure it, you don't end up with JS errors or anything...and it seems reasonable to fall back to using the date of the survey answer for severity

src/components/presentational/CalendarDay/CalendarDay.tsx
greinard: 👍 https://github.com/CareEvolution/MyDataHelpsUI/pull/176

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,149 @@
...
+            else {
+                calendarData.set(key, severityEntry);
+            }
+        });
+
+        setData(calendarData);
+    }
+
+    function createResultsMap(results: SurveyAnswer[]) {
+        var groupedByResult = new Map<Guid, SeverityDay>();
```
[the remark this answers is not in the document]
reneefromhold: Not technical, just preference. Are there any concerns with it?

src/components/container/SeverityCalendar/SeverityCalendar.tsx
reneefromhold: This makes sense.

src/components/presentational/CalendarDay/CalendarDay.tsx
reneefromhold: I am going to revert this change in favor of the PR @greinard attached to last comment.

src/helpers/query-all-survey-answers.ts
```diff
@@ -0,0 +1,25 @@
...
+        let dataPage = await getSurveyDataPage();
+        let allData = dataPage.surveyAnswers;
+        while (dataPage.nextPageID) {
+            dataPage = await getSurveyDataPage(dataPage.nextPageID);
+            allData = allData.concat(dataPage.surveyAnswers);
+        }
+        return allData;
+    }
+
+    async function getSurveyDataPage(pageID?: Guid): Promise<SurveyAnswersPage> {
```
greinard, line 15: Nitpick - Maybe rename this to `getSurveyAnswersPage`?

src/helpers/query-all-survey-answers.ts
```diff
@@ -0,0 +1,25 @@
...
+        var queryParameters: SurveyAnswersQuery = props;
```
greinard, line 16: Do we need the extra variable here?  I think you can just set the `pageID` on `props` in the conditional below.

If the intent was to not modify the `props` object, I think this may need to be `{...props}` instead.

src/helpers/query-all-survey-answers.ts
```diff
@@ -0,0 +1,25 @@
+import MyDataHelps, { Guid, SurveyAnswer, SurveyAnswersPage, SurveyAnswersQuery } from "@careevolution/mydatahelps-js";
+
+export default async function (props: SurveyAnswersQuery): Promise<SurveyAnswer[]> {
+
+    async function getSurveyAnswers(): Promise<SurveyAnswer[]> {
+        let dataPage = await getSurveyDataPage();
+        let allData = dataPage.surveyAnswers;
```
greinard, line 7: Nitpick - Maybe rename this to `allAnswers`?

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,158 @@
...
+            computeStateForDay={computeStateForDay}
+        />;
+    };
+
+    useInitializeView(initialize, [], [props.previewState]);
+
+    if (!data) {
+        return <LoadingIndicator />
+    }
+    else {
```
greinard, line 155: Minor:  My settings may be the ones that are off, but I think we usually like the `else {` to be on the same line as the `}`.

Not a big deal really, but I think we should try to all sync up on the settings so we can avoid diffs when we just autoformat files.

src/components/container/SeverityCalendar/SeverityCalendar.previewdata.tsx
```diff
@@ -0,0 +1,332 @@
...
+    {
+        id: "102345834898",
+        surveyResultID: "12",
+        surveyID: "2",
+        surveyVersion: 1,
+        surveyName: "Survey 1",
+        surveyDisplayName: "test",
+        date: endOfDay(addDays(today, -10)).toISOString(),
+        stepIdentifier: "1",
+        resultIdentifier: "swellingDate",
```
greinard, line 302: I see a couple references to "swelling" in the sample data.  May want to rename those to "severity" as well.

src/components/container/SeverityCalendar/SeverityCalendar.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+            <SeverityCalendar {...args} />
+        </DateRangeCoordinator>
+    </Card>
+</Layout>;
+
+export const Default = {
+    args: {
+        surveyName: 'Survey 1',
+        severityResultIdentifier: 'severityValue',
+        dateRecordedResultIdentifier: 'swellingDate',
```
greinard, line 31: I assume this will need to be renamed as well if the sample data gets updated.

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,158 @@
...
+                        severity = "severe";
+                        break;
+                }
+
+                if (severity) {
+                    let dateOfSeverity = parseISO(grouping.dateObservedAnswer ? grouping.dateObservedAnswer.answers[0] : grouping.severityAnswer.date);
+                    let key = format(startOfDay(dateOfSeverity), 'MM/dd/yyyy');
+                    let exists = resultByDateMap.get(key);
+                    if (exists) {
+                        if ((dateOfSeverity > exists.dateObserved) ||
```
greinard, line 74: Minor:  No need for the extra parenthesis on the first condition.

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,158 @@
...
+                    case "severity3":
+                        severity = "severe";
+                        break;
+                }
+
+                if (severity) {
+                    let dateOfSeverity = parseISO(grouping.dateObservedAnswer ? grouping.dateObservedAnswer.answers[0] : grouping.severityAnswer.date);
+                    let key = format(startOfDay(dateOfSeverity), 'MM/dd/yyyy');
+                    let exists = resultByDateMap.get(key);
+                    if (exists) {
```
greinard, line 73: I think you might be able to simplify this conditional update as: 

```
if (!exists || dateOfSeverity > exists.dateObserved || (dateOfSeverity === exists.dateObserved && grouping.severityAnswer.date > exists.dateEntered)){
	resultByDateMap.set(key, { dateObserved: dateOfSeverity, severity: severity, dateEntered: grouping.severityAnswer.date });
}
```

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+        var resultByDateMap = new Map<string, SeverityLogEntry>();
+        var groupedByResult = groupAnswersByResult(results);
+
+        groupedByResult.forEach((grouping) => {
+            var severity = "";
+            if (grouping.severityAnswer) {
+                switch (grouping.severityAnswer.answers[0]) {
+                    case "mild":
+                    case "severity1":
```
greinard, line 56: Do we need to support these alternate severity values in this generic component?  They seem like they represent some project/survey specific implementation details.

I wonder if this calendar should support just `mild`, `moderate`, and `severe` by default, but then allow for an optional value mapping function prop in case a particular project wants to use values like `severity1`, `severity2`, and `severity3`.

greinard (commented):
Looks pretty good @reneefromhold.  I just had the one functional question to discuss about the supported severity values.  The rest of the comments were just minor nitpicks that you can address if you see fit.  😃

src/components/container/SeverityCalendar/SeverityCalendar.tsx
greinard: Thinking about this a tad more.  Is the `dateOfSeverity > exists.dateObserved` condition even necessary?  If the `key` and `dateObserved` are always set to the same value, which is the `dateOfSeverity`, then I'm not sure how the `dateOfSeverity` could ever be greater than the `exists.dateObserved`.

Am I missing something in the logic here?

greinard: I *think* `dateOfSeverity` and `exists.dateObserved` will always be equal.  So maybe this can simplify even further to: 

```
if (!exists || grouping.severityAnswer.date > exists.dateEntered){
	resultByDateMap.set(key, { dateObserved: dateOfSeverity, severity: severity, dateEntered: grouping.severityAnswer.date });
}
```

reneefromhold: The severity entry can come in with and without a dedicated date observed result.
You are right that with the key being the dateObserved that testing if the date is greater than has no value. Good catch. 
If the user enters multiple entries for the same Observed date then the most recent needs to be used.

greinard: I think the `dateOfSeverity === exists.dateObserved` condition will always be true as well, given how the data is set.  The `key` and `dateObserved` are always set to the same thing, so any time the `key` from a given `dateOfSeverity` is used to look up an existing record, `dateOfSeverity` will match the existing record's `dateObserved` value.

src/components/container/SeverityCalendar/SeverityCalendar.tsx
reneefromhold: I created a severityValueMapper on the props

src/components/container/SeverityCalendar/SeverityCalendar.tsx
reneefromhold: We slacked and agreed to use } else {

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,149 @@
...
+    dateObserved: Date,
+    severity: string,
+    dateEntered: string
+}
+
+export default function (props: SeverityCalendarProps) {
+    const surveyAnswerQuery: SurveyAnswersQuery = {
+        surveyName: props.surveyName,
+        resultIdentifier: props.dateRecordedResultIdentifier ? [props.dateRecordedResultIdentifier, props.severityResultIdentifier] : [props.severityResultIdentifier]
+    }
```
greinard, line 31: Minor: missing semi-colon.

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+        var resultByDateMap = new Map<string, SeverityLogEntry>();
+        var groupedByResult = groupAnswersByResult(results);
+        var ceSeverityValueMapper = function(severityAnswer : string){
+            let severity = "";
+            if (['mild', 'moderate', 'severe'].includes(severityAnswer.toLowerCase())){
+                severity = severityAnswer.toLowerCase();
+            }
+            return severity;
+        }
```
greinard, line 57: Minor: missing semi-colon.

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,149 @@
...
+                transformToCalendarData(results);
+            });
+        }
+    }
+
+    function transformToCalendarData(results: SurveyAnswer[]) {
+
+        var resultByDateMap = new Map<string, SeverityLogEntry>();
+        var groupedByResult = groupAnswersByResult(results);
+        var ceSeverityValueMapper = function(severityAnswer : string){
```
greinard, line 51: This can probably be a `const` instead of a `var`.

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,149 @@
...
+import { CalendarDay, CalendarDayStateConfiguration, DateRangeContext, LoadingIndicator } from "../../presentational";
+import { previewSeverityData } from "./SeverityCalendar.previewdata";
+import { useInitializeView } from "../../../helpers/Initialization";
+export type SeverityCalendarPreviewState = "Default" | "NoData";
+
+export interface SeverityCalendarProps {
+    surveyName: string,
+    dateRecordedResultIdentifier?: string,
+    severityResultIdentifier: string,
+    severityValueMapper? : () => void,
```
greinard, line 16: I think this should have a type of: `(value: string) => string`.

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,149 @@
...
+                            exists.dateObserved = dateOfSeverity;
+                            exists.severity = severity;
+                            exists.dateEntered = grouping.severityAnswer.date;
+                        }
+                    } else {
+                        resultByDateMap.set(key, { dateObserved: dateOfSeverity, severity: severity, dateEntered: grouping.severityAnswer.date });
+                    }
+                }
+            }
+        })
```
greinard, line 78: Minor: missing semi-colon.

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,149 @@
...
+            day={day}
+            stateConfiguration={stateConfiguration}
+            computeStateForDay={computeStateForDay}
+        />;
+    };
+
+    useInitializeView(initialize, [], [props.previewState]);
+
+    if (!data) {
+        return <LoadingIndicator />
```
greinard, line 145: Minor: missing semi-colon.

src/components/container/SeverityCalendar/SeverityCalendar.tsx
```diff
@@ -0,0 +1,149 @@
...
+    } else {
+        return <Calendar className="mdhui-simple-calendar" year={intervalStart.getFullYear()} month={intervalStart.getMonth()} dayRenderer={renderDay} />
```
greinard, line 147: Minor: missing semi-colon.

src/components/container/SeverityCalendar/SeverityCalendar.stories.tsx
```diff
@@ -0,0 +1,78 @@
...
+        severityResultIdentifier: '',
+        dateRecordedResultIdentifier: '',
+        previewState: 'NoData'
+    },
+    render: render
+};
+
+export const Live = {
+    args: {
+        surveyName: 'Swelling and Urine Protein Measurements',
```
greinard, line 73: Minor:  A couple more references to swelling.

greinard (commented):
Looks pretty good.  I think the only remaining functional issue would be the mapping function type declaration.

greinard (approved):

PR #177, opened 2023-12-21
Add an "expanded" display variant to the SingleSurveyTask component.

greinard:
## Overview

This branch adds a new `expanded` display variant to the `SingleSurveyTask` component.  The existing component has undergone a bit of refactoring to make this happen, but no changes to the default display variant or its functionality are intended.

I'll get the new Spanish strings translated prior to merging.

I also spent a little extra time on the storybook entry.  Rather than trying to enumerate a bunch of different state permutations, I have customized the story editor controls such that it makes it easy to test out different scenarios.

image

## Security

- No security implications.  Just adding a new rendering variant for survey tasks.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation


src/components/presentational/SingleSurveyTask/SingleSurveyTask.stories.tsx
```diff
@@ -1,91 +1,83 @@
...
-export const Complete = {
-	args: {
-		...commonProps,
-		task: {
-			...commonTask,
-			status: 'complete',
-			endDate: add(new Date(now), {days: -2}).toISOString(),
+		status: {
+			control: 'radio',
+			options: ['incomplete', 'complete', 'closed']
```
reneefromhold, line 60: Closed is new and shows a blank page. Is that by design? I'm not clear on the value it adds

src/components/presentational/SingleSurveyTask/SingleSurveyTask.tsx
```diff
@@ -17,75 +16,98 @@ import { ButtonVariant } from '../Button/Button';
...
 
-	let today = new Date();
-	let tomorrow = add(new Date(), { days: 1 });
-	let dueDateString = "";
-	let dueDateIntent = "";
-	let locale = MyDataHelps.getCurrentLanguage().toLowerCase().startsWith("es") ? es : enUS;
-	if (props.task.dueDate) {
+	const getDueDate = () => {
+		let today = startOfToday();
+		let tomorrow = add(new Date(today), {days: 1});
```
reneefromhold, line 41: nit: startOfTomorrow

src/helpers/functions.ts
```diff
@@ -0,0 +1,2 @@
+export function noop() {
+}
```
reneefromhold, line 2: Should something be here?

src/helpers/strings-es.ts
```diff
@@ -84,6 +84,8 @@
     "app-download-app-store-link-alt": "Descargar en la tienda de aplicaciones de Apple",
     "start": "Inicio",
     "resume": "Continuar",
+    "start-survey": "Start Survey",
```
reneefromhold, line 87: I believe if there is no translation in this file, it will default to English. Worth confirming bc this entry does not appear to be translated.

src/components/presentational/SingleSurveyTask/SingleSurveyTask.tsx
```diff
@@ -17,75 +16,98 @@ import { ButtonVariant } from '../Button/Button';
 import checkMark from '../../../assets/greenCheck.svg';
 import Action from '../Action';
 import LoadingIndicator from '../LoadingIndicator';
+import { getLocale } from '../../../helpers/date-helpers';
+import { noop } from '../../../helpers/functions';
```
reneefromhold, line 20: I'm not clear on what noop is doing? Is there a different way to accomplish what it is doing that does not involve a blank file + import?

src/components/presentational/SingleSurveyTask/SingleSurveyTask.stories.tsx
greinard: `closed` is one of the three possible survey task statuses you can get back from the server.  It happens when a coordinator has closed a survey manually.  We don't typically query for `closed` tasks in our out-of-the-box survey list components, but this component could be given one.  And yes, the expected (existing) behavior is that it renders `closed` survey tasks as blank.

https://github.com/CareEvolution/MyDataHelps.js/blob/master/src/types.ts#L1

src/components/presentational/SingleSurveyTask/SingleSurveyTask.tsx
greinard: I'm not sure it needs to be that specific.  The "start of" is just being used to remove the fluctuating time component from both dates since that part isn't really relevant for the "same day/next day" logic being applied here.

src/helpers/functions.ts
greinard: Nope.  This is intended to be a no-op function that can be used anywhere that a void function is needed without having to use `() => {}`.

src/helpers/strings-es.ts
greinard: I'll substitute in the Spanish translations prior to merging.

src/components/presentational/SingleSurveyTask/SingleSurveyTask.stories.tsx
```diff
@@ -1,91 +1,83 @@
...
+		},
+		dueDate: {
+			name: 'due date',
+			control: 'date',
```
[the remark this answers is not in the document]
greinard: I don't think so.

src/components/presentational/SingleSurveyTask/SingleSurveyTask.tsx
greinard: You can see the usage down here (`Button` requires a click listener, but I don't want it to do anything when clicked):  https://github.com/CareEvolution/MyDataHelpsUI/pull/177/files/feebf11de5146de56601851c90c9b37568d203ee#diff-dcf4df28a6983ef66aeb389be69aee68e9b916c3823798af4cb6b827d10900a4R74

I just thought it looked cleaner in the code to have a specific "no-op" function available rather than using `() => {}` all over the place.  The use of `noop` looks more intentional to me, and less like a click listener that someone forgot to finish implementing.

reneefromhold (approved):

PR #189, opened 2024-01-19
Add Inbox UI components and views.

greinard:
## Overview

This branch adds new "Inbox" components and views.

Inbox functionality goes beyond an ordinary task list by additionally supporting messages and external resource links (i.e. external articles, blog posts, pdfs, etc).

The messages can also reference external resources (i.e. to link to further information related to a message).

External resource links can optionally include an image header to increase their visibility.

Below are some screenshots of the new reusable UI components and views.

&nbsp;&nbsp;

Dark mode and Spanish are also supported. 

## Security

- No security risks.  These new components use existing API's and components to provide the new "Inbox" functionality.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation


src/components/container/InboxItemList/InboxItemList.css
```diff
@@ -0,0 +1,8 @@
+.mdhui-inbox-item-list-title {
+    color: var(--mdhui-text-color-2);
+}
+
+.mdhui-inbox-item-list-empty-text {
+    color: var(--mdhui-text-color-3);
+    padding: 16px 0;
```
reneefromhold, line 7: var --mdhui-padding-md

src/components/presentational/InboxCompletedListItem/InboxCompletedListItem.css
```diff
@@ -0,0 +1,22 @@
+.mdhui-inbox-completed-list-item {
+    background: var(--mdhui-background-color-0);
+    border-radius: 12px;
+    margin: 12px 0;
+}
+
+.mdhui-inbox-completed-list-item .mdhui-action, .mdhui-inbox-completed-list-item .mdhui-card {
+    border-radius: 12px;
+    padding: 16px;
```
reneefromhold, line 9: --mdhui-padding-md

src/components/container/InboxItemList/InboxItemList.tsx
```diff
@@ -0,0 +1,130 @@
...
+    let inboxContext = useContext(InboxItemListCoordinatorContext);
+    if (inboxContext && props.syncOnChanges) {
+        inboxContext.registerChangeListener(loadData);
+    }
+
+    useInitializeView(loadData, [], [props.previewState]);
+
+    if (loading && !items) {
+        return <div className="mdhui-inbox-item-list" ref={props.innerRef}>
+            {props.title && props.showTitleWhenEmpty &&
```
reneefromhold, line 77: Nothing wrong with the code here but more the presentation. 
When I pull up the Inbox View the font is very similar to the survey description. It took a moment to orient my eyes with what I was looking at vs being clearly drawn to Message/Survey/Resource/

src/components/presentational/InboxResourceDisplay/InboxResourceDisplay.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import './InboxResourceDisplay.css'
+import { InboxMessageResource, InboxResource } from '@careevolution/mydatahelps-js'
+import '@fortawesome/fontawesome-svg-core/styles.css';
+import UnstyledButton from '../UnstyledButton/UnstyledButton';
+import Button from '../Button/Button';
+import { noop } from '../../../helpers/functions';
+import language from '../../../helpers/language';
+
+export interface InboxResourceDisplayProps {
```
reneefromhold, line 10: The presentation components don't look like they need to be directly tied to Inbox objects. Is it possible to decouple the inbox aspect? 
Reflecting on studies that have asked for resource lists, this would satisfy showing a resource list that doesn't have to be inbox driven.

src/components/container/InboxItemList/InboxItemList.css
greinard: I wasn't aware of those variables, but I'm also a little hesitant to use them.

I think sharing colors (and maybe some other things) via variables makes sense, but I'm concerned that sharing dimensional settings could mean we end up with unforeseen or unintended consequences if/when we ever adjust those.  For example, if we decide at some point that `--mdhui-padding-md` should be `12px` or `20px`, we'll need to go through and find all the components that are using that variable and make sure things still rendered correctly.

I'd be interested to hear others thoughts on it though.  I'm not totally opposed to it, but it just doesn't feel as obvious to me to be something we should use vars for, like colors.

src/components/container/InboxItemList/InboxItemList.tsx
greinard: Would you be able to provide a screenshot?

I think the Inbox UI is likely to undergo some more polish and improvement in the very near future as we get some mileage on it in the wild, so I'd rather not spend too much time trying to make it "perfect" without some user feedback, but if there is an obvious way we can improve things out of the box, I'm open to that.

src/components/presentational/InboxResourceDisplay/InboxResourceDisplay.tsx
greinard: I think that would be okay for this component.  I'll refactor it to make it reusable outside of the Inbox.

On a similar note to the dimensional vars discussion, I do worry about the possibility of us oversharing non-building-block components in some cases.  The more places we use a given component, the more places have to be okay with that component changing in response to a need elsewhere.  For something complex, it makes sense to move the complex logic into reusable code, but I think it can sometimes make just as much sense to have "your own copy" of less complex code, depending on the context, so you can modify it as needed for a particular use case without affecting other implementations across and outside the library.  I think we just have to try to strike the right balance between reusability/interdependence and context specific customization.

greinard: Again, interested to hear others thoughts on the above. ^

src/components/container/InboxItemList/InboxItemList.tsx
reneefromhold: @greinard 
image

src/components/container/InboxItemList/InboxItemList.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+    let inboxContext = useContext(InboxItemListCoordinatorContext);
+    if (inboxContext && props.syncOnChanges) {
+        inboxContext.registerChangeListener(loadData);
+    }
+
+    useInitializeView(loadData, [], [props.previewState]);
+
+    if (loading && !items) {
+        return <div className="mdhui-inbox-item-list" ref={props.innerRef}>
```
reneefromhold, line 76: There is no class definition for mdhui-inbox-item-list

src/components/container/InboxItemList/InboxItemList.css
reneefromhold: Having the variable is good for consistency, especially for across container/presentation components. For views, I could see more wiggle room on not using.
If we do not want to use them at all, is there value in keeping the variable?

src/components/container/InboxItemList/InboxItemList.tsx
reneefromhold: Some notes from slack convo. Noting here for posterity
- Font size of message title vs task title would look nicer if consistent
- The view has not limit to how many messages/surveys/etc. Would reconsider showing message titles in bold as it could be a list of bold text
- Section titles are like floating signs. My eye goes toward the bold message and survey titles, then the section Message/Surveys/Resources come into focus. Does not look as clean as it could

reneefromhold (approved):
lgtm overall. Left comments on css variable for padding, and a request to decouple 'inbox' from the presentational components, or at very lease resource and message list items and the padding variable. There are existing studies asking for resource lists where the resource list is the same for all ppt and not something that has to be assigned/inboxed.

src/components/presentational/InboxResourceDisplay/InboxResourceDisplay.tsx
greinard: I reworked the `InboxResourceDisplay` component so that it is now a `Resource` component that is not tied to the Inbox.

I would like to keep the "list item" components tied to the Inbox for now though.  They have some very specific inbox list functionality built into them.

src/components/container/InboxItemList/InboxItemList.tsx
greinard: Yeah, it didn't need any customization at this time.  I always like to have classes assigned to tags, though, so they can be styled later or by users of this library when necessary.

src/components/container/InboxItemList/InboxItemList.tsx
greinard: As discussed, this design is likely to have some iterations in the near future to adjust for user feedback.  Since this design has been used previously without issue, I hesitate to modify it too much until we get some real world feedback.

src/components/container/InboxItemList/InboxItemList.css
greinard: I guess I'm still not quite sold on it, at least not at this general of a level.

If the variable were something like `--mdhui-padding-for-empty-list-text`, then it would make more sense to me as something we are using to have a consistent look and feel for empty text across our list components.  However, just having the padding set to a generic "medium" padding seems less meaningful and more difficult to update.  For example, what if we wanted to adjust the "medium" padding used around titles or buttons or something, but we didn't want to change the padding around empty text in lists?  If they are using the same variable, it would all have to change, because there is no contextual link between the variable and where it is being used.

greinard:
Thanks for the detailed review @reneefromhold.  I will pass along your design feedback.  I'm sure there will be a couple more PRs related to the Inbox in the future as we iterate on the design.

src/components/container/InboxItemList/InboxItemList.css
chrisnowak: I don't feel super strongly about it FWIW, so long as we are consistent about spacing and basing it on the 4/8/12/16px rhythm 

I originally added this because I was thinking about SolveME / Metrodora - we slightly tweaked this and ended up using 24px padding instead of 16px in a lot of places to provide a little more breathing room and a simpler, less busy display.

It would perhaps be nice to achieve that kind of thing without CSS overrides aplenty like I ended up doing for Solve; however:
- Don't think it's a massive difference or super high priority necessarily
- I think it probably only applies for specific components/situations; lots of edge cases where it could break things otherwise

So probably the right solution here is to have a CSS variable that is used only in certain places to be able to "stretch" the UI a bit.  That being said again not really that concerned about it

PR #196, opened 2024-02-01
Add RotatingText component.

greinard:
## Overview

This branch adds a new `RotatingText` component that can be used to display text in a round robin fashion on a daily, weekly, or monthly interval.

Each "text" entry can have a title and a global title prefix can be set on the component.

text

## Security

### Testing
- [x] MyDataHelps iOS app
- [x] MyDataHelps Android app

### Documentation


src/components/presentational/RotatingText/RotatingText.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import './RotatingText.css'
+import { differenceInDays, differenceInMonths, differenceInWeeks, min } from 'date-fns';
+
+export type RotatingTextInterval = 'day' | 'week' | 'month';
+
+export interface RotatingTextEntry {
+    title?: string;
```
chrisnowak, line 8: I could see more use cases for this beyond just title/text...but any child components (like a rotating series of news articles for instance).

And in particular if we do it that way we could just embed a text component with markdown in the view builder. 

I think though you could also achieve this using conditional blocks and some of the new expression functions we are adding to the view builder, since it'll include differenceInMonths / differenceInWeeks / differenceInDays - so maybe that is the answer if you want arbitrary rotating stuff and this is more of a simple use case...

greinard: Are you thinking that it would just have a `children` property, but rather than render them all, it would render just one child at a time based on the interval computation?

So in ViewBuilder, you could add as many children of whatever type you wanted to it, and it would display them one at a time in round robin fashion?

I think that makes sense and should allow me to implement the functionality I need with a couple less hard coded components.

chrisnowak: I didn't get so far as to think about how to achieve it, but that actually sounds like it would work great!  It would only render 1 of it's children each time based on the day

greinard: It wasn't too much trouble.  I renamed the component to `RotatingComponentCoordinator`.  It round robin displays each of its children based on the interval computation.

greinard:
Closing this PR.  The change (with revisions) is included in: https://github.com/CareEvolution/MyDataHelpsUI/pull/203

PR #203, opened 2024-02-06
Asthma UI Components

greinard:
## Overview

This branch contains a bunch of new asthma related components.  I apologize in advance for the size of this PR.  I had hoped to break things down into a bunch of smaller sequential PRs, but due to time constraints, that was simply going to take too long.

Please feel free to review incrementally as you have time.  If at all possible, it would be great if I could merge this in by the end of the week.

The best way to explore the new components is probably via the storybook, but if you would like to test out the components in practical use, I can get you set up in a dry run environment where they are being actively used/tested.  Just let me know.

I'm still working on adding a bunch of Spanish translations to that language file, but the rest of the files should be good for review.

## Security

### Testing
- [x] MyDataHelps iOS app
- [x] MyDataHelps Android app

### Documentation


src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
```diff
@@ -0,0 +1,102 @@
+import React, { useState } from 'react';
+import './AsthmaActionPlanManager.css';
+import { Button, LoadingIndicator, UnstyledButton } from '../../../presentational';
+import MyDataHelps from '@careevolution/mydatahelps-js';
+import { useInitializeView } from '../../../../helpers/Initialization';
+import sampleActionPlan from '../../assets/sample_aap.png'
```
chrisnowak, line 6: I would not bundle this with the app but instead host this somewhere else (e.g. just upload it with MDHD).  Otherwise it will add 50kb to the view builder package for instance...which seems unnecessary

src/components/asthma/views/AsthmaActionPlanView/AsthmaActionPlanView.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { Layout, NavigationBar } from '../../../presentational';
+import { AsthmaActionPlanManager } from '../../components';
+import { AsthmaActionPlan } from '../../model';
+
+export interface AsthmaActionPlanViewProps {
+    colorScheme?: 'light' | 'dark' | 'auto';
+    previewState?: 'loading' | 'loaded without action plan' | 'loaded with action plan'
+    learnMoreUrl: string;
+    onViewActionPlan: (actionPlan: AsthmaActionPlan) => void;
```
chrisnowak, line 10: What does this typically invoke?  Seems like it is always just viewing the file no...

src/components/asthma/model/types.ts
```diff
@@ -0,0 +1,155 @@
...
+}
+
+export interface AsthmaLibraryArticle {
+    title: string;
+    subTitle?: string;
+    url: string;
+    imageUrl?: string;
+}
+
+export interface AsthmaActionPlan {
```
chrisnowak, line 152: Duplicate interface definition

src/components/container/DailyDataChart/DailyDataChart.css
```diff
@@ -2,13 +2,18 @@
     position: relative;
 }
 
-.mdhui-daily-data-chart .title {
-    padding: 16px;
-    padding-bottom: 0;
-    position: relative;
+.mdhui-daily-data-chart-title {
+    padding: 0 16px;
+    font-size: 1.2em;
```
chrisnowak, line 7: Why the font increase here?  (This will impact all existing charts for people using this component etc.)

src/helpers/strings-es.ts
```diff
@@ -158,7 +159,16 @@
     "inbox-view-resources-title": "Recursos",
     "inbox-view-recently-completed-title": "Reciente",
     "inbox-view-recently-completed-empty-text": "No tienes elementos recientemente completados en tu bandeja de entrada.",
-    "inbox-view-history-button-text": "Ver historial de bandeja de entrada"
+    "inbox-view-history-button-text": "Ver historial de bandeja de entrada",
```
chrisnowak, line 162: I would populate all the Asthma strings just using ChatGPT

src/helpers/strings-en.ts
```diff
@@ -162,7 +163,151 @@
...
+    "asthma-sleep-view-chart-title": "Sleep Disturbances",
+    "asthma-sleep-view-alert-message": "Your sleep disturbances are above your normal level.",
+    "asthma-log-entry-editor-view-symptom-level-title": "Symptom level",
+    "asthma-log-entry-editor-view-symptom-level-subtitle": "Select one",
+    "asthma-log-entry-editor-view-symptoms-title": "Symptoms",
+    "asthma-log-entry-editor-view-symptoms-subtitle": "Select all that apply",
+    "asthma-log-entry-editor-view-impacts-title": "Impacts",
+    "asthma-log-entry-editor-view-impacts-subtitle": "Select all that apply",
+    "asthma-log-entry-editor-view-triggers-title": "Triggers",
+    "asthma-log-entry-editor-view-triggers-subtitle": "Select all that apply"
```
chrisnowak, line 310: Would not create separate strings for these...does not really make sense IMO...we do not separate each "done" button in every context for instance...we have a single string for "done".  A string for "select-all-that-apply" is much more sensible than duplicating the strings

src/components/container/DailyDataChart/DailyDataChart.css
```diff
@@ -2,13 +2,18 @@
     position: relative;
 }
 
-.mdhui-daily-data-chart .title {
-    padding: 16px;
-    padding-bottom: 0;
-    position: relative;
+.mdhui-daily-data-chart-title {
+    padding: 0 16px;
```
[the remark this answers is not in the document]
greinard: Good catch, sorry, please ignore the changes you see here to the DailyDataChart component.  I thought I had reverted all of those changes at some point.  I will do so momentarily.  I ended up going a different route for the Asthma chart, so I don't need any modifications to the `DailyDataChart` component.

src/components/presentational/SingleDataPoint/SingleDataPoint.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export default function (props: SingleDataPointProps) {
+    const getContents = () => {
+        return <>
+            <div className="mdhui-single-data-point-value-units">
+                {!props.value &&
+                    <div className="mdhui-single-data-point-value">...</div>
+                }
+                {props.value &&
```
chrisnowak, line 23: When the value is 0 (which could be common if this were used for something like steps) it does weird stuff

Screenshot 2024-02-07 at 4 27 12 PM

(I would probably default to treating 0 as a "no data" state...that is what most of the device data queriers do and I think the chart does...because often for Fitbit for instance data we get 0s when someone is not wearing a device)

chrisnowak: However that logic is maybe best handled upstream from here actually.  Is there a reason we just don't have value as a string?  That might make the most sense for this presentational component since I think this same component could be useful for non-numeric things

src/components/container/DailyDataChart/DailyDataChart.css
greinard: The `DailyDataChart` changes have been reverted.

src/helpers/strings-es.ts
greinard: Yep, that is what I have been doing.  I just have a bunch more to go.  😄

src/components/presentational/SingleDataPoint/SingleDataPoint.tsx
```diff
@@ -0,0 +1,49 @@
...
+                }
+            </div>
+            <div className="mdhui-single-data-point-label">{props.label}</div>
+            {props.statusText &&
+                <div className="mdhui-single-data-point-status" style={props.statusColor ? {color: props.statusColor} as CSSProperties : undefined}>{props.statusText}</div>
+            }
+        </>;
+    };
+
+    return <div ref={props.innerRef}>
```
chrisnowak, line 37: This pattern makes me wonder if we should move this into UnstyledButton...e.g. UnstyledButton becomes a div if there is no click handler.  Probably gets complicated with the refs though and we'd need a div wrapper or something which might mess stuff up I guess

src/components/asthma/model/types.ts
greinard: Fixed.

src/helpers/strings-en.ts
greinard: I waffled on that a bit and then went with this in case we wanted to change the subtitle for one of the value selectors without changing it for the others.  That is probably unlikely, though, and I don't feel strongly about it.

I have updated the log edit view selector subtitles to share common strings.

src/components/asthma/helpers/asthma-data.ts
```diff
@@ -0,0 +1,338 @@
...
+            } else {
+                return answers;
+            }
+        };
+
+        return getPage().then(function (surveyAnswers) {
+            return surveyAnswers;
+        });
+    },
+    loadAsthmaActionPlan: async function (): Promise<AsthmaActionPlan | undefined> {
```
chrisnowak, line 332: Hmm...wondering if this is necessary...this ends up being a report on the back-end right?  Can we just get the report ID and reference it?  Or does it need to be base64 encoded for some reason

src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
```diff
@@ -0,0 +1,102 @@
...
+        if (props.previewState) return;
+        MyDataHelps.startSurvey(props.editActionPlanSurveyName);
+    };
+
+    const onUploadActionPlan = (): void => {
+        if (props.previewState) return;
+        MyDataHelps.startSurvey(props.editActionPlanSurveyName);
+    };
+
+    return <div className="mdhui-asthma-action-plan-manager">
```
chrisnowak, line 79: Not critical for this PR but this component feels like it is close to being something that could be generic...e.g. upload your vaccine card or something of that nature...can wait until we have more robust APIs for image capture though

src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
```diff
@@ -0,0 +1,102 @@
...
+                setLoading(false);
+            }
+        });
+    };
+
+    useInitializeView(initialize, [], [props.previewState]);
+
+    const onLearnMore = (): void => {
+        if (props.previewState) return;
+        MyDataHelps.openApplication(props.learnMoreUrl, {modal: true});
```
reneefromhold, line 61: In terms of flexibility, does this have to be a URL? Idk what to expect in terms of this being repurposed, but it were to be reused, this could be difficult to workaround

src/components/asthma/helpers/daily-data-providers/air-quality-home.ts
```diff
@@ -0,0 +1,6 @@
+﻿import { DailyDataQueryResult } from '../../../../helpers/query-daily-data';
+import { collateMaxValueDataPoints, queryAsthmaDeviceData } from './shared';
+
+export default function (startDate: Date, endDate: Date): Promise<DailyDataQueryResult> {
+    return queryAsthmaDeviceData('AirNowApi', 'HomeAirQuality', startDate, endDate, collateMaxValueDataPoints);
```
chrisnowak, line 5: I would move these AirNowApi daily data providers out to the normal daily data providers...this should be able to be reused no

src/components/asthma/components/AsthmaAirQualities/AsthmaAirQualities.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { useInitializeView } from '../../../../helpers/Initialization';
+import { asthmaDataService, getAsthmaDataStatusColor, getAsthmaDataStatusText } from '../../helpers';
+import { LoadingIndicator, SingleDataPoint, UnstyledButton } from '../../../presentational';
+import { AsthmaAirQualitiesPreviewState, previewData } from './AsthmaAirQualities.previewData';
+import MyDataHelps from '@careevolution/mydatahelps-js';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faGear } from '@fortawesome/free-solid-svg-icons';
+import language from '../../../../helpers/language';
+
+export interface AsthmaAirQualitiesProps {
```
reneefromhold, line 13: Is this something that has to be coupled with Asthma? or can it be more general?

src/components/asthma/helpers/daily-data-providers/air-quality-home.ts
reneefromhold, line 5: Can this be decoupled from Asthma? For other studies where AirQuality may be relevant

src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.css
```diff
@@ -0,0 +1,64 @@
+.mdhui-asthma-action-plan-manager {
+    padding: 16px;
+}
+
+.mdhui-asthma-action-plan-manager-title {
```
chrisnowak, line 5: Can we just use a <Title> component here

src/components/asthma/components/AsthmaAirQualities/AsthmaAirQualities.css
```diff
@@ -0,0 +1,53 @@
+.mdhui-asthma-air-qualities {
+    padding: 16px;
+}
+
+.mdhui-asthma-air-qualities-header {
+    display: grid;
+    grid-template-columns: auto max-content;
+}
+
+.mdhui-asthma-air-qualities-title {
```
chrisnowak, line 10: Same here this feels like we can just use a <Title> component

src/components/asthma/components/AsthmaAirQualities/AsthmaAirQualities.tsx
```diff
@@ -0,0 +1,110 @@
...
+        if (airQuality.status === 'not-configured' && props.editZipCodesSurveyName) {
+            MyDataHelps.startSurvey(props.editZipCodesSurveyName);
+        } else {
+            MyDataHelps.openApplication(props.airQualityUrl, {modal: true});
+        }
+    };
+
+    return <div className="mdhui-asthma-air-qualities" ref={props.innerRef}>
+        <div className="mdhui-asthma-air-qualities-header">
+            <div className="mdhui-asthma-air-qualities-title">{language('asthma-air-qualities-title')}</div>
```
chrisnowak, line 77: I would probably just drop the title since you could just use a ```<Title>``` component

chrisnowak: Wait nm, you need a gear on the right

chrisnowak: Position of the gear seems slightly off, feels like it should be 16px from the right edge of the card?


Screenshot 2024-02-07 at 5 25 14 PM

src/components/presentational/SingleDataPoint/SingleDataPoint.tsx
greinard: No reason.  Sounds reasonable to me.  I've change the type of the `value` prop to be a `string`.

src/components/presentational/SingleDataPoint/SingleDataPoint.tsx
greinard: Yeah.  I'm also not sure when you would use an `UnstyledButton` without a click listener (it is not currently optional).

It might be nice to have an "MDH-UI div" component that renders as an `UnstyledButton` when a click listener is provided, else a `div`.  That could be used anywhere you want to use a regular `div`, and if you happen to add a click listener to one, you don't have to think about reworking it for the focus traversal stuff.

src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
greinard: Done.

src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
greinard: I think we typically implement drilldowns as separate modal views which launch via URL (as opposed to a within-view pop up).

Did you have something specific in mind for how a learn more drilldown might be implemented differently?

src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
greinard: Yeah, I can see that.  Happy to do a refactoring to be more generic once we have more of the image capture/query API stuff ironed out.

src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
reneefromhold: Surveys can be easier for sites to create info pages vs urls.

src/components/asthma/helpers/daily-data-providers/air-quality-home.ts
greinard: Yep, good idea.  I've moved the AQI queries to the the normal daily data providers folder and reworked the Asthma ones a little bit to clear up the confusion we discussed yesterday.

src/components/presentational/RotatingComponentCoordinator/RotatingComponentCoordinator.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { Children } from 'react';
+import { differenceInDays, differenceInMonths, differenceInWeeks } from 'date-fns';
+import isAfter from "date-fns/isAfter";
+
+export type RotatingComponentInterval = 'day' | 'week' | 'month';
+
+export interface RotatingComponentCoordinatorProps {
```
reneefromhold, line 7: The name of this component is misleading. Was expecting to see text rotating, or something to that effect.
This is showing specific text based on a the diff in date intervals.

src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
greinard: That is a good point.  As we discussed, I think I am going to leave this as a URL for now, since that is the only way it will be used for the foreseeable future, but I do think we should think about this on a more global level.

Perhaps we can think about bringing something like the `ActionDefinition` concept into MDH-UI from the view builder for cases where we want the user of the component to be able to define the action to be taken vs having the component define what the action taken will be.

src/components/asthma/helpers/daily-data-providers/air-quality-home.ts
greinard: Yep, I have moved this into the common daily data providers area.

src/components/presentational/Section/Section.tsx
```diff
@@ -7,6 +7,7 @@ export interface SectionProps {
 	children?: React.ReactNode;
 	className?: string;
 	noTopMargin?: boolean;
+	noBottomMargin?: boolean;
```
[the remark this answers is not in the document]
greinard: I was just mirroring the existing `noTopMargin` prop and wanted to retain backward compatibility (including a bottom border by default).

Do you have a suggestion for what to name it instead?

src/components/presentational/RotatingComponentCoordinator/RotatingComponentCoordinator.tsx
greinard: Happy to rename it, but I couldn't think of something better.  Feel free to suggest a new name.  😃

src/components/asthma/components/AsthmaBarChart/AsthmaBarChart.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { checkDailyDataAvailability, DailyDataProvider, DailyDataQueryResult, queryDailyData } from '../../../../helpers/query-daily-data';
+import { LoadingIndicator } from '../../../presentational';
+import { add, format, startOfToday } from 'date-fns';
+import getDayKey from '../../../../helpers/get-day-key';
+import { AxisDomain } from 'recharts/types/util/types';
+import { Bar, BarChart, CartesianGrid, Cell, LabelList, ResponsiveContainer, XAxis, YAxis } from 'recharts';
+import { useInitializeView } from '../../../../helpers/Initialization';
+import './AsthmaBarChart.css';
+
+export interface AsthmaBarChartProps {
```
reneefromhold, line 11: Does this need to be coupled with Asthma? Just a bar chart
Also, need stories?

src/components/asthma/components/AsthmaLibraryArticles/AsthmaLibraryArticles.tsx
```diff
@@ -0,0 +1,42 @@
...
+export default function (props: AsthmaLibraryArticlesProps) {
+    const onClick = (article: AsthmaLibraryArticle): void => {
+        if (props.previewState) return;
+        MyDataHelps.openEmbeddedUrl(article.url);
+    };
+
+    let articles = props.previewState ? previewData[props.previewState].articles : props.articles;
+
+    return <div className="mdhui-asthma-library-articles" ref={props.innerRef}>
+        <>
```
reneefromhold, line 25: Decoupling opportunity.
This is resource list

src/components/asthma/components/AsthmaAirQualities/AsthmaAirQualities.tsx
greinard: We can probably create a general component for displaying AQI in a similar fashion, but I think this one is pretty asthma specific in that the data being loaded and the statuses being computed are dependent upon whether we are loading them for a particular date or to reflect the data being used to compute the Asthma control state.

The statuses and status colors are also Asthma specific at this time.

src/components/presentational/RotatingComponentCoordinator/RotatingComponentCoordinator.tsx
ajmenca: I'll note that I had the same reaction to this name.  I'm also struggling to come up with an alternative name.  I'm fine with something verbose, but everything I come up with is *extremely* verbose.  This is a `TimeBasedComponentCarousel`?  It's not really a carousel because it has no paging controls though.  `TimeBasedComponentSwitcher`?  `TimeIntervalComponentSwitcher`?

src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.css
greinard: Yep, fixed.  I've also gone through and updated the other components that have a title div to use `<Title>` instead.

src/components/asthma/components/AsthmaAirQualities/AsthmaAirQualities.css
greinard: Yep, done.

src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
chrisnowak: I think you would not need to bring in "ActionDefinition", you just need to expose function prop for handling the click and then in the view builder you can have that processed by the ActionDefinition.

(ActionDefinition feels inherently like a view builder thing; programmatically I can just handle the event)

src/components/asthma/components/AsthmaBarChart/AsthmaBarChart.tsx
chrisnowak: Functionally it is not much different than the DailyDataChart set to start 7 days ago w/ no date range nav but it is a bit more refined and obviously supports highlighting/formatting...(and there is no reason for us to only support a single style of bar chart...)

src/components/presentational/Section/Section.tsx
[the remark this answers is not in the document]
chrisnowak: Yeah this was just because I very commonly wanted to exclude the top margin; I think view builder has noTopMargin set universally since all other cards/sections have a margin by default.

We should probably just follow the pattern from TextBlock though and pass through ```style```:
https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/components/presentational/TextBlock/TextBlock.tsx#L11

And have this in the view builder like we do for text:
Screenshot 2024-02-08 at 2 16 33 PM

I would keep noTopMargin universally on section in the viewbuilder though so by default it's still 0 top margin, but can be overridden with^

src/components/presentational/RotatingComponentCoordinator/RotatingComponentCoordinator.tsx
chrisnowak: IDK...personally of the options I prefer RotatingComponentCoordinator...yeah it is a bit verbose but when I try to succinctly describe this thing to myself I end up saying that it's a rotation of content...

src/components/asthma/views/AsthmaHeartAndLungsView/AsthmaHeartAndLungsView.tsx
```diff
@@ -0,0 +1,137 @@
...
+import language from '../../../../helpers/language';
+import { randomDataProvider } from '../../../../helpers/daily-data-providers';
+
+const DaytimeRestingHeartRateDailyDataType = 'Asthma.DaytimeRestingHeartRate';
+const NighttimeRestingHeartRateDailyDataType = 'Asthma.NighttimeRestingHeartRate';
+const RespiratoryRateDailyDataType = 'Asthma.RespiratoryRate';
+const DaytimeBloodOxygenLevelDailyDataType = 'Asthma.DaytimeBloodOxygenLevel';
+const NighttimeBloodOxygenLevelDailyDataType = 'Asthma.NighttimeBloodOxygenLevel';
+
+registerDailyDataProvider(DaytimeRestingHeartRateDailyDataType, daytimeRestingHeartRateDataProvider, simpleAvailabilityCheck('Project', ['DaytimeRestingHeartRate']));
```
chrisnowak, line 17: There's no cost to "registering" a daily data provider and this feels awkward to have to declare in order for a view to work?  I would probably chuck this code into asthma-data or something that's imported at a higher level?

src/components/asthma/components/AsthmaBiometrics/AsthmaBiometrics.tsx
```diff
@@ -0,0 +1,135 @@
...
+    }, [], [props.previewState]);
+
+    const onClick = (type: AsthmaBiometricType): void => {
+        if (props.previewState) return;
+        if (type === 'daytime-resting-heart-rate' || type === 'nighttime-resting-heart-rate' || type === 'respiratory-rate' || type === 'daytime-blood-oxygen-level' || type === 'nighttime-blood-oxygen-level') {
+            MyDataHelps.openApplication(props.heartAndLungsUrl, {modal: true});
+        } else if (type === 'steps') {
+            MyDataHelps.openApplication(props.activityUrl, {modal: true});
+        } else if (type === 'sleep-disturbances') {
+            MyDataHelps.openApplication(props.sleepUrl, {modal: true});
```
chrisnowak, line 69: Yeah here's  another place you could have these be event handlers that go to an ActionDefinition in the view builder.  Don't feel super strongly about it though

src/components/asthma/components/AsthmaLibraryArticles/AsthmaLibraryArticles.tsx
chrisnowak: Agreed^ nothing Asthma specific here that I can see?

src/components/asthma/components/AsthmaAirQualities/AsthmaAirQualities.tsx
greinard: The gear was not an issue when switching to use `<Title>`.

I also adjusted the position of the gear.  It was originally right-justified, but I moved it in because the focus the focus traversal on the larger hidden touch target looked odd (see images below).  That is probably of little concern though, so I moved it back to the right.

**Previous with focus:**
aqi1
**Current with focus:**
aqi2
**Current without focus:**
aqi3

src/components/asthma/components/AsthmaAirQualities/AsthmaAirQualities.tsx
```diff
@@ -0,0 +1,110 @@
...
+            {props.editZipCodesSurveyName &&
+                <UnstyledButton className="mdhui-asthma-air-qualities-settings" onClick={() => onSetup()}>
+                    <FontAwesomeIcon icon={faGear}/>
+                </UnstyledButton>
+            }
+        </div>
+        {loading && <LoadingIndicator/>}
+        {!loading &&
+            <div className="mdhui-asthma-air-qualities-data">
```
ajmenca, line 86: Minor suggestion: It might be useful to actually create a `GridLayout` or something to that effect as a generic Presentational Component. It recently came up in other conversations that someone wanted to put two buttons on the same line; same problem you're having here with laying out `SingleDataPoints`.  The css class you added here and for Biometrics presentation are identical, which also sort of helps imply that this this could be a unified presentation option.

src/components/asthma/helpers/asthma-data.ts
```diff
@@ -0,0 +1,343 @@
...
+        let mostRecentDataPoint = dataPoints.sort(sortDataPointsDesc)[0];
+        if (mostRecentDataPoint.properties?.hasOwnProperty('viewed')) return undefined;
+
+        mostRecentDataPoint.properties = mostRecentDataPoint.properties ?? {};
+        mostRecentDataPoint.properties['viewed'] = formatISO(new Date());
+        MyDataHelps.persistDeviceData([mostRecentDataPoint]).then();
+
+        return mostRecentDataPoint.value;
+    },
+    loadSurveyAnswers: async function (surveyNames: string[]): Promise<SurveyAnswer[]> {
```
chrisnowak, line 315: This feels like it could be a generic "queryAllSurveyAnswers" like we have "queryAllDeviceData" since it's not asthma specific at all, just gets everything

src/components/asthma/helpers/asthma-data.ts
chrisnowak: Yeah, I think given we have a pending refactor of the image stuff it probably doesn't matter too much if this works for now.

chrisnowak (approved):
Definitely some stuff here that can be generalized which is awesome...don't know that we need to make that a bottleneck to deploying this though unless there is obvious low hanging fruit.  Still massive value here and it makes it easier to generalize since these components all live here now!

I only skimmed some pretty large chunks of this (e.g. stories) and a lot of the deep logic since you are the expert at that...but overall this is awesome

src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
greinard: Fair enough.  That makes sense.

I think the bigger question will be a more philosophical one.  Should library components decide which actions to use by directly calling things like open application or start survey themselves, or should they all have event handler properties which are implemented externally to the component, thus allowing those events to be handled in a view-builder-configurable way using action definitions?

In other words, should we ever directly call open application or start survey from within in a component?  If yes, what is the criteria for deciding when the component should define which action to perform vs allowing it to be determined by an action definition config in the view builder?

src/components/asthma/components/AsthmaAlertTakeoverNotice/AsthmaAlertTakeoverNotice.tsx
```diff
@@ -0,0 +1,62 @@
...
+    useInitializeView(() => {
+        setLoading(true);
+
+        if (props.previewState) {
+            setLoading(props.previewState === 'loading');
+            return;
+        }
+
+        asthmaDataService.loadLogEntries(add(new Date(), {days: -2})).then(entries => {
+            const todayLogEntry = entries.find(e => e.identifier === dateToAsthmaLogEntryIdentifier(new Date()));
```
reneefromhold, line 29: dateToAsthmaLogEntryIdentifier can be called just once vs and compared on each iteration vs called + compared

src/components/asthma/components/AsthmaAlertTakeoverTrigger/AsthmaAlertTakeoverTrigger.tsx
```diff
@@ -0,0 +1,47 @@
...
+        }
+    };
+
+    useInitializeView(() => {
+        if (props.previewState) return;
+
+        asthmaDataService.loadAndClearAlertTakeover().then(alertTakeover => {
+            if (alertTakeover) {
+                asthmaDataService.loadLogEntries(add(new Date(), {days: -2})).then(entries => {
+                    if (!entries.find(e => e.identifier === dateToAsthmaLogEntryIdentifier(new Date()))) {
```
reneefromhold, line 38: Same here for calling dateToAsthmaLogEntryIdentifier(new Date()) on each item

src/components/asthma/components/AsthmaControlCalendar/AsthmaControlCalendar.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+        let today = startOfToday();
+        let logEntryLookup: Record<string, AsthmaLogEntry | undefined> = {};
+
+        let earliestLogEntry = [...logEntries].sort((e1, e2) => e1.identifier.localeCompare(e2.identifier))[0];
+        let earliestDate = parseISO(earliestLogEntry.identifier);
+        for (let i = 0; i <= differenceInDays(today, earliestDate); i++) {
+            let targetDate = add(earliestDate, {days: i});
+            if (!isBefore(targetDate, intervalStart) && isBefore(targetDate, add(intervalStart, {months: 1}))) {
+                logEntryLookup[dateToAsthmaLogEntryIdentifier(targetDate)] = logEntries.find(e => e.identifier === dateToAsthmaLogEntryIdentifier(targetDate));
```
reneefromhold, line 112: Remove calling dateToAsthmaLogEntryIdentifier(targetDate from each iteration in find

src/components/asthma/helpers/asthma-data.ts
```diff
@@ -0,0 +1,343 @@
...
+            type: ['Asthma-LogEntry']
+        };
+        if (fromDate) {
+            params.observedAfter = fromDate.toISOString();
+        }
+        if (toDate) {
+            params.observedBefore = toDate.toISOString();
+        }
+        let result = await MyDataHelps.queryDeviceData(params);
+        return result.deviceDataPoints.map(dp => JSON.parse(dp.value) as AsthmaLogEntry);
```
reneefromhold, line 247: There are a lot of .finds against this data. An object/map keyed by the date used in finds would be more direct

src/components/presentational/RotatingComponentCoordinator/RotatingComponentCoordinator.tsx
reneefromhold: TextDisplayByDateDifference ... ?

src/components/asthma/views/AsthmaLibraryCategoryView/AsthmaLibraryCategoryView.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Layout, NavigationBar, ResourceImageAlignment, Title } from '../../../presentational';
```
reneefromhold, line 2: Can be decoupled from Asthma

src/components/asthma/components/AsthmaBarChart/AsthmaBarChart.tsx
greinard: I've moved it out to the core components as `RecentDailyDataBarChart`.

src/components/presentational/RotatingComponentCoordinator/RotatingComponentCoordinator.tsx
greinard: It has a rotation of content that can be things other than just text, so probably don't want "Text" in the name.

src/components/presentational/Section/Section.tsx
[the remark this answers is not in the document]
greinard: Done.

src/components/asthma/components/AsthmaAirQualities/AsthmaAirQualities.tsx
greinard: When I first started working on this, I actually had a prototype for a `ColumnLayout` component that allowed for a configurable number of columns.  I punted on that due to time constraints, but I agree that a reusable multi-column layout component would be nice to have for use here and other places.  I'll probably work on something like that in a follow up PR.

src/components/asthma/components/AsthmaLibraryArticles/AsthmaLibraryArticles.tsx
greinard: I have move this out to the common component area and renamed it `ResourceList`.

src/components/asthma/views/AsthmaHeartAndLungsView/AsthmaHeartAndLungsView.tsx
greinard: Done.  Moved to `ashtma-data`.

src/components/asthma/helpers/asthma-data.ts
greinard: Turns out that @reneefromhold already added one for that.  I've updated this code to call that existing function.

src/components/asthma/components/AsthmaAlertTakeoverNotice/AsthmaAlertTakeoverNotice.tsx
greinard: Done.

src/components/asthma/components/AsthmaAlertTakeoverTrigger/AsthmaAlertTakeoverTrigger.tsx
greinard: Done.

src/components/asthma/components/AsthmaControlCalendar/AsthmaControlCalendar.tsx
greinard: Done.

src/components/asthma/helpers/asthma-data.ts
greinard: I'm not sure it would be more efficient to always have a map.

There are quite a few places where `.find` is used against this returned list, but not usually in the course of the single load/render cycle.  In other words, when this data is loaded for any particular component, usually there is only a single search of the list, which can short circuit on the first element that matches.  Converting the array to a map would always require looping through the entire list of entries, while building up a larger in-memory data structure.  The single find per load/render only has to loop the entire array if the entry it is searching for happens to be the last one, and no additional memory is used while doing so.

There is one spot in the calendar where more than one of the entries is to be rendered, and it would require multiple iterations on the list to find each one each time.  In that situation, a map of identifier-to-entry is being built and used.

src/components/asthma/views/AsthmaLibraryCategoryView/AsthmaLibraryCategoryView.tsx
greinard: Good idea.  Moved to the common components and renamed to `ResourceListView`.

src/components/asthma/components/AsthmaBiometrics/AsthmaBiometrics.tsx
greinard: As we discussed, I'm going to punt on the inversion of control (event handler/action definition) stuff for now until I can think through the best way to apply it across the components with some level of consistency.

At present, the components are less flexible in that they don't allow for that level of customization, but they are at least consistent in being that opinionated about how each action should be handled.  I don't think that all of the current places where we'd make that change will be able to be converted, so it would have us in a state where some events are handled internal and some externally (even within the same component), which feels less warm and fuzzy to me.  😃

greinard:
Thanks all for the in depth and timely review.

src/components/asthma/helpers/asthma-data.ts
greinard: Can you please link me to the spot you are referring to?

PR #217, opened 2024-03-05
A few component adjustments (mostly Asthma related).

greinard:
## Overview

This branch contains three bug fixes.

1) The Asthma Action Plan Manager now waits a half second before loading the action plan to give any survey submissions the necessary time to update the task run UUID custom field appropriately.  On Android, we noticed that, occasionally, the custom field was not yet updated when the component attempted to reload the AAP.

2) The HR and SpO2 charts on the Asthma Heart and Lung view have been updated to round all values to the nearest whole number rather than displaying decimal values.

3) The "no data" state for the recent daily data bar chart has been updated to handle larger font sizes.

## Security

### Testing
- [x] MyDataHelps iOS app
- [x] MyDataHelps Android app

### Documentation

n/a

### DIFF
--- a/src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
+++ b/src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
@@ -1,8 +1,7 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import './AsthmaActionPlanManager.css';
 import { Button, LoadingIndicator, Title, UnstyledButton } from '../../../presentational';
 import MyDataHelps, { DeviceInfo } from '@careevolution/mydatahelps-js';
-import { useInitializeView } from '../../../../helpers/Initialization';
 import { asthmaDataService } from '../../helpers';
 import { AsthmaActionPlan } from '../../model';
 import language from '../../../../helpers/language';
@@ -37,7 +36,7 @@ export default function (props: AsthmaActionPlanManagerProps) {
         });
     };
 
-    const initialize = (): void => {
+    useEffect(() => {
         setLoading(true);
 
         if (props.previewState === 'loading') {
@@ -58,10 +57,13 @@ export default function (props: AsthmaActionPlanManagerProps) {
             setDeviceInfo(deviceInfo);
         });
 
-        setTimeout(loadActionPlan, 500);
-    };
+        loadActionPlan();
 
-    useInitializeView(initialize, [], [props.previewState]);
+        MyDataHelps.on('surveyDidFinish', loadActionPlan);
+        return () => {
+            MyDataHelps.off('surveyDidFinish', loadActionPlan);
+        }
+    }, [props.previewState]);
 
     const onLearnMore = (): void => {
         if (props.previewState) return;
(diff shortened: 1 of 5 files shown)


src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
```diff
@@ -58,7 +58,7 @@ export default function (props: AsthmaActionPlanManagerProps) {
             setDeviceInfo(deviceInfo);
         });
 
-        loadActionPlan();
+        setTimeout(loadActionPlan, 500);
```
ajmenca, line 61: I feel slightly obligated to explore this a bit, since I think we all know that `setTimeout` as a solution to a problem always feels a little awkward since it doesn't really *guarantee* anything.  In the PR Description you mention a custom field that's not yet updated - how is that update occuring?  Being a little less familiar with the overall Asthma flow and only looking at the widgets in the storybook it's not very clear.

greinard: Yep, for sure.  I am not real fond of it either, but I'm not sure how else to solve it.

Here is what appears to be happening on Android:
1) User visits a view that contains the `AsthmaActionPlanManager`.
2) `AsthmaActionPlanManager` queries for the participant and finds an empty `AAPTaskRunUUID` custom field.  It displays the "Tap to add photo" button.
3) User taps to add a photo and the image capture survey gets launched.
4) User completes the image capture survey, which has a web view step that populates the `AAPTaskRunUUID` custom field with the task run UUID for that survey submission.
5) The `AsthmaActionPlanManager` comes back to the foreground and re-queries for the participant to check for a new value in the `AAPTaskRunUUID` custom field.
6) About 50% of the time during my testing, the `AAPTaskRunUUID` custom field was not yet populated when the `AsthmaActionPlanManager` returned to the foreground and the reload occurred.

I assumed the survey submission loading indicator would remain in the foreground until all custom fields were updated as part of the survey submission, but some part of it must be asynchronous, or the survey submission loading indicator screen on Android must be closing before it completes.

Adding this slight delay to the loading seems to allow enough time to pass for the custom field to always be populated.

greinard: I'm going to tinker around with using the `surveyDidFinish` event instead of the `applicationDidBecomeVisible` event to see if that behaves more like I want it to.  If so, I'll revert this change.  😃

greinard: @ajmenca - Thanks for the nudge on this.  It has the desired behavior when I use the `surveyDidFinish` event instead.  I was able to remove the `setTimeout(...)`.

ajmenca (approved):

src/components/asthma/components/AsthmaActionPlanManager/AsthmaActionPlanManager.tsx
chrisnowak: @greinard Can you run this scenario by @hallc and @dyerjo ... applicationDidBecomeVisible *should* work reliably for this

mmertsock: For the record, here's how it works in the iOS app. First, the app uploads the survey task result. After completing that API request, it triggers a few events simultaneously and asynchronously:

- Uploading any auxiliary files (e.g. Active Task data files)
- Refresh the app UI
- Post surveyDidFinish and applicationDidBecomeVisible JS SDK events

The web content receives those SDK events ±10 milliseconds after the survey task upload API request completes. surveyDidFinish is received first, and immediately afterward (less than a millisecond) applicationDidBecomeVisible.

Issue #224, opened 2024-03-21
Bug: FitbitConnected Widget disappears when Fitbit is in "Needs Reconnection" State


Suvarna-Thejas-CE:
I was using view builder to build a simple UI that allows you to connect Fitbit. 

This Fitbit Connect widget has a flag for "Hide when connected".

This widget remains hidden when the Fitbit Connection Status is "Needs Reconnection", so there is no way to reconnect the Fitbit account from the dashboard. 

i.e. .. this never shows up is "Hide when connected is checked off"
<img width="1126" alt="image" src="https://github.com/CareEvolution/MyDataHelpsUI/assets/60715340/70be93f6-7a79-4e84-88b6-a3c1286c2fe9">

brandoncharnesky:
Fixed via https://github.com/CareEvolution/MyDataHelpsUI/pull/226 @Suvarna-Thejas-CE

PR #226, opened 2024-03-21
For ConnectEhr, ConnectFitbit, and ConnectGarmin widgets, update hideWhenConnected flag to display connections in error

brandoncharnesky:
## Overview

Addresses: https://github.com/CareEvolution/MyDataHelpsUI/issues/224

Adds stories to the ConnectFitbit, ConnectGarmin, and ConnectEhr to cover cases when the hideWhenConnected flag is set.  Updates the default behavior of "hideWhenConnected" to not hide the widget in the case the connection is in an unauthorized or error state.  This will allow a user to view the widget and be able to reconnect or fix a connection error. 

This change can be see in the `ConnectEhr->Hide Connected Enabled Needs Attention` story, the `Fitbit->Hide Connected Unauthorized` and `Garmin->Hide Connected Unauthorized` stories.  Prior to the code change these would be hidden in the UI.

## Security

Changes to how existing connections are displayed in the user interface only.  No changes to underlying connection code.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/container/ConnectFitbit/ConnectFitbit.stories.tsx
+++ b/src/components/container/ConnectFitbit/ConnectFitbit.stories.tsx
@@ -39,3 +39,15 @@ NotEnabledHide.args = { previewState: "notEnabled", title: "Fitbit", disabledBeh
 
 export const NotEnabledDisplayError = Template.bind({});
 NotEnabledDisplayError.args = { previewState: "notEnabled", title: "Fitbit", disabledBehavior: "displayError" };
+
+export const HideConnectedFetchComplete = Template.bind({});
+HideConnectedFetchComplete.args = { previewState: "fetchComplete", title: "Fitbit", hideWhenConnected: true };
+
+export const HideConnectedNotConnected = Template.bind({});
+HideConnectedNotConnected.args = { previewState: "notConnected", title: "Fitbit", hideWhenConnected: true };
+
+export const HideConnectedUnauthorized = Template.bind({});
+HideConnectedUnauthorized.args = { previewState: "unauthorized", title: "Fitbit", hideWhenConnected: true };
+
+export const HideConnectedFetchingData = Template.bind({});
+HideConnectedFetchingData.args = { previewState: "fetchingData", title: "Fitbit", hideWhenConnected: true };
--- a/src/components/container/ConnectGarmin/ConnectGarmin.stories.tsx
+++ b/src/components/container/ConnectGarmin/ConnectGarmin.stories.tsx
@@ -39,3 +39,15 @@ NotEnabledHide.args = { previewState: "notEnabled", title: "Garmin", disabledBeh
 
 export const NotEnabledDisplayError = Template.bind({});
 NotEnabledDisplayError.args = { previewState: "notEnabled", title: "Garmin", disabledBehavior: "displayError" };
+
+export const HideConnectedNotConnected = Template.bind({});
+HideConnectedNotConnected.args = { previewState: "notConnected", title: "Garmin", hideWhenConnected: true };
+
+export const HideConnectedUnauthorized = Template.bind({});
+HideConnectedUnauthorized.args = { previewState: "unauthorized", title: "Garmin", hideWhenConnected: true };
+
+export const HideConnectedFetchComplete = Template.bind({});
+HideConnectedFetchComplete.args = { previewState: "fetchComplete", title: "Garmin", hideWhenConnected: true };
+
+export const HideConnectedFetchingData = Template.bind({});
+HideConnectedFetchingData.args = { previewState: "fetchingData", title: "Garmin", hideWhenConnected: true };
--- a/src/components/container/ConnectDevice/ConnectDevice.tsx
+++ b/src/components/container/ConnectDevice/ConnectDevice.tsx
@@ -117,7 +117,7 @@ export default function (props: ConnectDeviceProps) {
 		}
 	}
 
-	if (props.hideWhenConnected && deviceExternalAccount) {
+	if (props.hideWhenConnected && deviceExternalAccount && deviceExternalAccount?.status != "unauthorized") {
 		return null;
 	}
 
--- a/src/components/container/ConnectDevice/ConnectDevice.tsx
+++ b/src/components/container/ConnectDevice/ConnectDevice.tsx
@@ -117,7 +117,7 @@ export default function (props: ConnectDeviceProps) {
 		}
 	}
 
-	if (props.hideWhenConnected && deviceExternalAccount && deviceExternalAccount?.status != "unauthorized") {
+	if (props.hideWhenConnected && deviceExternalAccount && deviceExternalAccount?.status != "unauthorized" && deviceExternalAccount?.status != "error") {
 		return null;
 	}
 
--- a/src/components/container/ConnectEhr/ConnectEhr.stories.tsx
+++ b/src/components/container/ConnectEhr/ConnectEhr.stories.tsx
@@ -3,6 +3,7 @@ import { ComponentMeta, ComponentStory } from "@storybook/react"
 import Card from "../../presentational/Card"
 import Layout from "../../presentational/Layout"
 import ConnectEhr, { ConnectEhrProps } from "./ConnectEhr";
+import { truncateByDomain } from "recharts/types/util/ChartUtils";
 
 
 export default {
@@ -81,3 +82,12 @@ export const CustomConnectText = {
     render: render
 };
 
+export const HideConnectedEnabledConnected = {
+    args: { previewState: "enabledConnected", applicationUrl: "preview", hideWhenConnected: true },
+    render: render
+};
+
+export const HideConnectedEnabledNeedsAttention = {
+    args: { previewState: "enabledNeedsAttention", applicationUrl: "preview", hideWhenConnected: true },
+    render: render
+};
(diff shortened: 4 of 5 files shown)


src/components/container/ConnectFitbit/ConnectFitbit.stories.tsx
```diff
@@ -38,4 +38,16 @@ export const NotEnabledHide = Template.bind({});
...
 export const NotEnabledDisplayError = Template.bind({});
-NotEnabledDisplayError.args = { previewState: "notEnabled", title: "Fitbit", disabledBehavior: "displayError" };
\ No newline at end of file
+NotEnabledDisplayError.args = { previewState: "notEnabled", title: "Fitbit", disabledBehavior: "displayError" };
+
+export const HideConnectedFetchComplete = Template.bind({});
+HideConnectedFetchComplete.args = { previewState: "fetchComplete", title: "Fitbit", hideWhenConnected: true };
+
+export const HideConnectedNotConnected = Template.bind({});
+HideConnectedNotConnected.args = { previewState: "notConnected", title: "Fitbit", hideWhenConnected: true };
```
reneefromhold, line 47: I don't think this is needed bc it is covered by the Not Connected story

src/components/container/ConnectGarmin/ConnectGarmin.stories.tsx
```diff
@@ -38,4 +38,16 @@ export const NotEnabledHide = Template.bind({});
 NotEnabledHide.args = { previewState: "notEnabled", title: "Garmin", disabledBehavior: "hide" };
 
 export const NotEnabledDisplayError = Template.bind({});
-NotEnabledDisplayError.args = { previewState: "notEnabled", title: "Garmin", disabledBehavior: "displayError" };
\ No newline at end of file
+NotEnabledDisplayError.args = { previewState: "notEnabled", title: "Garmin", disabledBehavior: "displayError" };
+
+export const HideConnectedNotConnected = Template.bind({});
+HideConnectedNotConnected.args = { previewState: "notConnected", title: "Garmin", hideWhenConnected: true };
```
reneefromhold, line 44: Same here, bc it is covered by the Not Connected story

src/components/container/ConnectGarmin/ConnectGarmin.stories.tsx
```diff
@@ -38,4 +38,16 @@ export const NotEnabledHide = Template.bind({});
...
+
+export const HideConnectedUnauthorized = Template.bind({});
+HideConnectedUnauthorized.args = { previewState: "unauthorized", title: "Garmin", hideWhenConnected: true };
```
reneefromhold, line 47: This is covered under Unauthorized. It doesn't feel like a special case. I am open to discuss why this should be different

src/components/container/ConnectFitbit/ConnectFitbit.stories.tsx
reneefromhold: Same feeling on HideConnectedUnauthorized

src/components/container/ConnectGarmin/ConnectGarmin.stories.tsx
```diff
@@ -38,4 +38,16 @@ export const NotEnabledHide = Template.bind({});
...
+
+export const HideConnectedFetchComplete = Template.bind({});
+HideConnectedFetchComplete.args = { previewState: "fetchComplete", title: "Garmin", hideWhenConnected: true };
```
reneefromhold, line 50: nit; the new stories here are nearly the same as fitbit, but the order is different.

reneefromhold (approved):
A few comments, but overall lg 👍

src/components/container/ConnectGarmin/ConnectGarmin.stories.tsx
brandoncharnesky: This is the "test case" that was failing.  Previously, this would be hidden, despite the connection being in a "needs to reconnect" state.  After the logic change, this is no longer hidden when the "hideWhenConnected" flag is set.  

Now the that logic is changed, the behavior is identical in both Unauthorized cases, and it does seem to clutter up sidebar quite a bit with all of these "test cases" for this flag left in.  I'm not sure the right balance of covering all of the behind the scenes logic with keeping the UI crisp and functional for what the end user will want to see.

greinard: I think it makes sense to have the extra permutations, even if the outcome is identical to a different setup.  It confirms the new functionality.  If it wasn't here, there would be no real assertion that the new functionality works the way it does.

That said, I was also getting frustrated by having all the permutations written out, too, especially for complex components with many variables (and thus many permutations).  My solutions was to start using a single `Default` case, which just has configurable preview states and settings.  A storybook user can then go in and test any permutations they would like rather than relying on just the hard coded permutations.

Here are a couple examples:

https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/components/container/InboxItemList/InboxItemList.stories.tsx
https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/components/asthma/components/AsthmaLogEntryHeader/AsthmaLogEntryHeader.stories.tsx

src/components/container/ConnectFitbit/ConnectFitbit.stories.tsx
brandoncharnesky: These stories have been combined

src/components/container/ConnectGarmin/ConnectGarmin.stories.tsx
brandoncharnesky: This is a great solution and I have implemented it for the hideWhenConnected uses cases.  There is now only one story per component, with a radio select for the multiple options of previewState.  

In order for this update to be view in real time in storybook, it was needed to add `props.previewState` to the list of variable dependencies of the useEffect function.  Is this the standard way to ensure this functionality works in storybook, or is there a preferred method?

brandoncharnesky:
Minor updates to the stories, and updates to UseEffect

src/components/container/ConnectFitbit/ConnectFitbit.stories.tsx
```diff
@@ -40,14 +40,6 @@ NotEnabledHide.args = { previewState: "notEnabled", title: "Fitbit", disabledBeh
...
-HideConnectedNotConnected.args = { previewState: "notConnected", title: "Fitbit", hideWhenConnected: true };
-
-export const HideConnectedUnauthorized = Template.bind({});
-HideConnectedUnauthorized.args = { previewState: "unauthorized", title: "Fitbit", hideWhenConnected: true };
-
-export const HideConnectedFetchingData = Template.bind({});
-HideConnectedFetchingData.args = { previewState: "fetchingData", title: "Fitbit", hideWhenConnected: true };
+export const HideConnected = Template.bind({});
+HideConnected.args = { previewState: "fetchComplete", hideWhenConnected: true };
+HideConnected.argTypes = { previewState: { name: "Connection State", control: "radio", options: ["notConnected", "notEnabled", "unauthorized", "error", "fetchComplete", "fetchingData"]}};
```
reneefromhold, line 45: Are notEnabled and error new?

src/components/container/ConnectGarmin/ConnectGarmin.stories.tsx
```diff
@@ -40,14 +40,8 @@ NotEnabledHide.args = { previewState: "notEnabled", title: "Garmin", disabledBeh
 export const NotEnabledDisplayError = Template.bind({});
 NotEnabledDisplayError.args = { previewState: "notEnabled", title: "Garmin", disabledBehavior: "displayError" };
 
-export const HideConnectedNotConnected = Template.bind({});
-HideConnectedNotConnected.args = { previewState: "notConnected", title: "Garmin", hideWhenConnected: true };
+export const HideConnected = Template.bind({});
+HideConnected.args = { previewState: "fetchComplete", hideWhenConnected: true };
+HideConnected.argTypes = { previewState: { name: "Connection State", control: "radio", options: ["notConnected", "notEnabled", "unauthorized", "error", "fetchComplete", "fetchingData"]}};
```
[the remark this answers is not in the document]
brandoncharnesky: updated back to the original 4 options

src/components/container/ConnectFitbit/ConnectFitbit.stories.tsx
brandoncharnesky: The control currently has display messages for the `fetchComplete`, `fetchingData`, and `unauthorized`.  There is no specific message displayed for an `error` state.

I'm not sure if this is a state that is intended to be reached - or what the should be displayed if it is - as there is currently no logic to display anything particular for an "error" state.  I've removed the logic handling the "error" state, as there are no stories regarding it in any configuration.  

@greinard Did you have any feedback on that the `error` state?

greinard: I think we should probably treat the `error` state as one that "needs attention" and therefore causes the widget to show up, including when `hideWhenConnected` is `true`.  If we are not displaying an error message of some kind on that widget when it is in the `error` state, perhaps we should be?

It is a valid status from the server: https://github.com/CareEvolution/Consumers/blob/master/Mogul/Libraries/RKStudioLibrary/Models/ExternalAccountPrimitives.cs#L36-L42

And we treat it as "needing attention" in the Connect EHR widget: https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/components/container/ConnectEhr/ConnectEhr.tsx#L63

src/helpers/strings-en.ts
```diff
@@ -23,6 +23,7 @@
     "connect-ehr-not-enabled": "EHR linking is not enabled for this project.",
     "search-for-provider": "Search for Provider or Health Plan",
     "expired-reconnect": "Connection expired.  Please reconnect.",
+    "connect-error-reconnect": "Unexpected error.  Please reconnect.",
```
reneefromhold, line 26: Nice 👍

PR #154, opened 2023-11-13
EHR News Feed for Lab Reports / Reports / Immunizations / Procedures

chrisnowak:
## Overview

This adds an out of box "EhrNewsFeed" set of components to be used for browsing the history of lab results, reports, immunizations, or procedures.  There are also some supporting components that have been added.

There are other potential types of information that can be displayed in the feed, but for now this just encompasses the types of information displayed in the out of box Health & Wellness Dashboard.

The components that have been added are:
- Container/EhrNewsFeed: This is the main component which displays a searchable, reverse chronological list of Procedures, Immunizations, LabReports, or Reports
- Container/EhrNewsFeedDetail: A view to be displayed when one of the items on the EhrNewsFeed is selected
  - I am conflicted about whether I should just have ProceduresView/LabReportsView/ReportsView/ImmunizationsView to be consistent with allergies/conditions/medications
- Presentational/ReportContent: A simple view for displaying report content.  If the content is html, it will render in an iframe.  If it is a PDF, it will present a button to download the PDF.
- Presentational/StatBlock: Used for displaying a list of attributes about something
- View/EhrNewsFeedView: View wrapper for the EhrNewsFeed with a navigation bar etc.
- View/EhrNewsFeedDetailView: View wrapper for EhrNewsFeedDetail
- View/ReportView: View for displaying report detail information

The goal is to have the entire Health and Wellness Dashboard in MyDataHelpsUI.  This is the largest remaining chunk of work to achieve that.  However, there is definitely some followup work remaining to complete this, in particular:
- Term information view
- Customizing the events/actions for the out of box H&W dashboard (getting all the various links set up to new versions)

Screenshot 2023-11-13 at 5 02 37 PM
Screenshot 2023-11-13 at 5 02 21 PM
Screenshot 2023-11-13 at 5 02 14 PM
Screenshot 2023-11-13 at 5 01 49 PM
Screenshot 2023-11-13 at 5 01 37 PM

## Security

REMINDER: All file contents are public.

- [x] I have ensured no secure credentials or sensitive information remain in code, metadata, comments, etc.
  - [x] Please verify that you double checked that .storybook/preview.js does not contain your participant access key details. 
  - [x] There are no temporary testing changes committed such as API base URLs, access tokens, print/log statements, etc.
- [x] These changes do not introduce any security risks, or any such risks have been properly mitigated.

Describe briefly what security risks you considered, why they don't apply, or how they've been mitigated.

### Testing
- [ ] MyDataHelps iOS app
[...]

PR #239, opened 2024-04-04
EHR News Feed, "Standalone" Health and Wellness View

chrisnowak:
## Overview

Replaces https://github.com/CareEvolution/MyDataHelpsUI/pull/154

This adds an out of box "EhrNewsFeed" set of components to be used for browsing the history of lab results, reports, immunizations, or procedures. There are also some supporting components that have been added.

This also adds a "Standalone" version of the Health and Wellness View which can be used outside MyDataHelps - it essentially hosts it's own little navigation system to be able to push/pop views.

For the news feed, there are other potential types of information that can be displayed in the feed, but for now this just encompasses the types of information displayed in the out of box Health & Wellness Dashboard.

The components that have been added are:

Container/EhrNewsFeed: This is the main component which displays a searchable, reverse chronological list of Procedures, Immunizations, LabReports, or Reports
Container/EhrNewsFeedDetail: A view to be displayed when one of the items on the EhrNewsFeed is selected
I am conflicted about whether I should just have ProceduresView/LabReportsView/ReportsView/ImmunizationsView to be consistent with allergies/conditions/medications
Container/ExternalAccountConnectionAlert: A little alert that displays if any external accounts have an issue or need to be reconnected.  
Container/TermInformation: Fetches term information from medline plus for the supplied term
Presentational/ReportContent: A simple view for displaying report content. If the content is html, it will render in an iframe. If it is a PDF, it will present a button to download the PDF.
Presentational/StatBlock: Used for displaying a list of attributes about something
View/EhrNewsFeedView: View wrapper for the EhrNewsFeed with a navigation bar etc.
View/EhrNewsFeedDetailView: View wrapper for EhrNewsFeedDetail
View/ReportView: View for displaying report detail information
View/TermInformationView: View for displaying term information from medline plus
View/StandaloneHealthAndWellnessView: The new "standalone" H&W view

### Standalone View Notes
The standalone view is to support hosting the H&W dashboard in an iframe inside another application; so there's no MyDataHelps to handle the navigation and the H&W dashboard must still work to be able to navigate and drill into things.

We will likely have more use cases like this coming up.  At some point it may make sense to develop a separate harness or generalize a solution here; but I think that's premature at this point.  

One hacky thing that's done to achieve this is to override MyDataHelps.back and MyDataHelps.close - but these standalone views would never be used on the page with a non-standalone view, so that should be fine.

This also wasn't too challenging to implement overall; there were a couple components where I had to add onClick handlers instead of urls to link out to (e.g. the ConnectEhr component) but that is likely the correct thing to be doing anyways.

## Security

This should not have any security impact.  There is only 1 place that we call "dangerouslySetInnerHTML" but that is just for a translated html string that is in the repo, so not dynamic or user supplied content.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


ashleyrtate (commented):
I noticed a few issues while reviewing the storybooks:
- EhrNewsFeed - Storybooks for Live Procedures, Immunizations, Report, Lab Reports are all throwing handler errors and seem to not fully load
- EhrNewsFeedEventDetail - Storybook for Lab Report help ? throws handler errors
- HealthPreviewSection - Some storybooks not loading, Default Medications details throws click handler errors
- TermInformation - Live storybook not loading
- ViewEhr - storybook "View" click handler erroring

I think a lot of these are just missing click handlers that could be stubbed with a not implemented popup but some seem to prevent the preview from loading.

I know this code has migrated a few times so not quite sure how much of this is brand new vs migrated code. I expect mainly the stand alone wrapper views?

Are you planning to expose the standalone views in MyDataHelpsViewLibrary?

ashleyrtate:
That would take care of the first bullet item, but the rest all seem to be other errors.

chrisnowak:
> I know this code has migrated a few times so not quite sure how much of this is brand new vs migrated code. I expect mainly the stand alone wrapper views?

Yeah I mean, it's a complicated lineage here; this is mostly new code though including the EHR news feed since I did the work here of porting it to react.  I certainly probably reused a couple bits of CSS but not too much since I was able to use mostly our out of box MyDataHelpsUI components as building blocks

> Are you planning to expose the standalone views in MyDataHelpsViewLibrary?

That is the plan - and then I will also update the various hardcoded url references here to update MyDataHelpsViewLibrary instead of hw.careevolutionapps.com

chrisnowak:
> That would take care of the first bullet item, but the rest all seem to be other errors.

Yep, i'll take a look at the others

src/components/container/EhrNewsFeed/EhrNewsFeed.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+export interface EhrNewsFeedProps {
+    previewState?: "default" | "procedures" | "labReports" | "immunizations" | "reports"
+    onEventSelected(eventReference: EhrNewsFeedEventReference): void
+    feed: EhrNewsFeedFeed
+    onReportSelected(reportID: string): void
+}
+
+export interface EhrNewsFeedEventReference {
+    feed: EhrNewsFeedFeed
```
ashleyrtate, line 22: FeedFeed is awkward and looks like a mistake. Maybe EhrNewsFeedType? Also might consider not pluralizing these. Could we maybe just use the EhrNewsFeedEventTypes here?

src/components/container/EhrNewsFeedEventDetail/EhrNewsFeedEventDetail.tsx
```diff
@@ -0,0 +1,156 @@
...
+        subtitle={subtitle}
+        icon={<img src={handler.icon} width={24} />} />;
+}
+
+function ProcedureGroupDetail(props: { event: EhrNewsFeedEventModel }) {
+    let procedures = props.event.Event as EhrNewsFeedProcedureModel[];
+    return <>
+        {procedures.map((procedure, index) =>
+            <Card key={procedure.ID} style={{ marginTop: "0" }}>
+                <Title defaultMargin order={5}>{procedure.Procedure}</Title>
```
ashleyrtate, line 83: The use of the order property here is a  bit obscure and could benefit from some contextual documentation. Also think the simple enum pattern does not work as well with numeric values as it does with strings because they aren't self documenting. Maybe would consider a true enum for that order property.

src/components/container/ExternalAccountConnectionAlert/ExternalAccountConnectionAlert.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    useInitializeView(() => {
+        if (props.previewState === 'externalAccountNoIssue') {
+            setExternalAccounts([{
+                id: 1,
+                provider: {
+                    id: 1,
+                    name: "Provider",
+                    category: "Provider",
+                    logoUrl: "https://google.com"
```
ashleyrtate, line 28: Placeholder that needs to be replaced? Also below.

src/components/container/ExternalAccountsPreview/ExternalAccountsPreview.tsx
```diff
@@ -64,7 +65,11 @@ export default function (props: ExternalAccountsPreviewProps) {
         if (props.applicationUrl == "preview") {
             console.log("PREVIEW: Opening the external accounts application.");
         } else {
-            MyDataHelps.openApplication(props.applicationUrl);
+            if (props.onClick) {
```
ashleyrtate, line 68: Could also remove the "preview" here and use the click handler for stories..

src/components/container/ProviderSearch/ProviderSearch.stories.tsx
```diff
@@ -36,8 +36,19 @@ onProviderSelected.parameters = {
...
 }
 Live.parameters = {
 	externalAccounts: []
+}
+
+export const LiveStandalone = Template.bind({});
+LiveStandalone.args = {
+	connectExternalAccountOptions: {
+		openNewWindow: false,
+		standaloneModeFinalRedirectPath: "https://mydatahelps.org"// replace with actual redirect url for this to work
```
ashleyrtate, line 49: Should this be a todo?

src/helpers/news-feed/types.ts
```diff
@@ -0,0 +1,247 @@
+export type EhrNewsFeedFeed = "Immunizations" | "LabReports" | "Procedures" | "Reports";
+
+export interface EhrNewsFeedPageModel {
+    NextPageID?: string
```
ashleyrtate, line 4: So these are coming back from the server...but do we want to do the tolowercase() conversion with these? I'm a bit confused by the casing standards we are trying to follow with some of these model objects.

ashleyrtate (commented):
I made a pass over this and had a few suggestions. Probably could use one more pair of eyes since there's a lot here.

src/components/container/ConnectEhr/ConnectEhr.tsx
```diff
@@ -72,7 +73,11 @@ export default function (props: ConnectEhrProps) {
 		if (props.applicationUrl == "preview") {
 			console.log("PREVIEW: Opening the connect EHR application.");
 		} else {
-			MyDataHelps.openApplication(props.applicationUrl);
+			if (props.onClick) {
```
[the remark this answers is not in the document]
chrisnowak: Good call

src/components/container/EhrNewsFeed/EhrNewsFeed.tsx
chrisnowak: Type definitions are really supposed to be singular (assuming they represent a singular concept)...like any class or object.  Not sure why we would want to pluralize it?

But yeah I like FeedType better, updated to that

src/components/container/EhrNewsFeedEventDetail/EhrNewsFeedEventDetail.tsx
chrisnowak: This Title component prop names were borrowed from some other UI frameworks so it seems conventional, e.g. https://mantine.dev/core/title/

Not sure what you mean about not being self-documenting though, it will show red squigglies etc. if you try to use one not in the list...

src/components/container/ProviderSearch/ProviderSearch.stories.tsx
chrisnowak: I think we don't have an actual published / live redirect url we have shared here, so not really a TODO, more if you are doing local dev here is where you would put in your redirect url

src/components/container/ExternalAccountsPreview/ExternalAccountsPreview.tsx
chrisnowak: Done

src/components/container/EhrNewsFeed/EhrNewsFeed.tsx
ashleyrtate: I was talking about the EhrNewsFeedEventType enum values: 
"procedure" | "labReport" | "immunization" | "report"
Not renaming the type.

src/helpers/news-feed/types.ts
chrisnowak: Yeah the server has uppercase names for these.  I think not worth the effort to try to coerce them lower just for the JS/TS side of things.  This is not a formally documented API or SDK method; if we did make it one then I think we would tackle that then (or maybe we convert these visualizations to use FHIR...)

src/components/container/ExternalAccountConnectionAlert/ExternalAccountConnectionAlert.tsx
chrisnowak: Fixed

src/components/container/EhrNewsFeed/EhrNewsFeed.tsx
chrisnowak: Oh - those are actual parameters sent to the server.  Renaming them would actually require a server config change etc. (duplicating a lot of config probably, because it would need to be backwards compatible)...so not something worth doing I think

src/components/container/EhrNewsFeedEventDetail/EhrNewsFeedEventDetail.tsx
ashleyrtate: Ah, when I looked at the class I saw this:
```
    if (props.image && props.autosizeImage) {
        imageWidth = 24;
        switch (props.order) {
            case 1:
                imageWidth = 40;
                break;
            case 2:
                imageWidth = 30;
                break;
            case 3:
                imageWidth = 24;
                break;
            case 4:
                imageWidth = 20;
                break;
            case 5:
                imageWidth = 16;
                break;
            case 6:
                imageWidth = 12;
                break;
        }
        if (props.imageAlignment == "top") {
            imageWidth = imageWidth * 2;
        }
    }
```
And did not realize order meant title level. Thought it was just some weird formatting convention. 

By not self documenting, I mean there is no text with implied usage like you have for a string constant.

src/components/container/EhrNewsFeed/EhrNewsFeed.tsx
ashleyrtate: Makes sense. Wasn't sure of those constraints.

src/components/container/TermInformation/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from "./TermInformation";
```
greinard, line 1: Minor: You could probably export `TermInformationReference` from here as well, to shorten the its imports a little.

src/components/container/ConnectEhr/ConnectEhr.stories.tsx
```diff
@@ -52,32 +52,32 @@ export const NotEnabledDisplayError = {
 };
 
 export const Enabled = {
-    args: { previewState: "enabled", applicationUrl: "preview" },
+    args: { previewState: "enabled", onClick: () => { console.log("PREVIEW: Opening the connect EHR application.") } },
```
greinard, line 55: Minor:  Could probably define a single const function and just pass it to all these.

src/components/container/EhrNewsFeed/EhrNewsFeed.css
```diff
@@ -0,0 +1,57 @@
...
+}
+
+.mdhui-news-feed .mdhui-news-feed-list-item .mdhui-news-feed-list-item-date {
+    font-size: .7em;
+    color: var(--mdhui-text-color-3);
+}
+
+.mdhui-news-feed .mdhui-news-feed-list-item .mdhui-news-feed-list-item-acuity {
+    display: inline-block;
+    color: #FFF;
```
greinard, line 53: Maybe use a color variable here?

src/components/container/EhrNewsFeedEventDetail/EhrNewsFeedEventDetail.tsx
```diff
@@ -0,0 +1,156 @@
...
+}
+
+function NewsFeedDetailTitle(props: { event: EhrNewsFeedEventModel }) {
+    let handler = eventTypeDefinitions[props.event.Type];
+
+    let title = handler.getDetailTitle ? handler.getDetailTitle(props.event) : undefined;
+    if (!title) { return null; }
+
+    let date = parseISO(props.event.Date);
+    let subtitle = `${format(parseISO(props.event.Date), "EEEE, MMMM do, y h:mm a")} • ${props.event.Patient.RecordAuthority}`
```
brandoncharnesky, line 66: Are we able to localize this for Spanish?

src/components/container/EhrNewsFeed/EhrNewsFeed.tsx
```diff
@@ -0,0 +1,194 @@
...
+                <LoadingIndicator />
+            }
+            <OnVisibleTrigger onTrigger={() => loadMore()} enabled={!loading && !finished} />
+        </div>
+    )
+}
+
+function NewsFeedListItem(props: { event: EhrNewsFeedEventModel, onClick: (event: EhrNewsFeedEventModel) => void }) {
+    let definition = eventTypeDefinitions[props.event.Type];
+    let date = format(parseISO(props.event.Date), "h:mm a");
```
brandoncharnesky, line 167: Language/Localization possible here?

brandoncharnesky (approved):
Looks good.  Noted two spots where it may be possible to localize/translate dates.

ClaysGit:
@CodiumAI-Agent /review

ClaysGit:
@coderabbitai review

src/components/container/EhrNewsFeed/EhrNewsFeed.css
chrisnowak: In this case it's always on a dark background regardless of light / dark mode

src/components/container/EhrNewsFeedEventDetail/EhrNewsFeedEventDetail.tsx
chrisnowak: Good call, done.  I think we do probably need to do a pass in MyDataHelpsUI to ensure we use locale-specific date formats, like not using MM/dd/yyyy in europe, probably need to rely a bit more on the out of box date formats to achieve that rather than format strings

src/components/container/EhrNewsFeed/EhrNewsFeed.tsx
chrisnowak: Done

PR #244, opened 2024-04-17
Add six additional languages and locales

brandoncharnesky:
## Overview

Adds additional languages and locales.  Addresses https://github.com/CareEvolution/MyDataHelpsViewBuilder/issues/67

GetLanguageFromIso and GetLocaleFromIso have been moved into helper functions.

Translations done via chatGPT 3.5 and may need further review.  

LanguagesEnglish
LanguagesSpanish
LanguagesPolish
LanguagesItalian
LanguagesPortuguese
LanguagesFrench
LanguagesGerman
LanguagesDutch

Default language support using navigator language (if supported, English fallback if not):

LanguagesDefault

## Security

All changes are front end only and do not introduce security risks.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app
- [x]  Tested only via MDH-UI

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/helpers/language.ts
```diff
@@ -1,17 +1,51 @@
...
+export function getLanguageFromIso(language: string): "" | Language {
+    if (language.toLowerCase().startsWith("es")) return "es";
+    if (language.toLowerCase().startsWith("nl")) return "nl";
+    if (language.toLowerCase().startsWith("de")) return "de";
+    if (language.toLowerCase().startsWith("fr")) return "fr";
+    if (language.toLowerCase().startsWith("pt")) return "pt";
+    if (language.toLowerCase().startsWith("it")) return "it";
+    if (language.toLowerCase().startsWith("pl")) return "pl";
+
+    return "";
```
reneefromhold, line 48: Should this default to en?

reneefromhold (approved):
Just one question. LG otherwise

src/helpers/language.ts
brandoncharnesky: I have reworked this section to better handle this better.  The empty string is the "default" setting, and will now cause it to look at the user's language from their browser.

src/helpers/language.ts
```diff
@@ -1,17 +1,44 @@
...
-	} else {
-		if (spanishStrings[key] != null) {
-			return spanishStrings[key];
-		} else {
-			return englishStrings[key];
-		}
-	}
+	const currentLanguage: Language = getLanguageFromIso(MyDataHelps.getCurrentLanguage());
+
+	if (currentLanguage == "en") return englishStrings[key];
```
chrisnowak, line 16: I think we want to fall back to english if the individual string isn't defined in the target language; see the previous logic where it actually checked to see if the string was there...so it should probably be something like:
```
let resolvedString = null;
if(currentLanguage == "es") { resolvedString = spanishStrings[key]; }
//etc

if(!resolvedString){
   resolvedString = englishStrings[key];
}
return resolvedString;
```

src/helpers/language.ts
```diff
@@ -1,17 +1,44 @@
...
+	if (currentLanguage == "pl") return polishStrings[key];
+
+	return englishStrings[key];
+}
+
+export function getLanguageFromIso(language: string): Language {
+    if (language.length < 2) return "";
+
+    var beginningOfLanguage = language.slice(0, 2);
+    if (beginningOfLanguage == "en") return "en";
```
chrisnowak, line 32: Feels like there may be some ways of simplifying this by making Language an enum so you could enumerate the values and check to see if beginningOfLanguage is in that enum, otherwise return "" or something?  

Anyways, not a big deal, but it would be nice to have to touch fewer places when adding more languages in the future

chrisnowak (approved):
Looks great overall.  Might be worth playing around just a bit with the idea of trying to touch as few places / spots as possible when we add a language...

e.g. maybe a languageConfiguration lookup that has the strings and locale for each language keyed by their 2 letter code

But not a big deal for now

src/helpers/language.ts
brandoncharnesky: I do think long term once we get >10 languages this will be the way to go.

Issue #101, opened 2023-08-11
Combined Daily Data Provider Issues


ajmenca:
- The combined daily data provider for RestingHeartRate checks data availability for AppleHealth, then Fitbit, then Garmin.  The `combinedRestingHeartRate` provider function, however, only actually queries for AppleHealth and Fitbit data.  If you were to use this for a participant that only had Garmin data, it wouldn't return valid data.

- Both the combined resting heart rate provider and the combined steps provider are not robust against a Project not having one of the data types enabled.  When a project doesn't have a data type enabled, the server returns a 500 error response; for the `simpleAvailabilityCheck` this is caught properly and returns false.  However the combined data providers don't know which providers are valid, and also do not `catch` errors from their data query, so they fail outright.  This means that if, for example, you enable Apple Health and Fitbit, but not Garmin, you can't actually use the Combined Steps aggregator (in this specific case you can use the combined RHR function because of the omission of the Garmin query noted above).

ajmenca:
My second bullet point here was resolved in #112.

brandoncharnesky:
First bullet point is resolved by:  https://github.com/CareEvolution/MyDataHelpsUI/pull/245

PR #245, opened 2024-04-17
RestingHeartRate - add garmin

brandoncharnesky:
## Overview

Updates the combined RestingHeartRate data provider to look at Garmin data, as well as AppleHealth and Fitbit.  It will then average and round the combined results for each day.  ~~Added stories to the Resting Heart Rate Calendar container for each three individual data sources as well as the combined data source.~~  Added a "Live" story that defaults to the Combined datasource, with options to selected a specific datasource.

Addresses https://github.com/CareEvolution/MyDataHelpsUI/issues/101

Test data for these screenshots was: no AppleHealth rhr data, fitbit data for one day, and garmin data for two days.

image
image
image
image

## Security

No security risks, UI changes only.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app
- [x]  Tested via MDH-UI website only

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.stories.tsx
+++ b/src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.stories.tsx
@@ -39,3 +39,33 @@ SampleData.args = {
   month: 3,
   showPreviewData: "WithData",
 };
+
+export const Combined = Template.bind({});
+Combined.args = {
+  month: new Date().getMonth(),
+  year: new Date().getFullYear(),
+  showPreviewData: "Combined",
+};
+
+export const AppleHealth = Template.bind({});
+AppleHealth.args = {
+  month: new Date().getMonth(),
+  year: new Date().getFullYear(),
+  showPreviewData: "AppleHealth",
+};
+
+export const Fitbit = Template.bind({});
+Fitbit.args = {
+  month: new Date().getMonth(),
+  year: new Date().getFullYear(),
+  showPreviewData: "Fitbit",
+};
+
+export const Garmin = Template.bind({});
+Garmin.args = {
+  month: new Date().getMonth(),
+  year: new Date().getFullYear(),
+  showPreviewData: "Garmin",
+};
+
+
--- a/src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.stories.tsx
+++ b/src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.stories.tsx
@@ -23,6 +23,13 @@ export const Default = Template.bind({});
 Default.args = {
   month: new Date().getMonth(),
   year: new Date().getFullYear(),
+  showPreviewData: "WithData",
+};
+
+export const NoData = Template.bind({});
+NoData.args = {
+  month: new Date().getMonth(),
+  year: new Date().getFullYear(),
   showPreviewData: "NoData",
 };
 
@@ -33,20 +40,13 @@ Loading.args = {
   showPreviewData: "Loading",
 };
 
-export const SampleData = Template.bind({});
-SampleData.args = {
-  year: 2022,
-  month: 3,
-  showPreviewData: "WithData",
-};
-
 export const Live = Template.bind({});
 Live.args = {
   month: new Date().getMonth(),
   year: new Date().getFullYear(),
-  showPreviewData: "Live",
   dataTypeSource: "Combined"
 };
+
 Live.argTypes = {
   dataTypeSource: {
     name: 'dataTypeSource',
--- a/src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.stories.tsx
+++ b/src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.stories.tsx
@@ -40,32 +40,18 @@ SampleData.args = {
   showPreviewData: "WithData",
 };
 
-export const Combined = Template.bind({});
-Combined.args = {
+export const Live = Template.bind({});
+Live.args = {
   month: new Date().getMonth(),
   year: new Date().getFullYear(),
-  showPreviewData: "Combined",
+  showPreviewData: "Live",
+  dataTypeSource: "Combined"
 };
-
-export const AppleHealth = Template.bind({});
-AppleHealth.args = {
-  month: new Date().getMonth(),
-  year: new Date().getFullYear(),
-  showPreviewData: "AppleHealth",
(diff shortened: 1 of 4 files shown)


src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.tsx
```diff
@@ -10,7 +10,7 @@ import { queryDailyData, DailyDataType } from '../../../helpers/query-daily-data
 
 type HeartRateMap = { [key: string]: number };
 
-export type RestingHeartRateCalendarPreviewState = "WithData" | "NoData" | "Loading";
+export type RestingHeartRateCalendarPreviewState = "WithData" | "NoData" | "Loading" | "Combined" | "AppleHealth" | "Fitbit" | "Garmin";
```
chrisnowak, line 13: This definitely doesn't seem right; the "previewState" (or showPreviewData in this case) params are just for providing a preview of non-live data so I wouldn't conflate restricting it to Garmin / Fitbit / AppleHealth with this

I would add another prop for "namespace" that is "AppleHealth" | "Fitbit" | "Garmin" | "Combined" that allows you to control the namespace; the prop should default to "Combined"

chrisnowak:
The data provider change looks good, mostly think we need to tweak the RestingHeartRateCalendar a bit per my comment

brandoncharnesky:
> The data provider change looks good, mostly think we need to tweak the RestingHeartRateCalendar a bit per my comment

Good catch, I've created a new type to specify the data source.  I also combined these into a single "Live" story.

src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.stories.tsx
```diff
@@ -38,4 +38,20 @@ SampleData.args = {
   year: 2022,
   month: 3,
   showPreviewData: "WithData",
-};
\ No newline at end of file
+};
+
+export const Live = Template.bind({});
```
reneefromhold, line 43: @brandoncharnesky I recognize these items had already existed. Hoping they can be cleaned up as part of this PR. The Default and Sample Data stories do not align with the other components. 
I am advocating for
1) The current Default story be renamed to NoData
2) The Sample Data to be renamed to Default,

src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.tsx
```diff
@@ -10,12 +10,14 @@ import { queryDailyData, DailyDataType } from '../../../helpers/query-daily-data
...
 
-export type RestingHeartRateCalendarPreviewState = "WithData" | "NoData" | "Loading";
+export type RestingHeartRateCalendarPreviewState = "WithData" | "NoData" | "Loading" | "Live";
+export type RestingHeartRateDataSource = "Combined" | "AppleHealth" | "Fitbit" | "Garmin";
 
 export interface RestingHeartRateCalendarProps {
 	month: number,
 	year: number,
 	showPreviewData: RestingHeartRateCalendarPreviewState
+	dataTypeSource?: RestingHeartRateDataSource
```
reneefromhold, line 20: Advocating for 
1) RestingHeartRateDataSource to have Combined removed
2) The RestingHeartRateCalendarProps.dataTypeSource to be an array of RestingHeartRateDataSource 

This will allow for supporting a subset of devices instead of all/one

src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.tsx
```diff
@@ -10,12 +10,14 @@ import { queryDailyData, DailyDataType } from '../../../helpers/query-daily-data
 
 type HeartRateMap = { [key: string]: number };
 
-export type RestingHeartRateCalendarPreviewState = "WithData" | "NoData" | "Loading";
+export type RestingHeartRateCalendarPreviewState = "WithData" | "NoData" | "Loading" | "Live";
```
reneefromhold, line 13: The absence of a preview state would signal Live. The value can be removed from RestingHeartRateCalendarPreviewState

src/helpers/daily-data-providers/combined-resting-heart-rate.ts
```diff
@@ -1,42 +1,42 @@
...
-				var day = getDayKey(parseISO(d.startDate));
-				var value = parseFloat(d.value);
-				if (!dayValues[day]) {
-					dayValues[day] = [];
-				}
-				dayValues[day].push(value);
-			});
+	let providers: Promise<{ [key: string]: number }>[] = [];
+
+	return MyDataHelps.getDataCollectionSettings().then((settings) => {
```
reneefromhold, line 9: Is there a way to elegantly fail the Promise on line 24 instead of checking the settings each time around?

reneefromhold (changes requested):
Left some comments

src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.tsx
brandoncharnesky: I am curious about the use case for supporting this.  This data provider wants to return all available data.

src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.stories.tsx
brandoncharnesky: Good catch, I have renamed the stories.  The first story is now Default, which loads the sample data.  The NoData case is now named NoData.

src/components/container/RestingHeartRateCalendar/RestingHeartRateCalendar.tsx
reneefromhold: See here for an example https://github.com/CareEvolution/MyDataHelpsUI/blob/56f41e637154b522b43a888119e49fb86404cc69/src/components/container/BloodPressureVisualization/BloodPressureVisualization.tsx#L21

brandoncharnesky: The current framework for daily data providers is  straightforward, returning data without specifying options.  I think there is a future opportunity to create a new type of data provider, that would allow you to specify specific data sources that are rolled up together.  However, for this current work, returning all available data is what is expected.  I will open a new issue about generalizing this type of combined data provider.

reneefromhold (approved):

PR #247, opened 2024-04-24
Heart Rate Chart

reneefromhold:
## Overview

This line chart will display the heart rate data from Fitbit Heart Rate Intraday, or AppleHealth via the DeviceDataV2 endpoints.

- DataSources are AppleHealth and Fitbit
- AggregateOptions are min, max, count, avg, sum
- aggregationIntervalMinutes- the number of minutes to aggregate on
- LineColor - the color of the line.
- Thresholds - For each threshold defined, a reference line will be displayed and the color of the data line will render in the color defined. 

## Security

  - [ x] Please verify that you double checked that .storybook/preview.js does not contain your participant access key details. 
  - [ x] There are no temporary testing changes committed such as API base URLs, access tokens, print/log statements, etc.
- [ x] These changes do not introduce any security risks, or any such risks have been properly mitigated.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app
- [x ] [MyDataHelps website](mydatahelps.org)

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

## Summary by CodeRabbit

- **New Features**
  - Introduced `IntradayHeartRateChart` component to display heart rate data over specified intervals.
  - Added new chart customization options including showing/hiding dots and setting domain max for multi-series charts.

- **Enhancements**
  - Improved `TimeSeriesChart` with new line chart customization options and color logic.
  - Added support for gradient definitions in line charts.

- **Bug Fixes**
  - Updated day tick logic in `TimeSeriesChart` to improve color resolution.

- **Documentation**
  - Updated and refactored storybook examples for `TimeSeriesChart`.

- **Chores**
  - Updated `@careevolution/mydatahelps-js` package to version 3.15.0.


chrisnowak:
The x-axis here should be like the original mockup chart and just be spaced 3 hours apart with labels at 3am, 6am, 9am, 12pm, 3pm, 6pm, 9pm:
image

Additionally, I think a lot of the formatting / styling that I shared previously is not applied - e.g. this is showing solid black borders...my sense was that all we had to do was take the code from my branch and pipe in the data and fix up the gradient/threshold stuff

src/components/presentational/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+export interface LineChartProps {
+    data: LineChartDataPoint[],
+    axis?: LineChartAxis,
+    lineColor?: ColorDefinition,
+    thresholds?: LineChartThreshold[],
+    innerRef?: React.Ref<HTMLDivElement>
+}
+
+export default function MdhLineChart(props: LineChartProps) {
```
chrisnowak, line 37: I definitely wouldnt' call this "LineChart" as that is far too vague of a name.  I think we should make this at least somewhat opininated.  

Putting aside whether it makes sense to combine this with @ajmenca 's PR, I would propose that this be called "IntradayLineChart" and be hardcoded to a 24 hour interval with the x-axis as I described and it could take data as data points in that interval...do not think it makes sense to make a generally reusable "line chart" as that is what recharts already gives us

src/components/presentational/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,139 @@
...
+    if (!props.axis || (props.axis && !props.axis.yRange)) {
+        const yValues: number[] = data.map(d => d.value);
+        const yMax = Math.max(...yValues);
+        axis.yRange = [0, yMax];
+    }
+    if (!props.axis || (props.axis && !props.axis?.xRange)) {
+        const xValues: any[] = data.map(d => d.date.getTime());
+        const xMax = Math.max(...xValues);
+        const xMin = Math.min(...xValues);
+        axis.xRange = [xMin, xMax];
```
chrisnowak, line 53: Axis should be hardcoded

src/components/presentational/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const thresholdKey = "_threshold";
+    const data: LineChartDataPoint[] = props.data;
+    let layoutContext = useContext(LayoutContext);
+    const defaultLineColor = resolveColor(layoutContext.colorScheme, props.lineColor) || '#bbb';
+
+    var axis: LineChartAxis = props.axis ?? { yRange: [0, 100], yIncrement: 5, xRange: [0, 100], xIncrement: 1 }
+    if (!props.axis || (props.axis && !props.axis.yRange)) {
+        const yValues: number[] = data.map(d => d.value);
+        const yMax = Math.max(...yValues);
+        axis.yRange = [0, yMax];
```
chrisnowak, line 47: Curious why you need to do this...I think just not supplying a domain will be sufficient

src/components/presentational/LineChart/LineChart.tsx
```diff
@@ -0,0 +1,139 @@
...
+                    </linearGradient>
+                }
+            </defs>
+            <Tooltip wrapperStyle={{ outline: "none" }} content={<GraphToolTip />} />
+            <CartesianGrid vertical strokeDasharray="2 4" />
+            {(props.thresholds)?.filter(t => t.referenceLineColor)?.map((threshold, index) =>
+                <ReferenceLine key={`threshref_${index}`} y={threshold.value} stroke={resolveColor(layoutContext.colorScheme, threshold.referenceLineColor)} />
+            )}
+            <YAxis tick={YAxisTick} type='number' domain={axis?.yRange ?? [0,100]} />
+            <XAxis tick={XAxisTick} axisLine={true} dataKey="date" tickMargin={0} minTickGap={0} tickCount={24}
```
chrisnowak, line 124: This is where I think you need to specify the "interval" to get the axis ticks to show up at 3/6/9 etc.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,102 @@
...
+    value: number
+    referenceLineColor?: ColorDefinition
+    overThresholdLineColor?: ColorDefinition
+}
+
+export interface IntradayHeartRateChartProps {
+    previewState?: IntradayHeartRatePreviewState,
+    dataSource: DeviceDataV2Namespace,
+    aggregationOption: AggregationOption,
+    aggregationIntervalAmount: number,
```
chrisnowak, line 25: Let's just call this "aggregationIntervalMinutes" and not have aggregationIntervalType...this is intraday so no need to complicate with extra params

src/helpers/heart-rate-data-providers/intraday-aggregates.tsx
```diff
@@ -0,0 +1,29 @@
+import MyDataHelps, { Guid, DeviceDataV2AggregateQuery, DeviceDataV2AggregatePage, DeviceDataV2Aggregate } from "@careevolution/mydatahelps-js";
+
+export default async function (query: DeviceDataV2AggregateQuery): Promise<DeviceDataV2Aggregate[]> {
```
chrisnowak, line 3: What does this have to do with heart rate data - this seems like a general "queryAllDeviceDataV2" function more than anything to do with heart rates?

chrisnowak: So seems like it would go in the same bucket as query all survey answers etc.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,102 @@
...
+    aggregationIntervalType: "Hours" | "Minutes",
+    lineColor?: ColorDefinition,
+    thresholds?: IntradayChartThreshold[],
+    innerRef?: React.Ref<HTMLDivElement>
+}
+
+export default function IntradayHeartRateChart(props: IntradayHeartRateChartProps) {
+    const [data, setData] = useState<LineChartDataPoint[]>([]);
+    const [axis, setAxis] = useState<LineChartAxis>({ yRange: [0, 200], yIncrement: 30, xIncrement: props.aggregationIntervalAmount });
```
chrisnowak, line 34: I don't think yRange needs to be specified at all or yIncrement - in my branch it worked fine without having to specify these

chrisnowak:
Something kind of strange - the line seems to fly in from the right as the animation?  I think we'd want to use whatever animation the default DailyDataChart uses for instance (which I think kind of comes up from the bottom)...

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+export interface IntradayChartThreshold {
+    value: number
+    referenceLineColor?: ColorDefinition
+    overThresholdLineColor?: ColorDefinition
+}
+
+export interface IntradayHeartRateChartProps {
+    previewState?: IntradayHeartRatePreviewState,
+    dataSource: DeviceDataV2Namespace,
```
chrisnowak, line 23: Per the spec this should accept multiple, e.g. Apple Health AND Fitbit:

> It should query data for all data sources specified as props; if there happens to be data for 2 data sources for an interval it should use the data source that is first in the array (e.g. Fitbit by default)

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    function extractStateData(data: DeviceDataV2Aggregate[]) {
+        let axis: LineChartAxis = { yRange: [0, 200], yIncrement: 30, xIncrement: props.aggregationIntervalAmount ?? 3 };
+        let dataPoints : LineChartDataPoint[] = data.map(transformToDeviceDataV2AggregateReduced);
+        dataPoints = dataPoints.sort((a, b) => a.date.getTime() - b.date.getTime());
+
+        if (dataPoints.length){
+            let yValues: number[] = dataPoints.map(d => d.value);
```
chrisnowak, line 81: What happens if you have a gap in the middle of the day e.g. where the user wasn't wearing their device?

chrisnowak:
Broadly still a little confused...my branch seems like it was a good chunk of the way there on the presentation layer...seemed mostly like it just needed:
- Querying the data from the aggregates
- Converting the data into the proper format; probably doing some math on the axes compared to the aggregate interval
- Fixing up the gradients
- Cleaning up a bit

This seems to have gone off in a a bit of a different direction and doesn't quite match the spec or the example screenshot?

chrisnowak:
Let's also figure out a way to get the line colors perfect - right now it obviously is a bit fuzzy - but the line should solidly change color when it is above the threshold...I can maybe help with that

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
reneefromhold: The value of having some default x or y axis in place is so the chart renders when there is no data. 
If nothing displays at all, the user may not be aware their device is is not connected. An empty chart could reinforce that the participant should connect their device

chrisnowak: I think we would want to mimic the "Daily data chart"'s no data state here which does not display a y-axis but does display the x-axis

chrisnowak: https://ui.mydatahelps.org/?path=/story/container-dailydatachart--no-data

src/components/presentational/LineChart/LineChart.tsx
reneefromhold: I do agree LineChart is too vague. 
What I do want to avoid is recreating the threshold interface and the display of it across components. We currently consume BarChartThreshold in vbuilder. Adding LineChartThreshold here is no help. Would be nice to have a simple ChartThreshold interface that is used across charts, along with not having to copy/paste the display of thresholds across components.

reneefromhold: I will integrate the feedback around making it more opinionated and hang tight on the Threshold stuff

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
reneefromhold: @chrisnowak The line will continue without breaks.

chrisnowak: Right - my point is that is not the right behavior though, right?  Since you need each heart rate data point to be represented at the time for the data point

reneefromhold:
> Something kind of strange - the line seems to fly in from the right as the animation? I think we'd want to use whatever animation the default DailyDataChart uses for instance (which I think kind of comes up from the bottom)...

This is now corrected when not using the date range navigator. When using the navigator the data will swing in from the same direction navigated to.

reneefromhold:
> Let's also figure out a way to get the line colors perfect - right now it obviously is a bit fuzzy - but the line should solidly change color when it is above the threshold...I can maybe help with that

@chrisnowak I've addressed most of the feedback, though getting the colors to sharply change at the threshold is still outstanding. Will work on that in next round of changes

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
reneefromhold: This is corrected, and a story is in place for showing the behavior (line break).

reneefromhold:
@chrisnowak The issue with line colors + incomplete data is fixed.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -129,10 +159,36 @@ export default function TimeSeriesChart(props: TimeSeriesChartProps) {
                 } else if (domainMin !== undefined) {
                     domain = [domainMin, "auto"];
                 }
```
brandoncharnesky, line 161: Is there an option for `domain = ["auto", domainMax]` ?

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.data.ts
```diff
@@ -0,0 +1,44 @@
...
+        testData[(new Date(data.date)).getTime()] = data;
+    });
+    return testData;
+}
+
+let testData: DeviceDataV2Aggregate[] = [];
+for (let i = 0; i < 48; i++) {
+    testData.push({ participantID: "ABCDEFG", participantIdentifier: "ABCDEFG", date: addMinutes(today, (i * 5)).toISOString(), statistics: { "avg": getRandomInt(80, 185) } });
+}
+const halfDayData = loadForTest(testData);
```
brandoncharnesky, line 24: Nit:  Can we generate data that more closely matches a typical day of heart rate data?  Maybe an eight hour period with consistent low numbers, an eight hour period with normal numbers, and an eight hour period with higher activity and a few spikes?

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+			<IntradayHeartRateChart {...args} />
+		</DateRangeCoordinator>
+	</Card>
+</Layout>;
+
+export const Default = {
+	render: renderDefault,
+	args: {
+		previewState: "Default",
+		aggregationOption: "avg"
```
brandoncharnesky, line 32: I'm unclear what the different aggregation options control and when it would appropriate to use each of them.  Is there some more information we can add?

brandoncharnesky:
Does this support light/dark mode?

src/helpers/chartOptions.ts
```diff
@@ -7,13 +7,13 @@ export interface LineChartOptions {
 
 export interface BarChartOptions {
     barColor?: ColorDefinition
-    thresholds?: BarChartThreshold[]
+    thresholds?: ChartThreshold[]
 }
 
-export interface BarChartThreshold {
+export interface ChartThreshold {
```
greinard, line 13: It looks like the `ChartThreshold` interface already exists below.

src/helpers/chartOptions.ts
```diff
@@ -31,8 +31,11 @@ export interface AreaChartSeries extends ChartSeries {
 } 
 
 export interface MultiSeriesLineChartOptions {
-    connectNulls?: boolean,
+    connectNulls?: boolean
+    showDots? : boolean
```
greinard, line 35: Could we make this something like `dot?: ((props: any) => ReactElement<SVGElement>) | boolean` so that it matches up more closely with the react property.  I have a need to be able to render custom dots.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -228,12 +284,16 @@ export default function TimeSeriesChart(props: TimeSeriesChartProps) {
             }
             {props.chartHasData && props.chartType === "Line" &&
                 <ResponsiveContainer width="100%" height={150}>
-                    <LineChart width={400} height={400} data={dataToDisplay} syncId="DailyDataChart">
+                    <LineChart width={400} height={400} data={dataToDisplay} syncId="DailyDataChart" margin={{ left: 5, top: 5, bottom: 5, right: 40 }}>
```
greinard, line 287: I think these margins were just removed in this PR:  https://github.com/CareEvolution/MyDataHelpsUI/pull/268/files#r1644688975

src/helpers/heart-rate-data-providers/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as combinedIntradayHeartRateDataProvider } from "./combined-intraday-heart-rate-providers";
+export { IntradayHeartRateAggregationOption as IntradayHeartRateAggregationOption } from "./combined-intraday-heart-rate-providers";
```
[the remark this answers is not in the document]
greinard: I agree.  No need for the `as ...` here.

Actually, I'd just move `IntradayHeartRateAggregationOption` up to the prior export.

```
export { default as combinedIntradayHeartRateDataProvider, IntradayHeartRateAggregationOption } from "./combined-intraday-heart-rate-providers";
```

src/helpers/chartOptions.ts
ajmenca: I left this as two distinct interfaces for backwards compatibility reasons.  Removing the `BarChartThreshold` and renaming `overThresholdBarColor` is a breaking change that I was avoiding when I created the `TimeSeriesChart`.  It will require code changes in View Builder and requires code changes for anyone who may have written code against this.  I know its annoying, but I'd advise against this unless we think it's absolutely necessary.

src/components/container/IntradayHeartRateChart/index.ts
```diff
@@ -0,0 +1 @@
+export {default as IntradayHeartRateChart} from "./IntradayHeartRateChart";
```
greinard, line 1: For consistency, this can just export `default` without the `as IntradayHeartRateChart`.  Then the `container/index.ts` can export `default as IntradayHeartRateChart`.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,133 @@
...
+    rawValue?: number
+}
+
+export default function IntradayHeartRateChart(props: IntradayHeartRateChartProps) {
+    const [loading, setLoading] = useState(true);
+    const [data, setData] = useState<IntradayDataPoint[] | undefined>(undefined);
+    const [options, setOptions] = useState<MultiSeriesLineChartOptions | undefined>(undefined);
+    const dateRangeContext = useContext(DateRangeContext);
+    const intervalStart = startOfDay(dateRangeContext?.intervalStart ?? new Date());
+    const intervalEnd = startOfDay(add(intervalStart, { days: 1 }));
```
greinard, line 39: Minor:  I think you can drop the `startOfDay` here since `intervalStart` will already be at the start of the day.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.data.ts
```diff
@@ -0,0 +1,48 @@
...
+const today = startOfDay(new Date());
+
+function addHour(date: Date, days: number) {
+    return add(date, { hours: days });
+}
+
+function getRandomInt(min : number, max : number) {
+    min = Math.ceil(min);
+    max = Math.floor(max);
+    return Math.floor(Math.random() * (max - min + 1)) + min;
```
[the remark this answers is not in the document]
greinard: Minor:  I think you can drop the `+1` in here if you use `Math.round(...)` instead of `Math.floor(...)`.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.data.ts
greinard, line 1: I think the naming convention we've been using for the most part for this sort of file is: `IntradayHeartRateChart.previewdata.ts`.

ajmenca:
> Does this support light/dark mode?

@brandoncharnesky The Threshold colors are all defined as `ColorDefinition` objects, and they're all going through the resolution code (see https://github.com/CareEvolution/MyDataHelpsUI/pull/247/files#diff-c7d85721f79b358b3ab7238e978537aa98ffcf42fc45afbb6a1eb497b473892fR133) so that all looks correct.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.stories.tsx
```diff
@@ -0,0 +1,103 @@
+import React from "react";
+import { Card, DateRangeCoordinator, Layout } from "../../presentational";
+import IntradayHeartRateChart, { IntradayHeartRateChartProps } from "./IntradayHeartRateChart";
+
+
```
greinard, line 5: Minor:  Extra newline.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.stories.tsx
```diff
@@ -0,0 +1,103 @@
...
+export default {
+	title: "Container/IntradayHeartRateChart",
+	component: IntradayHeartRateChart,
+	parameters: {
+		layout: 'fullscreen',
```
greinard, line 10: Minor:  Extra `,`.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,133 @@
...
+            });
+        }
+    }
+
+    function transformToChartData(data: IntradayHeartRateData) {
+        const start = intervalStart;
+        const end = intervalEnd;
+        const interval: Interval = { start, end };
+        const options: any = { step: 5 };
+        var expectedIntervals = eachMinuteOfInterval(interval, options);
```
greinard, line 81: Minor:  I think you can probably just inline all this:

`let expectedIntervals = eachMinuteOfInterval({ intervalStart, intervalEnd }, { step: 5 });`

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.data.ts
```diff
@@ -0,0 +1,44 @@
+import { add, addMinutes, startOfDay } from "date-fns";
+import { DeviceDataV2Aggregate } from "@careevolution/mydatahelps-js";
+export type IntradayHeartRateData = { [key: string]: DeviceDataV2Aggregate };
```
greinard, line 3: This should be: `import { IntradayHeartRateData } from '../../../helpers/heart-rate-data-providers/combined-intraday-heart-rate-providers'`.

The type definition here is out of date (using a `string` key).

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,133 @@
...
+        const yMinValue = Math.min(...yDomain);
+
+        setOptions({ domainMin: yMinValue, domainMax: yMaxValue, thresholds: props.thresholds, connectNulls: false });
+        setData(chartData);
+        setLoading(false);
+    }
+
+    function transformToDataPoint(date: Date, data: IntradayHeartRateData) {
+        var dayKey = date.getTime();
+        var intradayDataPoint: IntradayDataPoint = { timestamp: dayKey, date: date };
```
greinard, line 99: Minor:  My IDE is warning me that all of these `var`'s should be `let`'s or `const`'s.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,133 @@
...
+        if (data[dayKey]) {
+            intradayDataPoint.value = (data[dayKey]).statistics[`${props.aggregationOption}`];
```
greinard, line 101: Minor:  No need for the parens around `data[dayKey]`.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,133 @@
...
+    lineColor?: ColorDefinition,
+    thresholds?: ChartThreshold[],
+    innerRef?: React.Ref<HTMLDivElement>
+}
+
+export interface IntradayDataPoint {
+    timestamp: number 
+    date: Date
+    value?: number
+    rawValue?: number
```
greinard, line 30: Is `rawValue` being used?

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
greinard: And no need for the string interpolation with the aggregation option.

`intradayDataPoint.value = data[dayKey].statistics[props.aggregationOption];`

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    function transformToChartData(data: IntradayHeartRateData) {
+        const start = intervalStart;
+        const end = intervalEnd;
+        const interval: Interval = { start, end };
+        const options: any = { step: 5 };
+        var expectedIntervals = eachMinuteOfInterval(interval, options);
+        var chartData: any[] = expectedIntervals.map((date) => transformToDataPoint(date, data));
```
greinard, line 82: Minor:  No need for parens around `date`.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
reneefromhold: @brandoncharnesky the domainMax is only needed for setting the gradient. The graph will detect the max value.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.stories.tsx
reneefromhold: @brandoncharnesky More detail on the aggregation options can be found [here](https://developer.mydatahelps.org/api/device_data_v2.html)

To your point and offhand, I don't have a good example of when to use max/min on this particular graph. I kept it as an enum for flexibility but can restrict to avg.  @chrisnowak Do you have any thoughts on supporting all the aggregate functions vs limiting to avg?

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.data.ts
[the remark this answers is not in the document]
greinard: Thanks for the clarification.  I had to do some maths, but I do agree with your assessment.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
greinard, line 82: The `transformToDataPoint(...)` function has an implicit return type of `IntradayDataPoint`, so the `chartData` can be declared as an `IntradayDataPoint[]` instead of `any[]`, or the type can be omitted so that it gets that type implicitly.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,133 @@
...
+    };
+
+    useInitializeView(initialize, [], [props.previewState, props.dataSources, dateRangeContext?.intervalStart]);
+
+    return <div ref={props.innerRef}>
+        {loading && <LoadingIndicator></LoadingIndicator>}
+        {!loading &&
+            <>
+                {data && <TimeSeriesChart intervalType='Day' intervalStart={intervalStart} data={data}
+                    chartHasData={(!loading && isNumber(options?.domainMax) && options?.domainMax > 0)}
```
greinard, line 127: I think this can simplify to: `chartHasData={!loading && !!options?.domainMax && options.domainMax > 0}`.

My IDE is complaining about `options?.domainMax` possibly being `undefined` at the moment.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    useInitializeView(initialize, [], [props.previewState, props.dataSources, dateRangeContext?.intervalStart]);
+
+    return <div ref={props.innerRef}>
+        {loading && <LoadingIndicator></LoadingIndicator>}
+        {!loading &&
+            <>
+                {data && <TimeSeriesChart intervalType='Day' intervalStart={intervalStart} data={data}
+                    chartHasData={(!loading && isNumber(options?.domainMax) && options?.domainMax > 0)} 
+                    chartType={"Line"} series={[{ dataKey: 'value', color: lineColor }]} tooltip={IntradayHrToolTip}
```
greinard, line 128: Minor:  `chartType="Line"` (no braces necessary).

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,133 @@
...
+                </div>
+            );
+        }
+        return null;
+    };
+
+    useInitializeView(initialize, [], [props.previewState, props.dataSources, dateRangeContext?.intervalStart]);
+
+    return <div ref={props.innerRef}>
+        {loading && <LoadingIndicator></LoadingIndicator>}
```
greinard, line 123: Minor:  `<LoadingIndicator />`

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,133 @@
...
+        {!loading &&
+            <>
+                {data && <TimeSeriesChart intervalType='Day' intervalStart={intervalStart} data={data}
+                    chartHasData={(!loading && isNumber(options?.domainMax) && options?.domainMax > 0)} 
+                    chartType={"Line"} series={[{ dataKey: 'value', color: lineColor }]} tooltip={IntradayHrToolTip}
+                    options={options}  ></TimeSeriesChart>}
```
greinard, line 129: Minor: You can just close out the empty `TimeSeriesChart` tag with `/>` instead of `</TimeSeriesChart>`.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
```diff
@@ -204,6 +229,32 @@ export const multipleLineChart : Story = {
     ]
 };
 
+const thresholds : ChartThreshold[] = [
+    { value: 80, referenceLineColor: "green", overThresholdColor: "green" },
+	{ value: 120, referenceLineColor: "orange", overThresholdColor: "#ffdd21" },
+	{ value: 180, referenceLineColor: "red", overThresholdColor: "#ff0000" }
```
greinard, line 235: Minor:  Formatting looks a tad off here.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -129,10 +159,36 @@ export default function TimeSeriesChart(props: TimeSeriesChartProps) {
...
                 }
+
+                if (isNumber(domainMin)) nDomainMin = domainMin;
+                let domainMax = (props.options as MultiSeriesLineChartOptions).domainMax ?? 'auto';
+                if (isNumber(domainMax)) nDomainMax = domainMax;
             }
         }
 
         return <>
+            <defs>
```
greinard, line 170: Instead of having these defs in the standard chart components, can you move them down into the line chart code similar to how it is done with bar charts?

Also, I'm not sure I understand the significance of the `lineKeyPrefix`.  Can't the line chart threshold gradients just always use the `gradientKey` like the bar chart ones do?

reneefromhold: @greinard Good feedback, I cleaned this up

src/helpers/chartOptions.ts
reneefromhold: I reverted this

src/helpers/chartOptions.ts
reneefromhold: @greinard I would prefer to hold off on the custom dots implementation in this PR to avoid further hold up.

chrisnowak:
Something odd i've noticed; looks like the x-axis is getting squished to the left/right when you change the date and then reverting back to it's original position - it should remain fixed:
squishaxis

src/helpers/chartOptions.ts
greinard: Could we at least name it `dot` to match the recharts impl?

At the moment we do a lot of attempted casting of the options before using them.

I think it would be nice if we could move toward just using `{...props.options}` on things like the `Line` component, but that will only work if the property names match.

src/helpers/heart-rate-data-providers/combined-avg-intraday-heart-rate-providers.ts
```diff
@@ -0,0 +1,54 @@
...
+        return { runningAverage: (existing.runningAverage + newData.statistics[aggregationOption]) / existing + 1, count: existing.count + 1 };
+    }
+
+    return Promise.all(providers).then((values) => {
+        let data: IntradayHeartRateData = {};
+        var runningAverage: { [key: number]: { runningAverage: number, count: number } } = {};
+        values.forEach((nameSpaceData) => {
+            nameSpaceData.forEach((intervalData) => {
+                var dayKey = (new Date(intervalData.date)).getTime();
+                if (data[dayKey]) {
```
chrisnowak, line 41: This running average stuff isn't necessary...I actually included this in the spec:  "It should query data for all data sources specified as props; if there happens to be data for 2 data sources for an interval it should use the data source that is first in the array (e.g. Fitbit by default)"

I think averaging a max between Fitbit / Apple Health seems wrong; but mostly it overcomplicates the code.  So basically if data[dayKey] exists you don't need to do anything

src/helpers/heart-rate-data-providers/combined-avg-intraday-heart-rate-providers.ts
```diff
@@ -0,0 +1,54 @@
...
+    const buildRunningAverage = function (existing: any, newData: DeviceDataV2Aggregate) {
+        return { runningAverage: (existing.runningAverage + newData.statistics[aggregationOption]) / existing + 1, count: existing.count + 1 };
+    }
+
+    return Promise.all(providers).then((values) => {
+        let data: IntradayHeartRateData = {};
+        var runningAverage: { [key: number]: { runningAverage: number, count: number } } = {};
+        values.forEach((nameSpaceData) => {
+            nameSpaceData.forEach((intervalData) => {
+                var dayKey = (new Date(intervalData.date)).getTime();
```
chrisnowak, line 40: It's no a "dayKey" anymore so can you rename instances of "dayKey" (I see them in various other places) to a better var like "intervalKey" or "timestamp" or something?

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -0,0 +1,125 @@
...
+        } else {
+            combinedAvgIntradayHeartRateDataProvider(
+                props.dataSources,
+                intervalStart,
+                intervalEnd,
+                props.aggregationIntervalMinutes
+            ).then((data: IntradayHeartRateData) => {
+                transformToChartData(data);
+            }).catch((error: any) => {
+                console.log("error");
```
chrisnowak, line 64: Would remove console.log

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -156,6 +183,37 @@ export default function TimeSeriesChart(props: TimeSeriesChartProps) {
         </>
     }
 
+    function createLineChartDefs() {
```
chrisnowak, line 186: Since this is a pretty big function and this chart is already pretty complex, it would be nice to split this into a separate helper file, including the createStopsFromThresholds function, so this file can remain smaller

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -117,8 +117,35 @@ export default function TimeSeriesChart(props: TimeSeriesChartProps) {
         }
     }
 
-    //ensures that gradients are unique for each chart
```
chrisnowak, line 120: Maybe keep this comment?

chrisnowak:
Broadly looks good; I think the big thing is probably the animation/transition between days is off somehow.

reneefromhold:
@chrisnowak The LoadingIndicator had been there it just wasn't working. I removed it altogether and did some refactoring so the data loads similar to the DailyDataChart /limited state. Things look good with my test ppt but please lmk how this looks against your.

chrisnowak:
Hmm, it still doesn't seem quite right; when I page through days it shows "No Data" in between.  

When i'm paging through days in the DailyDataChart and switch days, it shows the loading indicator over the chart while the data loads:
loading

While i'm paging through with IntradayHeartRateChart, it's showing "No Data" when I page through while it's still loading:
loading2
So if it's slow at loading data for instance, one would think that there's just no data

ajmenca:
Just a note - working with @reneefromhold yesterday, some of her fixes here resulted in seeing the X-Axis ticks bouncing around.  I determined that it wasn't anything she was doing in this PR and was a symptom of the TimeSeriesChart itself.  I have a fix for that out here: https://github.com/CareEvolution/MyDataHelpsUI/pull/277.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -85,35 +81,35 @@ export default function (props: IntradayHeartRateChartProps) {
...
-    let yMinValue = 0;
-
     if (data) {
-        iHrData = expectedIntervals.map(date => transformToDataPoint(date, data));
-
         var yDomain: number[] = [];
-        Object.values(iHrData).forEach(dataPoint => {
-            dataPoint.value !== undefined && yDomain.push(dataPoint.value);
-        });
+        var keys = Object.keys(data).map((key) => parseInt(key));
```
reneefromhold, line 86: I am going to move this to data layer

ajmenca (approved):
I don't have live data to access - but the Stories all look good to me, and I think the code has had more than enough review.  As I said in an earlier comment I think this just needs https://github.com/CareEvolution/MyDataHelpsUI/pull/277 merged alongside it to get rid of the last UI flickers.

reneefromhold:
@chrisnowak The LoadingIndicator is showing as expected. Thanks for catching that.

reneefromhold:
@chrisnowak @greinard I will plan to merge this later today unless there are objections.

PR #248, opened 2024-04-25
"Relative Activity" Related Components 

chrisnowak:
## Overview

This adds a handful of new components related to displaying your device / wearable activity relative to certain thresholds.     

### RelativeActivityWeekNavigator / RelativeActivityWeekCoordinator
RelativeActivityWeekNavigator uses the WeekCalendar and SparkBarCharts to show an at-a-glance view of your week's activity relative to thresholds:
Screenshot 2024-04-25 at 9 35 26 AM
It allows passing in a set of data types, as well as a threshold for each data type and a color to change to when you are over the threshold.

It might be nice in the future to disconnect the threshold from the color changes, because sleep is a bit of an outlier here - you might want to highlight sleep if it is under 5 hours or over 10 hours or something - probably not when it's over 8.  (My plan was just to have sleep not have an overThresholdColor).  

You can swipe left/right to change the current week and tap a day to select a day.

Following a similar pattern to DateRangeNavigator / DateRangeCoordinator, the RelativeActivityWeekCoordinator contains a RelativeActivityWeekNavigator and provides a DateRangeContext.  

Somewhat confusingly - and I would definitely be open to naming suggestions - RelativeActivityWeekCoordinator only provides a *single day* date range context, not a week - because you are actually selecting a *specific date* on that week for display of further information.  I could change both of these to just be RelativeActivityDateRangeNavigator and RelativeActivityDateRangeCoordinator, or RelativeActivityDayNavigator / RelativeActivityDayCoordinator...I think I like the latter and will probably change to that if no other suggestions come up. 

### RelativeActivity / RelativeActivityToday
A new RelativeActivity component which is a more flexible version of RelativeActivityToday that:
- Is not anchored to the current day
- Allows specifying specific thresholds for your activity (e.g. 120 bpm for HR) instead of always using 30 day average

It is meant to be used in a "Day" DateRangeContext, e.g.
Screenshot 2024-04-25 at 9 47 43 AM

(As you can imagine, it is really meant to be used with RelativeActivityWeekCoordinator)

### DateRangeTitle
This is a simple component for displaying the full title of a date range or date inside a DateRangeContext.  Mostly again meant to be used with the RelativeActivityWeekCoordinator
Screenshot 2024-04-25 at 9 49 28 AM

### CombinedSleepDataProvider
As part of this I also added a combined "SleepMinutes" data provider that queries across Fitbit / Garmin / Apple Health, for ease of use, in particular in scenarios where one has both Apple Health and Fitbit sleep data (probably fairly common for anyone with a Fitbit) it wouldn't make sense to see both represented on the RelativeActivityWeekCoordinator.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,184 @@
...
+                if (threshold == undefined) { return; }
+
+                let fillPercent = data[dayKey] / (threshold * 2);
+                if (fillPercent > 1) { fillPercent = 1; }
+                computedResults.push({
+                    dataType: dataType,
+                    fillPercent: fillPercent,
+                    value: dataType.formatter(data[dayKey])
+                });
+                setResults(computedResults);
```
greinard, line 134: Should this be outside the `forEach` loop?

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,184 @@
...
+}
+
+interface RelativeActivityResult {
+    dataType: RelativeActivityDataType;
+    fillPercent: number;
+    value: string;
+}
+
+export default function (props: RelativeActivityProps) {
+    let [result, setResults] = useState<RelativeActivityResult[] | null>(null);
```
greinard, line 40: Maybe rename to `results`?

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+    let dateRangeContext = useContext(DateRangeContext);
+    let date = props.date;
```
greinard, line 43: I think this can simplify to:
`let date = props.date ?? dateRangeContext?.intervalStart ?? startOfDay( new Date() );`

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,184 @@
...
+import language from "../../../helpers/language";
+import { useInitializeView } from "../../../helpers/Initialization";
+import { ColorDefinition } from "../../../helpers/colors";
+
+export interface RelativeActivityProps {
+    dataTypes: RelativeActivityDataType[];
+    previewState?: "Default";
+    title?: string;
+    innerRef?: React.Ref<HTMLDivElement>
+    specifyThresholds?: boolean;
```
greinard, line 19: This property seems to act more like `requireThresholds`.  The data type thresholds can be specified regardless of this setting, but when this is `true`, those without a threshold are hidden.

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,184 @@
...
+        {props.title && <CardTitle title={props.title} />}
+        {result.map(c => <div key={c.dataType.dailyDataType} className="mdhui-relative-activity-datatype">
+            <ActivityMeter className="mdhui-relative-activity-meter"
+                label={c.dataType.label}
+                value={c.value}
+                fillPercent={c.fillPercent}
+                averageFillPercent={0.5}
+                icon={c.dataType.icon}
+                color={getColor(c.dataType, c.fillPercent)} />
+            {c.dataType.threshold !== undefined &&
```
greinard, line 164: I think there exists a couple of edge cases we'll want to account for here (and maybe below).

The first is that this will display if `c.dataType.threshold` is specified, `props.specifyThresholds` is `false`, and there were at least 5 values for the data type.  In that case, the actual threshold in use will be the 30 day average threshold, not this specified one.

The second is similar, with `c.dataType.threshold` being specified, `props.specifyThresholds` set to `false`, and there were less than 5 values for the data type.  It will then be using this specified threshold as the fallback, making this label correct, yet we'll also have the label below saying the midline for all the activities are their 30 day averages.

src/components/container/RelativeActivityToday/RelativeActivityToday.tsx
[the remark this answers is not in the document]
chrisnowak: Yeah just backwards compatibility

src/components/container/RelativeActivityWeekNavigator/RelativeActivityWeekNavigator.tsx
```diff
@@ -0,0 +1,111 @@
...
+            let resultsMap: { [key: string]: DailyDataQueryResult } = { ...dailyData };
+            let currentDailyData = dailyData || {};
+            results.forEach((result, index) => {
+                let newResult = { ...currentDailyData[props.dataTypes[index].dailyDataType], ...result };
+                resultsMap[props.dataTypes[index].dailyDataType] = newResult;
+            });
+            setDailyData(resultsMap);
+        });
+
+        if (props.previewState === "default") {
```
greinard, line 60: Is this code block necessary?  Seems like the daily data is already going to get set appropriately when in the preview state.

chrisnowak:
@greinard thanks for the initial look!  Maybe hold off on reviewing more for now...as I work on this in view builder I'm realizing I want to adjust/rethink a few things so converting to draft for now

src/components/presentational/WeekCalendar/WeekCalendar.tsx
```diff
@@ -12,22 +13,23 @@ export interface WeekCalendarProps {
...
+	innerRef?: React.Ref<HTMLDivElement>;
 }
 
 export default function (props: WeekCalendarProps) {
 	const element = useRef<HTMLDivElement>(null);
 
 	useEffect(() => {
 		if (props.onStartDateChange) {
-			var scrollListener = function (ev: Event) {
+			var scrollListener = debounce(function (ev: Event) {
```
chrisnowak, line 24: This debounce changed fixed a longstanding weird issue with this where it never really worked right on desktop due to weird orders of events; since essentially when it scrolls left to 0 we want it to immediate re-render and snap back to the middle scroll position...anyways, it works now

src/components/container/RelativeActivity/RelativeActivity.tsx
chrisnowak: I refactored this property away so it just does stuff automatically based on whether thresholds are set or not

src/components/container/RelativeActivityWeekNavigator/RelativeActivityWeekNavigator.tsx
chrisnowak: Removed as part of refactoring

greinard:
Great, I will take another look this afternoon.

src/helpers/query-daily-data.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+export async function queryPreviewDailyData(type: string, startDate: Date, endDate: Date) {
+	var result: DailyDataQueryResult = {};
+	let range = getDailyDataTypeDefinition(type as DailyDataType).previewDataRange;
+
+	//poor man's seeded javascript rng
+	//just for generating a relatively large random integer from a string seed,
+	//which then uses modulo to get a number in a range
+	async function randomNumber(message: string) {
```
greinard, line 66: Why go through all this trouble for a preview state random number?  Why not just use `Math.random()`?

src/helpers/query-daily-data.tsx
```diff
@@ -0,0 +1,94 @@
...
+	allTypeDefinitions.map((typeDefinition) => [typeDefinition.type, typeDefinition])
+);;
+
+export function registerDailyDataProvider(type: string, provider: DailyDataProvider, availabilityCheck: DailyDataAvailabilityCheck) {
+	dailyDataTypes.set(type, {
+		type: type,
+		getLabel: () => type,
+		dataProvider: provider,
+		availabilityCheck: availabilityCheck,
+		icon: <FontAwesomeSvgIcon icon={faPersonRunning} />,
```
greinard, line 24: Do we want this to be hard coded to the same icon for all data providers?

src/helpers/query-daily-data.tsx
```diff
@@ -0,0 +1,94 @@
...
+		formatter: defaultFormatter,
+		previewDataRange: [0, 10000]
```
greinard, line 26: Same question here.  Should the icon and this data range be optional params with these defaults?

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,111 @@
...
+    innerRef?: React.Ref<HTMLDivElement>
+    date?: Date;
+}
+
+export default function (props: RelativeActivityProps) {
+    let [results, setResults] = useState<{ [key: string]: RelativeActivityQueryResult } | undefined>(undefined);
+    let relativeActivityContext = useContext(RelativeActivityContext);
+
+    let dataTypes: RelativeActivityDataType[] = props.dataTypes || [];
+    if (props.useContext && relativeActivityContext?.dataTypes) {
```
greinard, line 30: Under what circumstances would you wrap this component with a relative activity context, but *not* want to use it?

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,111 @@
...
+            let computedResult: { [key: string]: RelativeActivityQueryResult } = {};
+            dataTypes.forEach(dataType => {
+                computedResult[dataType.dailyDataType] = results[dataType.dailyDataType][getDayKey(date!)];
+            });
+            return computedResult;
+        }
+
+        setResults(undefined);
+        if (props.useContext) {
+            setResults(transformResults(relativeActivityContext!.data));
```
greinard, line 49: I don't see anywhere that `relativeActivityContext.data` is being set.  Is this still needed?

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    let dateRangeContext = useContext(DateRangeContext);
+    let date = props.date ?? dateRangeContext?.intervalStart ?? startOfDay(new Date());
+
+    function loadData() {
+        function transformResults(results: { [key: string]: { [key: string]: RelativeActivityQueryResult } } | undefined) {
+            if (!results) return;
+            let computedResult: { [key: string]: RelativeActivityQueryResult } = {};
+            dataTypes.forEach(dataType => {
+                computedResult[dataType.dailyDataType] = results[dataType.dailyDataType][getDayKey(date!)];
```
greinard, line 42: Minor, but when I saw this named `computedResult`, I was looking somewhere for a computation.  Maybe something like `transformedResults` or `resultsForDate` would fit a little better.

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,111 @@
...
+            return;
+        }
+        queryRelativeActivity(date!, date!, dataTypes, !!props.previewState).then(results => {
+            setResults(transformResults(results));
+        });
+    }
+
+    useInitializeView(() => {
+        loadData();
+    }, ["externalAccountSyncComplete"], [dateRangeContext, relativeActivityContext, props.date, props.dataTypes], 0);
```
greinard, line 59: I'm curious about why you'd want to disable the debounce here by passing `0`.  The debounce in `useInitializeView` is meant to prevent double loading when an MDH tab first renders and a couple of events fire almost simultaneously.  It seems unlikely that a second event firing within 500ms of the first would result in a query that returns additional data.

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    if (results == null || !Object.keys(results).length) {
```
greinard, line 61: Minor: I think this may work because of the nature of `==`, but `result` is only ever `undefined` or an object, so it may be more appropriate to just use `!result` here.

src/helpers/daily-data-types/formatters.ts
```diff
@@ -0,0 +1,15 @@
...
+export const heartRateFormatter = (value: number) => `${Number(value.toFixed(0))} ${language("bpm")}`;
+export const minutesFormatter = (value: number) => {
+    var hours = Math.floor(value / 60);
+    var displayValue = hours > 0 ? (`${hours}${language("hours-abbreviation")} `) : "";
+    if (Math.round(value - (hours * 60)) !== 0) {
+        displayValue = `${displayValue}${(Math.round(value - (hours * 60)))}${language("minutes-abbreviation")}`;
+    }
+    return displayValue;
+};
+export const sleepYAxisConverter = (value: number) => value / 60.0;
```
greinard, line 15: Should this be named `sleepYAxisFormatter`?

src/components/container/DeviceDataMonthChart/DeviceDataMonthChart.tsx
```diff
@@ -96,16 +88,16 @@ export default function (props: DeviceDataMonthChartProps) {
...
-		if (props.previewState) { return; }
 
 		checkForAnyData();
 
 		currentInitialization.current = (currentInitialization.current ?? 0) + 1;
 		var initialization = currentInitialization.current ?? 0;
 		setLoading(true);
 		var loadData = function () {
-			var dataRequests = props.lines.map(l => queryDailyData(l.dailyDataType, monthStart, monthEnd));
+			var dataRequests = props.lines.map(l => queryDailyData(l.dailyDataType, monthStart, monthEnd, props.previewState !== undefined));
```
greinard, line 100: Maybe `!!props.previewState`.

src/components/container/DailyDataChart/DailyDataChart.tsx
```diff
@@ -146,10 +152,14 @@ export default function DailyDataChart(props: DailyDataChartProps) {
     const GraphToolTip = ({ active, payload, label }: any) => {
         if (active && payload && payload.length) {
             var date = payload[0].payload.date;
+            let formatter = props.valueFormatter;
```
greinard, line 155: Minor, but can simplify to: 

`let formatter = props.valueFormatter ?? getDailyDataTypeDefinition(props.dailyDataType).formatter;`

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,111 @@
...
+            let dataTypeResult = results![d.dailyDataType];
+            if (!dataTypeResult) {
+                return null;
+            }
+            let dataTypeDefinition = getDailyDataTypeDefinition(d.dailyDataType);
+
+            return <div key={d.dailyDataType} className="mdhui-relative-activity-datatype">
+                <ActivityMeter className="mdhui-relative-activity-meter"
+                    label={dataTypeDefinition.getLabel()}
+                    value={dataTypeDefinition.formatter(dataTypeResult.value) || dataTypeResult.value.toString()}
```
greinard, line 93: Do we really want this defaulting?  It may be unusual, but if the formatter returns something like empty string for a given value, maybe we should respect that rather than displaying the non-formatted value?

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    function getBarColor(dataType: RelativeActivityDataType, fillPercent: number) {
+        if (dataType.threshold !== undefined && dataType.overThresholdColor && fillPercent > .5) {
+            return dataType.overThresholdColor;
+        }
+        return dataType.color || "var(--mdhui-color-primary)";
+    }
+
+    function getThresholdLabel(dataType: RelativeActivityDataType, threshold: number) {
+        if (!dataTypes.some(d => d.threshold !== undefined)) {
```
greinard, line 73: Rather than the double negative here, I like the following that you used further below:

```dataTypes.every(d => d.threshold === undefined)```

I think those both evaluate to the same thing.

File: src/helpers/daily-data-types/garmin.tsx [hunk 21]
```diff
@@ -0,0 +1,229 @@
...
+import language from "../language";
+import React from "react";
+import { defaultFormatter, heartRateFormatter, minutesFormatter, sleepYAxisConverter } from "./formatters";
+import { simpleAvailabilityCheck } from "./availability-check";
+
+let garminTypeDefinitions: DailyDataTypeDefinition[] = [
+    {
+        type: DailyDataType.GarminSteps,
+        dataProvider: garminStepsDataProvider,
+        availabilityCheck: simpleAvailabilityCheck("Garmin", ["Daily"]),
```
greinard, line 14: Minor:  The second arg to all of these availability checkers can just be the single string rather than an array.

src/helpers/date-helpers.ts
```diff
@@ -45,3 +45,18 @@ export function getMonthName(month: number) {
 	}
 	return capitalizeFirstLetter(format(new Date(new Date().getFullYear(), month, 1, 0, 0, 0, 0), "MMMM", { locale: locale }));
 }
+
+export function titleForDateRange(intervalType: "Day" | "Week" | "Month", intervalStart: Date, variant?: "short" | "long") {
+	if (intervalType == "Month" && intervalStart.getDate() == 1) {
+		return `${getMonthName(intervalStart.getMonth())} ${intervalStart.getFullYear()}`;
+	}
+	else if (intervalType == "Week" || (intervalType == "Month" && intervalStart.getDate() != 1)) {
```
greinard, line 53: The `&& intervalStart.getDate() != 1` is not needed here.  It will always be `true` if `intervalType == "Month"`.

src/helpers/date-helpers.ts
```diff
@@ -45,3 +45,18 @@ export function getMonthName(month: number) {
...
+		return `${format(intervalStart, "MM/dd/yyyy")} - ${format(add(intervalStart, { days: -11 }), "MM/dd/yyyy")}`;
```
greinard, line 54: Where does the hard coded `-11` days come from?

src/helpers/date-helpers.ts
```diff
@@ -1,5 +1,5 @@
 import MyDataHelps from "@careevolution/mydatahelps-js";
-import { add, format, isSameDay } from "date-fns";
+import { add, format, isSameDay, isToday } from "date-fns";
```
greinard, line 2: `isToday` is not used.

src/components/container/RelativeActivity/RelativeActivity.tsx
chrisnowak: Ah yep, good call, this is from before when I was making formatters optional (which would have been wrong then too)

chrisnowak: Oh wait - nm, formatters are still optional if you call the legacy registerDailyDataProvider - so yeah I just need to fix this to default to value if there is no formatter

src/helpers/daily-data-types/formatters.ts
chrisnowak: I called these "converters" rather than "formatters" because they produce a number; right now they are just kind of used for sleep graphs because we want the y-axis to represent hours instead of minutes

src/components/container/RelativeActivity/RelativeActivity.tsx
chrisnowak: Fixed

src/components/container/RelativeActivity/RelativeActivity.tsx
chrisnowak: So yes, this actually is an issue I wanted to surface, because I think it's broader than this and may actually have effects on other places we use this.  You can see it if you actually set this back to 500 and check out the RelativeActivityDayCoordinator > WithRelativeActivity story.

Because we are listening for context changes here, when you click on each day, it actually takes 500ms to update the selected day which definitely is not a good UX - the actual selected day should update immediately.  Ultimately, context changes can result from user interaction so should happen immediately rather than having a debounce.

So yeah, i'm actually not sure what the best workaround is here.  I considered a couple things:
- We could make it a "leading" debounce, but of course if someone clicks a second day within 500ms (possible) that means it won't fire at all.  
- We could make the debounce only happen for the MyDataHelps events, e.g. instead of the useEffect calling debouncedInitialize() we call initialize() - I don't think this prevents a double render on first load if the app also fires applicationDidBecomeVisible though

Maybe other answers like disabling the debounce after the first debounce if this is just to prevent issues with initial load or something?

src/components/container/RelativeActivity/RelativeActivity.tsx
chrisnowak: It's set in RelativeActivityCoordinator - the "With Relative Activity" story should demonstrate this - basically the "Relative Activity" component does not need to fetch data if it's using context

src/components/container/RelativeActivity/RelativeActivity.tsx
chrisnowak: I think if you wanted to expand/change the data types that are shown on a detail view.  Like the RelativeActivityCoordinator could have "sleep" as one of the bars, when you click a day it shows a RelativeActivity component with "Sleep" but then a second one with REM, Deep, Light sleep etc.

src/helpers/query-daily-data.tsx
chrisnowak: Math.random() can't be seeded so can't be made deterministic.   See PR notes for why I wanted a deterministic solution.

Couple reasons:
- Randomly generating different data on each render means the graphs get very "jumpy" under certain circumstances, like in view builder in particular
- It can make it hard to tell a consistent story if you're building a flowboard or something; like lets say we have a graph of steps that shows 5,000 for 5/1, then below that a component that shows 3,000 for 5/1 because they were both generating their own random data
  - Originally I wasn't planning on passing the data from RelativeActivityCoordinator to the RelativeActivity component, so this was very noticeable there since the bars at the top for the selected day wouldn't match the bars for the day...

But yeah, this solution basically means "Steps" for 5/1 will always be the same number regardless of where it's showing

src/helpers/query-daily-data.tsx
chrisnowak: Ah I added a comment above here; so this method should essentially be deprecated in favor of "registerDailyDataTypeDefinition" which lets you specify all those things by sending in a DailyDataTypeDefinition itself.

I just kept this method for backwards compatibility; I could add additional arguments but I think the answer is more to just use the new method if you need to specify this stuff.

Ultimately, if you're currently using the registerDailyDataProvider method and upgrade MyDataHelpsUI, this stuff shouldn't really affect you unless you use one of the new components anyways

src/components/container/RelativeActivity/RelativeActivity.tsx
chrisnowak: Fixed

src/helpers/daily-data-types/formatters.ts
greinard: Sounds reasonable.  I just asked because of the file it lives in.  🙂

src/components/container/DailyDataChart/DailyDataChart.tsx
chrisnowak: Fixed

src/helpers/date-helpers.ts
chrisnowak: It will not always be true, you can have a month interval that starts on an arbitrary date

chrisnowak: (hmm this makes me realize that i'll need to adjust the date formats for international use cases at some point, since we're hardcoding ```"MM/dd/yyyy"``` here)

src/helpers/date-helpers.ts
chrisnowak: ...that's weird, must be a typo or something I messed up in moving this code around, definitely wrong

chrisnowak: Great catch on this, fixed...should have been subtracting one day from the interval end, somehow that got transformed into adding -11 days to interval start

src/components/container/RelativeActivity/RelativeActivity.tsx
chrisnowak: Fixed

src/helpers/query-daily-data.tsx
greinard: Makes sense.  Thanks for the explanation.

src/helpers/query-daily-data.tsx
greinard: Sounds good.

src/helpers/date-helpers.ts
greinard: It will always be `true` because if it were ever `false`, the first `if` condition would have already matched.

src/components/container/RelativeActivity/RelativeActivity.tsx
chrisnowak: What *might* be the answer is this:

- Determine whether this is the "initial mount" of the component via https://stackoverflow.com/a/55075818
- On the initial mount of the component, call debouncedInitialize in the useEffect
- On subsequent useEffect calls, just call initialize
- MyDataHelps.on stuff always uses debouncedInitialize

I'll see if that works

src/helpers/date-helpers.ts
greinard: What I mean is, if `intervalType == "Month"` and `intervalStart.getDate() != 1` is `false`, then `intervalStart.getDate() == 1` will be `true`, so we won't even get to this point in the code, because the first `if` condition was met.

src/components/container/RelativeActivity/RelativeActivity.tsx
chrisnowak: Seems like it does work - see adjustments to useInitializeView - wondering if you have any thoughts on that though since i'm not as familiar with the scenarios for why we added the debounce

src/helpers/date-helpers.ts
chrisnowak: Ohhh I see what you mean - though you were saying this code path would never be reached.  You are saying it's redundant - yep, makes sense

chrisnowak: Fixed

src/components/container/RelativeActivity/RelativeActivity.tsx
greinard: Yeah, I like that and I think it should work fine.

The initial load and MDH events would all be debounced, and other subsequent re-executions of `useEffect` should only be caused by dependency changes, I think, which should be okay to just use initialize immediately.

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    function getBarColor(dataType: RelativeActivityDataType, fillPercent: number) {
+        if (dataType.threshold !== undefined && dataType.overThresholdColor && fillPercent > .5) {
+            return dataType.overThresholdColor;
+        }
+        return dataType.color || "var(--mdhui-color-primary)";
+    }
+
+    function getThresholdLabel(dataType: RelativeActivityDataType, threshold: number) {
+        if (dataTypes.every(d => d.threshold !== undefined)) {
```
greinard, line 73: I think this should be `dataTypes.every(d => d.threshold === undefined)`, correct?

chrisnowak: Ah good catch, yes

src/components/presentational/DateRangeTitle/DateRangeTitle.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+    component: DateRangeTitle,
+    parameters: {
+        layout: 'fullscreen',
+    }
+};
+
+export const Week = {
+    args: {
+    },
+    render: (args: DateRangeTitleProps) => <Layout colorScheme="auto">
```
greinard, line 17: Any reason not to pull out a common render method for these that just accepts an interval type.  That seems to be the pattern we typically follow.

chrisnowak: Not really, fixed

chrisnowak:
One thing I realized since I had a window open for this all weekend - when the *current date* changes, we actually need to re-render the day navigator so it re-anchors to the current day being the right most day
Screenshot 2024-05-06 at 9 07 19 AM

src/components/container/RelativeActivityDayNavigator/RelativeActivityDayNavigator.tsx
```diff
@@ -0,0 +1,77 @@
...
+    let dayRenderer = function (year: number, month: number, day: number, selectedWeek: boolean) {
+        if (!props.dataTypes.length) { return null; }
+
+        var date = new Date(year, month, day);
+        var dayKey = getDayKey(date);
+        let bars: SparkBarChartBar[] = props.dataTypes.map(dataType => {
+            if (!dailyData || !dailyData[dataType.dailyDataType] || !dailyData[dataType.dailyDataType][dayKey]) {
+                return { color: "var(--mdhui-color-primary)", barFillPercent: 0 };
+            }
+            let value = dailyData[dataType.dailyDataType]?.[dayKey]?.value || 0;
```
greinard, line 43: Shouldn't need the `?`'s here.  The condition above ensures these will have defined values.

src/components/container/RelativeActivityDayNavigator/RelativeActivityDayNavigator.tsx
```diff
@@ -0,0 +1,77 @@
...
+            if (dataType.threshold != "30DayAverage" &&
+                dataType.threshold !== undefined &&
+                value > dataType.threshold &&
+                dataType.overThresholdColor) {
+                color = dataType.overThresholdColor;
+            }
+
+            return {
+                color: color,
+                barFillPercent: dailyData[dataType.dailyDataType]?.[dayKey]?.relativePercent || 0
```
greinard, line 54: Shouldn't need the `?`'s here.  The condition above ensures these will have defined values.

src/components/container/RelativeActivityDayNavigator/RelativeActivityDayNavigator.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+        return <div style={{ paddingTop: "8px" }}>
+            <SparkBarChart
+                averageFillPercent={0.5}
+                bars={bars} /></div>
+    }
+
+    let weekStartChanged = function (weekStart: Date) {
+        setWeekStart(weekStart);
+        props.onDateSelected(addDays(weekStart, 6));
```
greinard, line 66: Nothing wrong with this, but it is inconsistent with the `add( new Date(), { days: ... } )` calls above.  Maybe use one or the other for all the things, at least within a file.  I don't feel strongly about it, though.

src/helpers/date-helpers.ts
```diff
@@ -45,3 +45,21 @@ export function getMonthName(month: number) {
...
+
+	if (intervalType == "Month" && intervalStart.getDate() == 1) {
+		return `${getMonthName(intervalStart.getMonth())} ${intervalStart.getFullYear()}`;
+	}
+	else if (intervalType == "Week" || intervalType == "Month") {
+		return `${format(intervalStart, "MM/dd/yyyy")} - ${format(sub(intervalEnd, { days: 1 }), "MM/dd/yyyy")}`;
+	}
+	else if (intervalType == "Day") {
+		if (variant === "long") {
+			return `${getDayOfWeek(intervalStart)}, ${getFullDateString(intervalStart)}`;
```
greinard, line 61: I know it is not part of the PR here, but any thoughts on using `do` in `getFullDateString` instead of `d`, so that we get 1st, 2nd, 3rd, etc. instead of 1, 2, 3, etc.

src/helpers/daily-data-providers/combined-sleep.ts
```diff
@@ -0,0 +1,42 @@
...
+            return {};
+        }
+
+        return Promise.all(providers).then((values) => {
+            var data: { [key: string]: number } = {};
+            while (startDate < endDate) {
+                var dayKey = getDayKey(startDate);
+                var sleep: number | null = null;
+                values.forEach((value) => {
+                    if (value[dayKey] || sleep == null || sleep < value[dayKey]) {
```
greinard, line 30: Should this be: `value[dayKey] && (sleep == null || sleep < value[dayKey])`?

src/components/container/RelativeActivity/RelativeActivity.tsx
```diff
@@ -0,0 +1,117 @@
...
+            if (!dataTypeResult) {
+                return null;
+            }
+            let dataTypeDefinition = getDailyDataTypeDefinition(d.dailyDataType);
+            let formatter = d.formatter || dataTypeDefinition.formatter;
+
+            return <div key={d.dailyDataType} className="mdhui-relative-activity-datatype">
+                <ActivityMeter className="mdhui-relative-activity-meter"
+                    label={d.label || (dataTypeDefinition.labelKey ? language(dataTypeDefinition.labelKey) : d.dailyDataType)}
+                    value={formatter ? formatter(dataTypeResult.value) : dataTypeResult.value.toString()}
```
greinard, line 99: Just wanted to circle back on this since your earlier comment mentioned something about `formatter` being optional depending on how the data type was registered.  Re-reviewing that code, it looks to me like `formatter` should always be set now (unless someone intentionally works around the typescript error checking), correct?

src/helpers/daily-data-types.tsx
```diff
@@ -0,0 +1,81 @@
...
+	GoogleFitSteps = "GoogleFitSteps",
+	Steps = "Steps",
+	RestingHeartRate = "RestingHeartRate",
+	SleepMinutes = "SleepMinutes",
+	HomeAirQuality = "HomeAirQuality",
+	WorkAirQuality = "WorkAirQuality"
+};
+
+export interface DailyDataTypeDefinition {
+	dataSource?: "Unified" | "AppleHealth" | "Garmin" | "Fitbit" | "GoogleFit" | "AirQuality";
```
greinard, line 72: Sorry if I missed this in earlier PRs/discussions, but what is the "Unified" data source?

chrisnowak: Ah yeah I just updated this based on my work on the view builder PR.

I was trying to come up with a user-facing name for our "combined data sources".  Basically these query across Apple Health, Fitbit, Garmin to give you a single "Steps", "Resting Heart Rate", or "Sleep" value (these are the only "unified" data types at the moment).  Suspect we will add more in the future

"Unified" seems like a nice name for that I guess rather than "Combined" or "Deduplicated"

greinard: Got it.  Make sense.

src/helpers/relative-activity.tsx
```diff
@@ -0,0 +1,75 @@
...
+export interface RelativeActivityQueryResult {
+    relativePercent: number;
+    value: number;
+    threshold: number;
+}
+
+export interface RelativeActivityDataType {
+    dailyDataType: string;
+    color?: ColorDefinition;
+    overThresholdColor?: ColorDefinition;
```
greinard, line 15: It would probably be nice to support "under" thresholds as a first class option as well.  SpO2 and Steps come to mind as data types for which a low value is more likely to be highlighted.

I suppose you could achieve that by making the base color the highlight color and the `overThresholdColor` the normal color.

src/helpers/relative-activity.tsx
```diff
@@ -0,0 +1,75 @@
...
+            add(startDate, { days: dataType.threshold == undefined ? -31 : -1 }),
+            add(endDate, { days: 1 }),
+            preview));
+    return Promise.all(promises).then((results) => {
+        dataTypes.forEach((dataType, index) => {
+            let dataTypeData = results[index];
+
+            relativeActivityResults[dataType.dailyDataType] = {};
+
+            let currentDate = startDate;
```
greinard, line 36: I don't recall exactly where I saw it, but I just wanted to point out that while functionally the same, I do like how you converted the start date to `currentDate` here.  I think there are a couple other loops elsewhere that are just updating the `startDate` var, which doesn't break the functionality, but just feels a tad off because it is an incoming parameter.  🙂

src/helpers/relative-activity.tsx
```diff
@@ -0,0 +1,75 @@
...
+                let dayKey = getDayKey(currentDate);
+                let value = dataTypeData?.[dayKey] ?? 0;
+                let threshold = (dataType.threshold === "30DayAverage" || dataType.threshold === undefined) ? calculatePrevious30DayAverage(dataTypeData, currentDate) : dataType.threshold;
+                if (threshold !== undefined) {
+                    let fillPercent = value / (threshold * 2);
+                    if (fillPercent > 1) {
+                        fillPercent = 1;
+                    }
+
+                    relativeActivityResults[dataType.dailyDataType][dayKey] = {
```
greinard, line 47: Just want to verify my understanding here.  It looks like we will not return any results for data types for which a threshold cannot be established.  Is that correct?

src/helpers/relative-activity.tsx
chrisnowak: Exactly, that's just swapping the colors essentially, like with sleep you could highlight under 7.5 hours or something

src/helpers/strings-es.ts
```diff
@@ -317,7 +317,50 @@
     "device-data-month-chart-no-data": "Sin Datos",
     "device-data-month-chart-daily-average": "Promedio Diario",
     "device-data-month-chart-minutes": "Minutos",
-    "device-data-month-chart-sleep": "Dormir"
```
greinard, line 320: Any idea why the line count is off here?  Are we missing a few Spanish strings higher up in the file?  The rest of the language files all seem to be in sync on the line numbers.

src/helpers/relative-activity.tsx
chrisnowak: Correct

greinard (approved):
Looks great and adds some cool new functionality.

The only remaining potential bug I commented on is in the combined sleep file.

Aside from that, feel free to :ship: when ready.

src/helpers/query-daily-data.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+let dailyDataTypes = new Map(
+	allTypeDefinitions.map((typeDefinition) => [typeDefinition.type, typeDefinition])
+);
+
+// deprecated.  Instead use registerDailyDataTypeDefinition
+export function registerDailyDataProvider(type: string, provider: DailyDataProvider, availabilityCheck: DailyDataAvailabilityCheck) {
+	dailyDataTypes.set(type, {
+		type: type,
+		getLabel: () => type,
```
greinard, line 22: `getLabel` doesn't exist on `DailyDataTypeDefinition`.

src/helpers/daily-data-providers/combined-sleep.ts
chrisnowak: huh...yeah, that would certainly make more sense.  That's interesting, it's the same across steps / RHR combined providers as well, so likely we have been displaying lower values than we should be for those combined providers (steps would be the major one).

chrisnowak: Well, just for steps I guess, RHR uses average

src/components/container/RelativeActivity/RelativeActivity.tsx
chrisnowak: Good call, fixed

src/helpers/date-helpers.ts
chrisnowak: Makes sense to me

src/components/container/RelativeActivityDayNavigator/RelativeActivityDayNavigator.tsx
chrisnowak: Fixed

src/helpers/strings-es.ts
chrisnowak: It's consistent now after I merged in main, probably a missing string I copied in from main or something

PR #249, opened 2024-04-25
Survey Answer Data Chart

ajmenca:
## Overview

This PR adds a `SurveyDataChart` Container component.  This chart will graph responses to a Survey over time (configured by specifying the Survey Name, Step Identifier, and optionally the Result Identifier).  The presentation here was driven by a component I originally created for a customer where they have multiple surveys that calculate multiple scores per survey, and they want to see those scores side by side over time.  As a specific example, the project was giving repeated instances of the Five Factor Wellness Inventory; this survey generates a total score and five second order scores, and the aim was to show all of those scores over time.  The storybook entry I created is a simplified version of this:

image
image

The bigger swing I took here was to attempt to leverage the rendering code already established in the `DailyDataChart` by refactoring it into a more generic `DataChart` component that is used by *both* the `DailyDataChart` and the `SurveyDataChart`.  This is a pattern that I think we should really be following more often; separating the responsibilities of retrieving data and presenting data.  I've run into several cases in using the MDHUI library where I couldn't re-use one of the components because I needed to make the slightest tweak to the set of data being presented and couldn't because we're not separating these concerns (i.e. the `SurveyTaskList` gets a specific set of Tasks, rather than being a presentational component which renders a list of Tasks it's given, and we also provide an `OpenSurveyTaskList` which gets the data).  So I took the plunge in this PR to try and start introducing and encouraging this pattern.

Behaviorally, the only change I ended up making to the rendering code was the addition of being able to display multiple sets of data on one graph (something we'd had in the `DeviceDataMonthChart` but then did not support in the `DailyDataChart`).  I did this by having an optional `dataKeys` parameter for the `DataChart`.  If you don't pass any, it assumes a basic default key of `value` which is how the `DailyDataChart` formats its basic data.  So the Survey Chart can support lines, as shown above, as well as area and bar presentations (though you can get some truly awful color combinations of areas):
image
image

I also made it so that the "parent" container is the one who specifies how the hover detail works.  This was mostly for handling rather different formatting if there's only one line versus multiple.

Backwards compatibility for the `DailyDataChart` should be unchanged; it's external interface was unmodified.  All stories for it should still work, as verification.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

## Summary by CodeRabbit

- **New Features**
  - Introduced new chart types (Line, Bar, Area) for visualizing survey and daily data.
  - Added support for a new "6Month" interval type in date range components.

- **Enhancements**
  - Improved data handling and chart rendering with the new `TimeSeriesChart` component.
  - Updated chart options with additional properties for customization (colors, thresholds).

- **Bug Fixes**
  - Enhanced title generation for date ranges to include the new "6Month" interval.

- **Refactor**
  - Streamlined data processing and interval handling in chart components.
  - Replaced custom random number generator with a predictable one for consistent data previews.


basyl-durnan:
Very cool!

- Looks like you can graph multiple survey results on the same graph. Can those be results from the same and different surveys?
- I noticed you said "survey step". While I imagine it will almost always be a survey step's result, would they be able to handle form item results as well?
- Is there any validation to making sure we get numeric values? I of course hope study teams put validation on the result, but... you never know. How would it handle a non-numeric value?
- What happens if a survey is completed multiple times in one day? I'm not sure I have an answer for what to do... some people may want to visually see all of them, others may want an average. 
- Forgive me for not trusting myself to parse the code behind this, but I assume there is some check in case surveys share similar field names, right? We probably specify the survey?
- You can click on the bars to also see a legend, right?

basyl-durnan:
Adding @Kbalon in case she has some thoughts and/or feedback.

ajmenca:
> * Looks like you can graph multiple survey results on the same graph. Can those be results from the same and different surveys?

Yes, each line is configured individually and the data is queried for it individually, so you can combine things.

> * I noticed you said "survey step". While I imagine it will almost always be a survey step's result, would they be able to handle form item results as well?

Yes, the configuration actually takes both a step identifier and an (optional) result identifier.

> * Is there any validation to making sure we get numeric values? I of course hope study teams put validation on the result, but... you never know. How would it handle a non-numeric value?

Not at the moment.  A non numerical value will cause a failure in the graph; I can update it to just throw those values away if they come up though (which would really be the only option to proceed and graph what is valid).

> * What happens if a survey is completed multiple times in one day? I'm not sure I have an answer for what to do... some people may want to visually see all of them, others may want an average.

I can verify with a test, but I think the way I'm parsing the data, the later answer would be the one that would show up.  I agree there's not exactly a predictable expectation for this sort of case.  Seeing all the points wouldn't really be reasonable for this type of graph though; i.e. when you're making a line, which point do you run the line through?

> * Forgive me for not trusting myself to parse the code behind this, but I assume there is some check in case surveys share similar field names, right? We probably specify the survey?

Yup, survey name, step identifier, result identifier.  I'll update the PR description for clarity.

> * You can click on the bars to also see a legend, right?

Hover, but yes - same behavior.

basyl-durnan:
I think an error is probably best for non-numerical values. If someone is trying to graph results, we probably want to make it obvious there is an error in the survey values. The study team should enforce validation or a picklist.

Great point about showing multiple results on the same day. Let's revisit averages and such if that comes up.

Thanks!

chrisnowak:
I would definitely just throw the values away if it's non-numeric rather than have the graph error; makes it robust if you ever change anything in the survey definition etc...one string could crash the whole graph etc.

src/components/container/SurveyDataChart/SurveyDataChart.tsx
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        var dataRequests = props.lines.map(l => MyDataHelps.querySurveyAnswers({
+            surveyName: l.surveyName, 
+            stepIdentifier: l.stepIdentifier, 
+            resultIdentifier: l.resultIdentifier,
+            after: intervalStart.toISOString(),
+            before: intervalEnd.toISOString()
+        }));
+        Promise.all(dataRequests).then(function (data) {
```
reneefromhold, line 73: Consider using https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/helpers/query-all-survey-answers.ts

src/components/container/SurveyDataChart/SurveyDataChart.tsx
```diff
@@ -0,0 +1,176 @@
...
+    function processPages(pages: SurveyAnswersPage[]) {
+        var newDailyData: { [key: string]: SurveyAnswersPage } = {};
+        for (var i = 0; i < props.lines.length; i++) {
+            newDailyData[getDataKey(props.lines[i])] = pages[i];
+        }
+
+        return newDailyData;
+    }
+
+    useEffect(() => {
```
reneefromhold, line 91: Consider using the helper function useInitializeView in place of the useEffect

src/components/container/SurveyDataChart/SurveyDataChart.stories.tsx
```diff
@@ -0,0 +1,545 @@
...
+import { Card, Layout } from "../../presentational";
+import SurveyDataChart, { SurveyDataChartProps } from "./SurveyDataChart";
+import { SurveyAnswersPage } from "@careevolution/mydatahelps-js";
+
+export default { title: "Container/SurveyDataChart", component: SurveyDataChart, parameters: { layout: 'fullscreen' } };
+let render = (args: SurveyDataChartProps) => <Layout colorScheme="auto"><Card><SurveyDataChart {...args} /></Card></Layout>
+
+export const ffwelCreativeSelfLineChart = {
+    args: {
+        title: "FFWEL",
```
reneefromhold, line 11: nit on the title. I can see reading the code where it came from, but without that context just looking at storybook it looks like a typo

src/components/container/SurveyDataChart/SurveyDataChart.tsx
```diff
@@ -0,0 +1,176 @@
+import React, { useState, useEffect, useRef, useContext } from 'react'
+import { DateRangeContext } from '../../presentational/DateRangeCoordinator/DateRangeCoordinator'
+import { DailyDataProvider, DailyDataQueryResult, checkDailyDataAvailability, queryDailyData } from '../../../helpers/query-daily-data'
+import { add, format, parseISO } from 'date-fns'
+import MyDataHelps, { SurveyAnswer, SurveyAnswersPage } from '@careevolution/mydatahelps-js'
+import getDayKey from '../../../helpers/get-day-key'
```
reneefromhold, line 6: Is this used?

basyl-durnan:
@chrisnowak I worry about a team that made an error and not realizing it for some time since it wouldn't stand out. However, I could imagine there being a "skip" or "prefer not to answer" option and they wouldn't want an error because of that; so probably best to dump.

src/components/container/SurveyDataChart/SurveyDataChart.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+        return () => {
+            MyDataHelps.off("applicationDidBecomeVisible", loadCurrentInterval);
+            MyDataHelps.off("externalAccountSyncComplete", loadCurrentInterval);
+        }
+    }, [props.intervalType, props.weekStartsOn, dateRangeContext]);
+
+    var currentDate = intervalStart;
+    var data: any[] = [];
+    var chartHasData: boolean = false;
```
reneefromhold, line 121: I don't see this variable used - chartHasData

src/components/container/SurveyDataChart/SurveyDataChart.tsx
```diff
@@ -0,0 +1,176 @@
...
+        MyDataHelps.on("applicationDidBecomeVisible", loadCurrentInterval);
+        MyDataHelps.on("externalAccountSyncComplete", loadCurrentInterval);
+
+        return () => {
+            MyDataHelps.off("applicationDidBecomeVisible", loadCurrentInterval);
+            MyDataHelps.off("externalAccountSyncComplete", loadCurrentInterval);
+        }
+    }, [props.intervalType, props.weekStartsOn, dateRangeContext]);
+
+    var currentDate = intervalStart;
```
reneefromhold, line 119: Unused variable

src/components/container/SurveyDataChart/SurveyDataChart.tsx
reneefromhold, line 1: Would like to see a NoData and Live story

src/components/presentational/DataChart/DataChart.tsx
reneefromhold, line 1: This component will not show in storybook without a story. 
Can a simple Default and NoData be added?

reneefromhold (commented):
Overall, lg and worthwhile to have. 
I had a few nits and comments on unused items. 
I added a NoData story to the SurveyData component, but it was erroring (code below).  Could be I am misunderstanding how to set it up. Can we have Defaults, NoData, and Live versions for the SurveyData.

/* 
export const NoData = {
    args: {
        title: "No Data",
        options: {
            domainMin: 0,
            lineColor: ["#e41a1c", "#377eb8", "#4daf4a"],
            areaColor: ["#d41a1c", "#277eb8", "#3daf4a"]
        },
        intervalType: "Week",
        weekStartsOn: "6DaysAgo",
        lines: [{ label: "Creative Self", surveyName: "FFWEL", stepIdentifier: "CreativeSelf", resultIdentifier: "CreativeSelf", stroke: "#e41a1c" },
                { label: "Coping Self", surveyName: "FFWEL", stepIdentifier: "CopingSelf", resultIdentifier: "CopingSelf", stroke: "#377eb8" },
                { label: "Social Self", surveyName: "FFWEL", stepIdentifier: "SocialSelf", resultIdentifier: "SocialSelf", stroke: "#4daf4a" }],
        valueFormatter: (value: number) => Number(value.toFixed(0)).toLocaleString(),
        chartType: "Area",
        previewDataProvider: (start: Date, end: Date) => {
            const data: SurveyAnswersPage[] = [];
            return Promise.resolve(data);
        }
    },
    render: render
}; */

src/components/presentational/DataChart/DataChart.tsx
```diff
@@ -0,0 +1,242 @@
...
+import { AxisDomain } from 'recharts/types/util/types'
+import { ColorDefinition, resolveColor } from '../../../helpers/colors'
+
+export interface DataChartProps {
+    title?: string
+    intervalType?: "Week" | "Month",
+    intervalStart: Date,
+    data: any[],
+    dataKeys?: string[],
+    hasAnyData: boolean,
```
chrisnowak, line 15: I'm not sure there's much utility in moving hasAnyData and hideIfNoData down to this; feels like it's actually putting in more code than you're saving?

chrisnowak:
This refactoring is awesome since yeah, obviously we have 2 use cases for the same underlying display.  I had held off on that mostly since I didn't have conviction that the chart was generalizable and actually going to be used 1:1, and most of the "presentational" stuff is just recharts widgets anyways...but yeah this is a pretty clear win.

chrisnowak:
Something seems off here with the DailyDataChart - there's no bar here but it shows up with a tooltip value (looks like the previously "grey" bars do not appear in this story)

Screenshot 2024-04-26 at 12 59 02 PM

ajmenca:
> Something seems off here with the DailyDataChart - there's no bar here but it shows up with a tooltip value (looks like the previously "grey" bars do not appear in this story)
> 
> Screenshot 2024-04-26 at 12 59 02 PM

Which story are you seeing this in?  The "Steps Bar With Thresholds Chart" always seems to have data every day of the week, no matter how many times I refresh/remount it - I can't get data gaps to show up.

ajmenca:
> Something seems off here with the DailyDataChart - there's no bar here but it shows up with a tooltip value (looks like the previously "grey" bars do not appear in this story)
> 
> Screenshot 2024-04-26 at 12 59 02 PM

Ah, I'm seeing this too - I think since the data is randomly generated, I had gotten a couple of "lucky" views where everything was either the green or red bar, and not the gray one.  I'm guessing this is something weird about the gradientkey.  Will fix.

ajmenca:
image

src/components/presentational/DataChart/DataChart.tsx
ajmenca: More code in what way?  This seemed in line with the paradigm of the parent container doing the "heavy lifting" for the data and then passing parameters down to the child component to tell it how it needs to render at any point in time.  The empty graph placeholder is useful - so if these properties aren't passed down here, then the block of code for displaying the "No Data" label or the `LoadingIndicator` needs to get copy/pasted in both the `DailyDataChart` and the `SurveyDataChart`, right?  That's what I was avoiding as it explicitly felt like "more code."

As Renee noted above, I didn't really deal well with these for the SurveyDataChart; I'd punted on them in my initial push of doing this work a few months ago, and then I forgot that I didn't deal very well with them.  That was mainly because it didn't feel like there was a good equivalent for the `checkAvailability` call - but as noted above, I wanted to leverage the empty graph with the loading indicator.

chrisnowak:
Couple quick reactions:
- I would think we would want to mimic the behavior of the DailyDataChart in terms of the time axis?  e.g. ensuring it is stable so you can see the time between things instead of not displaying empty days?  Like your example here where it's 4 days between the first 2 data points and 30+ days between the 2nd and 3rd
Screenshot 2024-04-26 at 1 41 15 PM
- We could establish other grouping options than "day" for instance, e.g. week or month, but I'm struggling to see a scenario where don't want to space things to just graph them like they are next to each other if they have time in between them...so my 2c is to mimic the DailyDataChart
- If you have multiple answers on the same day, what happens now?  It takes the most recent?

An alternate design that I was considering before was just to allow the DailyDataChart to take in a separate set of parameters so it could query survey answers...but it would use the same underlying date grouping/rendering by day...that is actually why I originally named it "DailyDataChart" instead of "DailyDeviceDataChart"

Broadly I am a little concerned that "DataChart" is way too broadly named as well...what makes it worth componentizing / reusing externally vs. just using the recharts underlying recharts components...

chrisnowak:
Am getting one of these:
Screenshot 2024-04-26 at 1 52 17 PM

src/components/container/SurveyDataChart/SurveyDataChart.tsx
```diff
@@ -0,0 +1,176 @@
...
+	surveyName: string;
+	stepIdentifier: string;
+	resultIdentifier: string;
+}
+
+export interface SurveyDataChartProps {
+    title?: string
+    intervalType?: "Week" | "Month"
+    weekStartsOn?: WeekStartsOn
+    lines: SurveyAnswerChartLine[],
```
chrisnowak, line 21: Seems weird to call these lines when it might be a bar or area chart?

src/components/container/SurveyDataChart/SurveyDataChart.tsx
```diff
@@ -0,0 +1,176 @@
...
+import { DailyDataProvider, DailyDataQueryResult, checkDailyDataAvailability, queryDailyData } from '../../../helpers/query-daily-data'
+import { add, format, parseISO } from 'date-fns'
+import MyDataHelps, { SurveyAnswer, SurveyAnswersPage } from '@careevolution/mydatahelps-js'
+import getDayKey from '../../../helpers/get-day-key'
+import { WeekStartsOn, getMonthStart, getWeekStart } from '../../../helpers/get-interval-start'
+import DataChart, { AreaChartOptions, BarChartOptions, LineChartOptions } from '../../presentational/DataChart/DataChart'
+
+export interface SurveyAnswerChartLine {
+	label: string;
+	surveyName: string;
```
chrisnowak, line 12: I would leave surveyName / stepIdentifier as optional and just require resultIdentifier

src/components/container/SurveyDataChart/SurveyDataChart.tsx
```diff
@@ -0,0 +1,176 @@
...
+	stepIdentifier: string;
+	resultIdentifier: string;
```
chrisnowak, line 14: Actually maybe just have this take surveyAnswerQueryParameters?  I don't think the abstraction is useful and could get in the way of certain use cases, like if you have pain scores being captured from multiple distinct steps and you want to graph them as the same line.  surveyAnswerQueryParameters should let you pass in multiple result identifiers etc

src/components/presentational/DataChart/DataChart.tsx
```diff
@@ -0,0 +1,242 @@
...
+    areaColor?: ColorDefinition | ColorDefinition[],
+}
+
+export default function DataChart(props: DataChartProps) {
+    let layoutContext = useContext(LayoutContext);
+    let intervalType = props.intervalType || "Month";
+
+    let chartHasData = props.data.length > 0;
+
+    const DayTick = ({ x, y, stroke, payload }: any) => {
```
chrisnowak, line 50: So this chart is still definitely time-based and just supports week or month formatting and daily values.  Unfortunately "DailyDataChart" was taken (I guess I should have named it DailyDeviceDataChart).

How about "TimeSeriesChart" - that actually feels like it is the distinguishing factor here.  We have time-series data and want to graph it...the thing that separates this from any other data chart is that we have written some logic for the x-axis etc...to properly display day of week / month etc...

chrisnowak: I think we could also take a swing at extending this into Intraday data with https://github.com/CareEvolution/MyDataHelpsUI/pull/247/files - @reneefromhold I am not sure we should have a "LineChart" component...we could try to use this one and modify it so the x-axis properly supports intraday data (I think seems straightforward - just intervalType = day)

ajmenca:
> Couple quick reactions:
> 
> * I would think we would want to mimic the behavior of the DailyDataChart in terms of the time axis?  e.g. ensuring it is stable so you can see the time between things instead of not displaying empty days?  Like your example here where it's 4 days between the first 2 data points and 30+ days between the 2nd and 3rd
>   

This is an interesting perspective.  It totally makes sense to "normalize" the results when you're expecting the data set to *be* daily, meaning a gap is important (i.e. you somehow... took no steps).  At least with the graphs being generated by the project that initially drove this visualization, they weren't at a consistent cadence - or probably more correctly put, they weren't at a *perfectly* consistent cadence.  That is, you might be assigned the task once a week, but have the whole week to fill it out or something?  So for them, I don't think having a larger gap visually between the points was interesting/valuable.  However, a second case I developed for them they *did* want to highlight weeks where you missed the survey - but in that case I drove the x-axis dates based on the time of the assigned Task, not the date of the Survey Response. 

Very interesting to consider - let me put some thought into how to potentially handle those cases cleanly here.

> * If you have multiple answers on the same day, what happens now?  It takes the most recent?

Basyl asked this above - I think that should be correct.  He mentioned that some people might want an "average" of answers if they were taken on the same day; certainly from a graphing perspective we need to condense them down to a single point.  Hadn't totally resolved this yet.

> An alternate design that I was considering before was just to allow the DailyDataChart to take in a separate set of parameters so it could query survey answers...but it would use the same underlying date grouping/rendering by day...that is actually why I originally named it "DailyDataChart" instead of "DailyDeviceDataChart"
> 

Admittedly, I was a little confused by the name "DailyDataChart" when it was introduced; but it's name matches everything in `query-daily-data` where we never use the term "Device".  We have a `DailyDataType` enum which is a list of types of device data.  So I thought we were just trying to simplify the name somehow.

> Broadly I am a little concerned that "DataChart" is way too broadly named as well...what makes it worth componentizing / reusing externally vs. just using the recharts underlying recharts components...

I agree this is probably a little too generic.  It's worth componentizing because it's at least encapsualting a standard *way* of using Recharts (and hopefully simplifying its use, since it can be touchy), as far as sane defaults that match color schemes, etc.  It also standardizes the loading indicator behavior, things like that.

src/components/container/SurveyDataChart/SurveyDataChart.tsx
ajmenca: Well, the main driver was that the SurveyAnswerQueryParameters has parameters for time range which we're setting via the `DateRangeContext` - having this parameter be an object with fields that we might *ignore* seemed a bad interface since there's no way to indicate that we'd discard them (or if we didn't discard them, that they'd override the `DateRangeContext`).

As for limiting the other use cases - I'm not sure I understand how you'd graph multiple steps from the same survey on the same line?  The X-axis is time, so multiple responses from the same survey would be at the same x location, which means they can't be on the same line?  They'd have to be parallel lines, and so you'd configure them separately?

src/components/presentational/DataChart/DataChart.tsx
chrisnowak: Ah I forgot about the "noData" label.  Nevermind - I thought this was just to check for hiding the graph

src/components/container/SurveyDataChart/SurveyDataChart.tsx
ajmenca: Yeah, this is a legacy of bringing it over from a project where I only supported lines, will change.

src/components/presentational/DataChart/DataChart.tsx
ajmenca: Yeah - @reneefromhold have spoken; mainly she asked if I could integrate the line threshold colors she implemented there.  I'm working on updating things so that she could remove her LineChart in favor of this one.

ajmenca:
> * This is why line graphs exist; for making the data / trend look continuous when it's not

Well, there's a difference between continuous and normalized.  But that's getting pedantic.  

> * Graphing survey results by themselves is unlikely to be that interesting.  Many of our use cases involve co-locating, say, your mood score with your steps.  Being able to stack the graphs on top of each other is actually pretty important so you can correlate (this is actually another reason we might want to consider just extending DailyDataChart)
>

Well, to be fair, the simple case by itself was interesting for the project I implemented it for.  I see what you're saying though for other cases where the frequency of the survey is more rapid.  To be clear - I think this implementation *wouldn't* have actually worked for that project because the repeated survey was at more of a monthly cadence; long changes over time... which actually our Date Navigators don't really support at the moment (i.e. they'd want the time displayed from program entry to "now" most likely).  But that's a separate problem.

> Yeah, I was thinking we actually would extend queryDailyData and DataTypes to (somehow) include survey answer data...since you can register custom "DailyDataType" providers for instance. So right now you can actually get the DailyDataChart to graph survey answer (or whatever other) type of data so long as you register the provider

This is... a very interesting suggestion.  One that I certainly hadn't considered.  The main issue is that the current daily data providers don't need any parameters other than the type of data you're looking for.  For Surveys you need to select some combination of result identifier, etc as extra configuration.  Let me play around with this idea and see how ugly it gets.

chrisnowak:
Thinking about it more, I do maybe see *some* use cases for non-normalized survey answers, but I think it would be something like if you were getting scores taken at specific time points where you'd actually want to label the time points as the x-axis...like pain score at "visit 1", "visit 2", "visit 3" where there could be arbitrary time between those.  Harder to make that generic I think.

But I think otherwise not normalizing feels like it's only a potential shortcut if we wanted to add a "year" view or something of answers, or display them all-time...which admittedly may be a use case but doesn't really fit with the existing chart and interval options which are week/month based with daily ticks, and if you did add a "year" view you'd still want to have them normalized I think...e.g. if you miss a month of answers, you would want to know that

Anyways, I think if we add a not-normalized view we should probably have a strong use case for it and make sure it fits that use case well.  I *definitely* know the normalized daily chart using week/month intervals is a use case...e.g. graphing daily pain scores after surgery...so my personal bias is to stick with that but others may see other/broader use cases!

Oh...actually with that use case in mind...we may want to make intervalStart an optional prop that we can specify as well...since I think one use case for this might be to chart answers relative to a specific anchor date - (pain scores following surgery, mood scores following enrollment for a 30-day study, etc.).  Not critical but I see more use cases where that might be useful here vs. for device data

chrisnowak:
@ajmenca Not sure it *really* impacts this other than possible merge conflicts at some point, but I did tackle a bit of a refactor of querying daily data: https://github.com/CareEvolution/MyDataHelpsUI/pull/248

I wasn't really thinking about querying survey answers here though, mostly was for ease of use of setting up my new components in that PR and removing some stuff we hardcoded into view builder

It did make me think though, that the way the daily data providers etc. are set up that you could achieve a "Survey Answer Chart" component only in view builder.  Could just call registerDailyDataProvider to register a data provider that provides a numeric survey answer score daily based on the parameters to the component.  Obviously, we're limited to one survey answer there, but it might be a starting point for an MVP.

ajmenca:
> @ajmenca Not sure it _really_ impacts this other than possible merge conflicts at some point, but I did tackle a bit of a refactor of querying daily data: #248
> 
> I wasn't really thinking about querying survey answers here though, mostly was for ease of use of setting up my new components in that PR and removing some stuff we hardcoded into view builder
> 
> It did make me think though, that the way the daily data providers etc. are set up that you could achieve a "Survey Answer Chart" component only in view builder. Could just call registerDailyDataProvider to register a data provider that provides a numeric survey answer score daily based on the parameters to the component. Obviously, we're limited to one survey answer there, but it might be a starting point for an MVP.

Yeah, that's definitely merge conflicts down the line.  The "dynamic" registering of a daily data provider was definitely a solution I'd been considering (the alternative being modifying `queryDailyData` to take an extra optional parameters object to be able to convey query parameters for surveys).  

For a variety of reasons, the dynamic registration ends up being a better option, mostly because of the fact that if we want to have  the `DailyDataChart` be able to graph survey data directly so that it can be displayed *alongside* of device data, the `DailyDataChart` needs to have its properties updated to be able to take arrays of configuration instead of the single line that it takes now (i.e. I pushed this sort of change into the *renderer component* and made the `SurveyDataChart` support it out of the box, and need to move those types of changes into the `DailyDataChart` before removing the `SurveyDataChart`.

So, that's going to be my current push.

chrisnowak:
@ajmenca my 2c honestly is that having device / survey data on the same graph may not actually be worth it unless you have a firm use case in mind.  Or - maybe a "multiple data type" chart should be a separate component so we don't end up complicating the base case DailyDataChart

In particular I think it'll present a lot of complications with axes for instance - you potentially have to render multiple because obviously steps won't make sense directly next to something like a pain score.  

This is potentially well handled by the other set of components in that PR; the "relative activity" stuff; which also relies on the daily data types as well but renders everything relative to averages or set thresholds.

I could see a few paths forward:
- Adding a basic survey answer chart to view builder seems very straightforward if you just want to use the registerDailyDataProvider stuff (or registerDailyDataTypeDefinition in the new PR); it will be single parameter but this is likely sufficient for a large number of use cases
- Also potentially thinking about how to add support for survey answer data types to the Relative Activity stuff...maybe a general approach is needed for view builder where we actually have a set of properties on the top-level view for registering survey answer data sources, and these are picked up by the various charts as options for display?
- + maybe a multiple data source chart; which I think would be cool as well but presents some unique challenges and feels to me like a unique visualization (maybe sharing some of the underlying chart components though)

ajmenca:
After some experimentation, reviewing of implementations from previous projects, and regrouping with @reneefromhold to discuss supporting intraday data, here's where I'm at:

- The project that I originally implemented this for was displaying scores from a survey taken *monthly*.  This, by itself shifted my perspective of a handful of things.  Our DateRangeCoordinator had no way to helps support this type of view.  So I added a "SixMonth" option to the DateRangeNavigator and Coordinator.
- This also helped convince me that, for now at least, making a `dailyDataProvider` just didn't inherently make sense - at least not as far as being the *general* way to get survey responses.
- In implementing a six month view, I ran into a lot of problems with the current implementation of the chart renderer:
  - The CartesianGrid attempts to put it's guide lines at the same places where the X and Y axis place their ticks.  For a 6 month view, this was a problem.  Generating ticks for the X-Axis was tricky; I wanted to ensure there was a tick at the beginning of every month, but this wasn't possible using the `interval` setting of the X-Axis as we were using it.  To make it work, I'd have to set the `interval` to 0, have a "dense" array mostly full of null entries (6 entries out of 180 days), and customize the tick label to be empty in most scenarios.  But then the CartesianGrid would put massive numbers of vertical lines (one on every day, even if there's no label).
  - With the perspective in mind that this new reusable renderer is for Time Data - I made the decision to switch the `scale` property of the X-Axis to `time`.  This has a bunch of benefits:
    - Users of the Chart no longer have to jump through hoops to make "dense" data arrays full of empty data.  This simplifies the data processing in both the DailyDataChart and the SurveyDataChart, and in talking with @reneefromhold will be easier for the HeartRateChart.
    - Instead of setting a tick interval on the X-Axis, I can instead generate X-Ticks based on the `intervalType`.  This gives fine control over the ticks - helpful for both the Six Month view as well as the Intraday view.  The tick generation is also pretty straight forward because the x units are millisecond values, so it's easy to create dates and labels.
    - To allow gaps in data to be presented, I added a parameter to the TimeSeriesChart which indicates when the chart should show a gap in data; so the DailyDataChart simply passes the raw data it has received from the API, and says that the expected interval is 1 day; if the TimeSeriesChart finds a gap of more than 1 day between points, it inserts a null value in between, which causes the rendering of the line to be broken.  This allows flexibility in the SurveyChart version, as you can specify whether the survey data is coming in daily, monthly, 6 months, etc.  If you don't want gaps at all, you can simply omit the parameter.
- I added support for Intraday viewing on the TimeSeriesChart.
- I added lots of stories to display all these variations.


So, to summarize:
  - DailyDataChart and SurveyDataChart both use a new TimeSeriesChart as a common renderer.
  - Users of the TimeSeriesChart don't have to do anything special to their data, other than setting the `timestamp` field to unix milliseconds.
  - TimeSeriesChart supports configurable timegap detection between data points in order to indicate missing data.
  - TimeSeriesChart supports day, week, month, six month views - matching capabilities provided by the DateRangeNavigator.

I think this puts this widget in a really good place for the straight forward use cases of displaying survey responses, both the long time span use case for monthly projects, as well as potentially daily responses for things such as pain surveys.  The DailyDataChart still works as previously implemented.  The TimeSeriesChart will be useful for @reneefromhold instead of the `LineChart` she started implementing, which will make 3 components utilizing the same renderer.

image
image
image

chrisnowak:
@ajmenca curious if you could outline your use cases a bit more...mostly with regards to wanting a "6 month" date range?  That seems like it will blow up the daily data charts and also seems like it will struggle to satisfy use cases where you really want to graph things over months because things will eventually fall out of 6 months (or there will be a ton of space around it)

Wondering if we should try to walk back to use cases a bit more to validate that this is going to satisfy them; I have some in mind for instance...and there's a few things we've done for CARES etc...

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -0,0 +1,321 @@
...
+    referenceLineColor?: ColorDefinition
+    overThresholdBarColor?: ColorDefinition
+}
+
+export interface AreaChartOptions {
+    lineColor?: ColorDefinition | ColorDefinition[],
+    areaColor?: ColorDefinition | ColorDefinition[],
+}
+
+export default function TimeSeriesChart(props: TimeSeriesChartProps) {
```
chrisnowak, line 48: First glance looking at the storybook, this looks quite nice and robust!  I had to fiddle with the axes a lot to get the bar/area charts working right, but looking at the stories it seems largely consistent.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -0,0 +1,321 @@
...
+            }
+            return <>
+                <text className={isToday(currentDate) ? "today" : ""} fill="var(--mdhui-text-color-2)" x={x} y={y + 8} textAnchor="middle" fontSize="11">{dayOfWeek}</text>
+                <text className={isToday(currentDate) ? "today" : ""} fill="var(--mdhui-text-color-2)" x={x} y={y + 24} textAnchor="middle" fontSize="12">{currentDate.getDate()}</text>
+            </>;
+        } else if ( intervalType == "Day"){
+            const startTime = new Date(props.intervalStart);
+            startTime.setHours(0, 0, 0, 0);
+            return <>
+                {currentDate.getTime() === startTime.getTime() && <text className={isToday(currentDate) ? "today" : ""} fill="var(--mdhui-text-color-2)" x={x} y={y + 8} textAnchor="middle" fontSize="11">{format(currentDate, "EEEEEE")}</text>}
```
chrisnowak, line 81: Is including this intentional?  Looks like there's a random day of week at the beginning:
Screenshot 2024-05-03 at 2 35 23 PM

Also looks like there's some random bolding of the times

chrisnowak: On the "Line Chart Intraday" story

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -0,0 +1,321 @@
...
+import getDaysInMonth from 'date-fns/getDaysInMonth'
+import { ceil } from 'lodash'
+import addHours from 'date-fns/addHours'
+
+export interface TimeSeriesChartProps {
+    title?: string
+    intervalType?: "Week" | "Month" | "SixMonth" | "Day",
+    intervalStart: Date,
+    data: any[] | undefined,
+    dataGap?: Duration,
```
chrisnowak, line 17: "dataGap" is definitely of a weird name; as much as "interval" is overloaded it probably makes more sense to call this "interval" since that is what recharts axes would call this I would think;  or perhaps "betweenDataInterval" or something...

src/components/container/SurveyDataChart/SurveyDataChart.tsx
```diff
@@ -0,0 +1,176 @@
...
+    function processPages(pages: SurveyAnswer[][]) {
+        var newDailyData: { [key: string]: SurveyAnswer[] } = {};
+        for (var i = 0; i < props.charts.length; i++) {
+            newDailyData[getDataKey(props.charts[i])] = pages[i];
+        }
+        
+        return newDailyData;
+    }
+    
+    useEffect(() => {
```
chrisnowak, line 93: Should be able to use "useInitializeView" instead of useEffect

chrisnowak:
I am wondering if instead of doing a "6Month" interval how hard it would be to have the TimeSeriesChart automatically calculate the interval/axes based on the data provided if no interval was supplied

For instance, let's say we drop one of these on the page without a date range context OR without specifying an interval.  It could just fetch the data all time and graph the survey answers for all time.  

As optional parameters you could supply it a start and end date as well; the end dates could be specified as a specific date, "Today", or "LastDataDate" for instance.  

That being said, I dunno how painful determining the axes would get in that case; you could end up with things like year boundaries being crossed etc.  

It does feel like it might cover more use cases for graphing data across a wide range of time than just a "SixMonth" data provider; which I think may end up feeling pretty inelegant/awkward; but I guess would be good to understand the use cases a bit more

chrisnowak:
(Obviously, "All Time" would not be an option for DailyDataChart - an interval would be required there)

ajmenca:
> @ajmenca curious if you could outline your use cases a bit more...mostly with regards to wanting a "6 month" date range? That seems like it will blow up the daily data charts and also seems like it will struggle to satisfy use cases where you really want to graph things over months because things will eventually fall out of 6 months (or there will be a ton of space around it)
> 

It's a lot of data for the DailyDataChart, yes.  I put support for it in there only because the stories for the DateRangeCoordinator use it as the test component, so it was the fastest way to make sure that my changes were working.  I can pull support for it back out, and update the story for the 6 month use case to use the Survey Chart instead.

Again, as I mentioned a few times across the PR the use case for Survey Data wanting a 6 month view is pretty straight forward.  The program that I was initially supporting has their survey cycle on a monthly cadence.  So to graph those changes over time, you need more than the 1 month window that we currently support.  I chose 6 months here as sort of an "intermediate" window for monthly data; 3 months wouldn't show enough data, and 1 year the graph would look mostly empty for a long range of time.  6 months was a happy medium in this case.  It also happens to match the length of time that participants are in the specific program I'd worked with - but that's more of a coincidence than anything.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
ajmenca: It was intentional, but I can take it or leave it.  I was trying to provide a bit of "context" for the clock time being shown; it's perhaps unnecessary as that information will also (likely) be displayed in the DateRangeCoordinator.  I kind of put it there because the DateRangeCoordinator isn't *required* to provide context, in which case the data being displayed may be more confusing (though "Friday" in this case doesn't help much, where a full date may be better).  If it seems redundant, I'm fine to pull it out.

The bolding is... probably a holdover from the code that attempts to bold the labels on the X-Axis if the label is for "today" - should probably just pull that styling out in the intraday layout as it won't be providing any clarity.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
ajmenca: > "dataGap" is definitely of a weird name; as much as "interval" is overloaded it probably makes more sense to call this "interval" since that is what recharts axes would call this I would think; or perhaps "betweenDataInterval" or something...

Yes - I wasn't totally happy with this name.  I didn't think `interval` was specific enough.  The reason that I didn't call it something like `betweenDataInterval`, `expectedDataInterval`, or `expectedDataFrequency` was that I was hoping to convey that this variable was what was helping to display *gaps* in the data.  That is, I was trying to indicate why you would or would not bother to set the value.  I probably like `expectedDataInterval` the most at this point?

ajmenca:
> I am wondering if instead of doing a "6Month" interval how hard it would be to have the TimeSeriesChart automatically calculate the interval/axes based on the data provided if no interval was supplied
> 
> That being said, I dunno how painful determining the axes would get in that case; you could end up with things like year boundaries being crossed etc.

I considered something like this, and it didn't quite feel worth it at this point - mostly because it feels like we're *heavily* pushing toward the DateRangeCoordinator.  I think this could be worth it in the future, if say, we had widgets that could create date range context that *aren't* the DateRangeCoordinator, or the DRC supported arbitrary start/end dates.  At that point, it wouldn't be hard to change the TimeSeriesChart so that it based it's X-Axis on the Duration between the start and end points of it's data (i.e. IntraDay < 1 day, Daily <= 7 Days, Monthly <= 30 days, 6 month > 30 days).  But right now it didn't seem appropriate; it felt better to keep to similar-ish schemes.  I don't think anything I'm doing here would prevent backwards compatible tweaks in the future.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
chrisnowak: Ah yeah, I like expectedDataInterval or expectedDataFrequency

"Gap" feels weird because it's a negative; the idea of "gaps" is kind of an internal library thing

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
chrisnowak: Yeah i'd definitely take it out and make the styling non-bold for the intraday use case; I don't think the context would be useful since it'll either have an explicit label or be in the date range coordinator etc.

chrisnowak:
> Again, as I mentioned a few times across the PR the use case for Survey Data wanting a 6 month view is pretty straight forward. The program that I was initially supporting has their survey cycle on a monthly cadence. So to graph those changes over time, you need more than the 1 month window that we currently support. I chose 6 months here as sort of an "intermediate" window for monthly data; 3 months wouldn't show enough data, and 1 year the graph would look mostly empty for a long range of time. 6 months was a happy medium in this case. It also happens to match the length of time that participants are in the specific program I'd worked with - but that's more of a coincidence than anything.

Do you think this would have satisfied that use case in a really solid manner then?  I guess i'm feeling when it comes to view builder where this is just gonna not work/look good in a lot of cases...DateRangeCoordinator in view builder can't be defaulted to a specific start date for instance, so stuff is gonna move on the chart...feels like we're gonna have a lot of white space at the beginning/end of the graph for a 6 month interval etc...so for your case does it feel like it would be good enough out of box

In any case, if works for a 1 month / 1 week interval, that satisfies most of the use cases *i've* seen so addition of 6 month isn't a big deal to me, but I would probably not make it an option for DailyDataChart since it might cause a poor UX (that much data could take a long time to query in many cases, like Apple Health steps)...

> mostly because it feels like we're heavily pushing toward the DateRangeCoordinator

DateRangeCoordinator/DateRangeContext are really designed for specific simplified intervals; I don't think anything is saying that it's appropriate for all use cases of time series data; I think 1000% there is room for stuff that doesn't sit in a date range coordinator...it's just that we have a lot of use cases existing where it makes sense to navigate on a day/week/month basis...plenty of summary use cases where we may want to graph longer-term longitudinal data and those likely require a different approach (i.e. calculating date range based on data)

ajmenca:
> Do you think this would have satisfied that use case in a really solid manner then? I guess i'm feeling when it comes to view builder where this is just gonna not work/look good in a lot of cases...DateRangeCoordinator in view builder can't be defaulted to a specific start date for instance, so stuff is gonna move on the chart...feels like we're gonna have a lot of white space at the beginning/end of the graph for a 6 month interval etc...so for your case does it feel like it would be good enough out of box
> 

I think this would have gotten me 99% of the way there.  You're correct - from within ViewBuilder I would have still been slightly stuck - but I didn't build the project in ViewBuilder.  So I would have probably used the DRC but passed it an explicit start date, because it *does* support that, it's just not being exposed.  Or, I would have not had the DRC at all and just specified the start interval and intervalType directly on the SurveyDataChart.  But, if ViewBuilder supported setting the startInterval on *either* the SurveyDataChart with no DRC *or* setting it on the DRC, I would have been totally fine.

chrisnowak:
Got it, makes sense.  I mostly wanted to trace a path forward to view builder and make sure we were satisfying 1 anchor use case.  Having 6 months as an option on SurveyAnswerChart makes sense to me; it is mostly on DateRangeCoordinator where it feels weird; since as you pointed out you may not have used it there at all and just anchored it to a specific date (which feels like the right answer to me in your case if the study is less than 6 mos long)

So maybe when we add this to view builder I would suggest we:
- Do NOT add the 6 month option as an option for the date range coordinator itself; it probably will blow up DailyDataCharts etc.
- Do add it to this new SurveyAnswerChart component as a possible interval type
- Also add an editable "intervalStart" field which could be a string expression, e.g. it could be "{{enrollmentDate}}" or "05-01-2024", such that the chart could be anchored/stuck to a specific time period globally or per-participant
  - (This could be added to DailyDataChart as well)

That's my 2c though, i'll try to do a full review of this PR tomorrow

ajmenca:
> Got it, makes sense. I mostly wanted to trace a path forward to view builder and make sure we were satisfying 1 anchor use case. Having 6 months as an option on SurveyAnswerChart makes sense to me; it is mostly on DateRangeCoordinator where it feels weird; since as you pointed out you may not have used it there at all and just anchored it to a specific date (which feels like the right answer to me in your case if the study is less than 6 mos long)
> 
> So maybe when we add this to view builder I would suggest we:
> 
> * Do NOT add the 6 month option as an option for the date range coordinator itself; it probably will blow up DailyDataCharts etc.
> * Do add it to this new SurveyAnswerChart component as a possible interval type
> * Also add an editable "intervalStart" field which could be a string expression, e.g. it could be "{{enrollmentDate}}" or "05-01-2024", such that the chart could be anchored/stuck to a specific time period globally or per-participant
>   
>   * (This could be added to DailyDataChart as well)
> 
> That's my 2c though, i'll try to do a full review of this PR tomorrow

Most of this I agree with; I do have some extra thoughts on it that are not really relevant here.  Maybe worth doing a quick re-group on Slack with other folks that have been deal with dates/graphs/view builder as a whole - or maybe we just have another long discussion on a View Builder Issue instead.

chrisnowak (changes requested):
I got pulled away from my review due to some other issues, so gonna post my comments now after my initial first pass - but I still want to dig in a bit deeper into TimeSeriesChart

ClaysGit:
@coderabbitai review

src/components/container/DailyDataChart/DailyDataChart.tsx
```diff
@@ -120,27 +100,26 @@ export default function DailyDataChart(props: DailyDataChartProps) {
...
             var dayKey = getDayKey(currentDate);
-            if (currentData[dayKey] != undefined && currentData[dayKey] != null) {
-                dataDay.value = currentData[dayKey];
-                dataDay.rawValue = dataDay.value;
-                dataDay.date = currentDate;
-                if (props.valueConverter) {
-                    dataDay.value = props.valueConverter(dataDay.value);
-                }
-                chartHasData = true;
+            dataDay.value = currentData[dayKey];
```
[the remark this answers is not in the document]
ajmenca: Ah, it looks like (possibly on a reinstall) my VSCode is using a more recent version of TS than the workspace project is using.  Switching it to the workspace version gets me this error to popup.  Apparently TS gets smarter on version 5.  Maybe we should make a note to investigate upgrade the project as a whole.

src/components/container/DailyDataChart/DailyDataChart.tsx
```diff
@@ -60,21 +36,25 @@ export default function DailyDataChart(props: DailyDataChartProps) {
...
-    let layoutContext = useContext(LayoutContext);
-
     const dateRangeContext = useContext<DateRangeContext | null>(DateRangeContext);
     let intervalType = props.intervalType || "Month";
     let intervalStart: Date;
 
     if (dateRangeContext) {
-        intervalType = dateRangeContext.intervalType === "Day" ? "Week" : dateRangeContext.intervalType;
+        if(dateRangeContext.intervalType === "SixMonth") {
+            intervalType = "Week"
```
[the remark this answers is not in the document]
ajmenca: Yeah, I was trying to match the pre-existing condition.  However, I think changing it to Month but not changing the condition for Day makes the most sense.  If you "undershoot" and ask for Day, we bump it up to Week (the next smallest thing supported), and if you "overshoot" and ask for 6 Months of data, we go to the largest thing - Month.  I think it's going to be mildly confusing for an end user regardless (because we don't currently do a good job of indicating in Docs what's supported), but in this case they get the *closest* thing to what they might have actually asked for.

src/components/container/SurveyDataChart/SurveyDataChart.tsx
```diff
@@ -0,0 +1,170 @@
...
+    let intervalStart: Date;
+    if (intervalType === "Week") {
+        intervalStart = getWeekStart(weekStartsOn);
+    } else {
+        intervalStart = getMonthStart();
+    }
+    return intervalStart;
+}
+
+export default function SurveyDataChart(props:SurveyDataChartProps) {
```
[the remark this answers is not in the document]
ajmenca: I think I called it this because the specific values I was graphing were things that the survey *auto calculated* for you, so somewhere in my brain I concluded "These aren't technically *answers* to the survey."  But, I agree, from a technical perspective this gets data from the Survey Answer API, so it will be more clear if it's named that way.

src/components/container/SurveyDataChart/SurveyDataChart.tsx
```diff
@@ -0,0 +1,170 @@
...
+    const dateRangeContext = useContext<DateRangeContext | null>(DateRangeContext);
+    let intervalType = props.intervalType || "Month";
+    let intervalStart: Date;
+    
+    if (dateRangeContext) {
+        intervalType = dateRangeContext.intervalType === "Day" ? "Week" : dateRangeContext.intervalType;
+        intervalStart = dateRangeContext.intervalStart;
+    }
+    else {
+        intervalStart = getDefaultIntervalStart(intervalType, props.weekStartsOn);
```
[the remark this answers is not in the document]
ajmenca: True - current default behaviors are to go back the distance specified.  I didn't catch this discrepancy.  I've updated this to be similar to how the "Month" behavior works.  This will default to the 1st of the month from *5* months ago, making the current month the 6th month displayed in the data.  This is similar to how the Month graph works; it goes to the beginning of *this* month by default, rather than showing you "30 days" of data.

This also exposed an error in how I was generating the ticks for the 6 month graph, I've updated all that code so it's consistent and predictable when the 6 month span *doesn't* start on the 1st of a month.

src/components/container/SurveyDataChart/SurveyDataChart.tsx
```diff
@@ -0,0 +1,170 @@
...
+}
+
+export interface SurveyDataChartProps {
+    title?: string
+    intervalType?: "Week" | "Month" | "SixMonth"
+    weekStartsOn?: WeekStartsOn
+    charts: SurveyAnswerChartParameters[],
+    chartType: "Line" | "Bar" | "Area"
+    options?: LineChartOptions | BarChartOptions | AreaChartOptions,
+    expectedDataInterval?: Duration,
```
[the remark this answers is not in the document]
ajmenca: Sure - the FFWEL example that I keep going back to is expected at a monthly cadence.  So you could put an `expectedDataInterval` of `{months: 1}` here, and missing a month would come up with a gap in the chart.  Similarly that same program has a survey that's a Weekly Goal Check In where participants give a weekly update on a scale of 1-10 on how they're keeping up with their goals.  In that case the graph could be set to `expectedDataInterval: {weeks: 1}`.  In my customized version of that graph they actually wanted to graph "missed" weeks with an X icon which I think would be a pretty cool addition to this graph as an option, but I will keep that for a second pass.

So, the behavior is just like the DailyDataChart, except the user can set the expected cadence.

image

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -0,0 +1,320 @@
...
+            <CardTitle title={props.title}></CardTitle>
+        }
+        <div className="chart-container">
+            {(!props.chartHasData) &&
+                <div>
+                    {!!dataToDisplay &&
+                        <div className="no-data-label">No Data</div>
+                    }
+                    {!dataToDisplay &&
+                        <LoadingIndicator />
```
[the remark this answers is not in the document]
ajmenca: Fixed.

ClaysGit:
@coderabbitai pause

src/components/container/DailyDataChart/DailyDataChart.tsx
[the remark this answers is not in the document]
chrisnowak: Ah yep, we should upgrade Typescript, looks like 5.0 was released over a year ago

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -0,0 +1,320 @@
...
+}
+
+export interface LineChartOptions {
+    lineColor?: ColorDefinition | ColorDefinition[],
+    connectNulls?: boolean,
+    domainMin?: number | "Auto"
+}
+
+export interface BarChartOptions {
+    barColor?: ColorDefinition | ColorDefinition[],
```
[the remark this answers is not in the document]
ajmenca: I wrestled with this a bit because the parallel arrays didn't make me very happy either.  I was trying to walk a balance of not having 3 different interfaces for doing things between 3 different related but different widgets (the DailyDataChart, the SurveyAnswerChart, and the TimeSeriesChart).  I was trying to avoid having the DailyData and SurveyAnswer have "custom" interfaces that they each "adapted" into something for the TimeSeriesChart because that seemed more confusing.  In particular I really wanted them to be able to share these base FooChartOptions classes.  And in doing so, I was trying to not break the current interface of the DailyDataChart, so that any current users could keep on working, and I wouldn't have to change View Builder.  

Honestly, I feel strongly that changing the dataKeys for the TimeSeriesChart into objects instead of the current string array makes it clearer.  I think that having two separate places where you can define the colors is harder to discover and understand as an end user, and the fact that we'd have to convey the hierarchy of where we look for and prioritize those colors is a situation I'd prefer to avoid.

Here's two suggestions:

1) Move the `dataKeys` property off of the `TimeSeriesChartProps` and into each `FooChartOptions`.
```
export interface LineChartOptions {
    lineColor?: ColorDefinition | ColorDefinition[],
    dataKeys?: undefined | string[],
    connectNulls?: boolean,
    domainMin?: number | "Auto"
}

export interface BarChartOptions {
    barColor?: ColorDefinition | ColorDefinition[],
    dataKeys?: undefined | string[],
    thresholds?: BarChartThreshold[]
}
```
This at least *colocates* the parallel arrays so they're next to each other.  

2) Similar solution, but more like your suggestion, again removing the `dataKeys` property from `TimeSeriesChart`:
```
export interface LineChartOptions {
    lineColor?: ColorDefinition,
    seriesDefinition? : undefined | { dataKey: string, lineColor: ColorDefinition }[]
    connectNulls?: boolean,
    domainMin?: number | "Auto"
}

export interface BarChartOptions {
    barColor?: ColorDefinition,
    seriesDefinition? : undefined | { dataKey: string, barColor: ColorDefinition }[]
    thresholds?: BarChartThreshold[]
}
```
This has the issue of having to "prioritize" things, but at least in this case the properties are much closer together and maybe its more obvious?

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -0,0 +1,331 @@
...
+    thresholds?: BarChartThreshold[]
+}
+
+export interface BarChartThreshold {
```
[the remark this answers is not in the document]
ajmenca: I would prefer not to.  I think having such a generic name isn't helpful to have as an export when it's use is pretty narrow to its usage in this class.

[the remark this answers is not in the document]
reneefromhold: I'm indifferent to what it is named if Threshold is too generic. 
ChartThresholds is another option. 

Given a threshold lines are shown in both bar and line, it is confusing for the name to be bar chart specific.

src/components/container/SurveyDataChart/SurveyDataChart.tsx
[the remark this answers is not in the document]
chrisnowak: Ah...I'm sorry, I completely misunderstood/conflated what this was doing - probably because the storybook has an exact cadence with the graph - data is generated exactly 1 month apart.  So I assumed we actually were *grouping the data* by month; but we are not; data is still grouped by day for survey answers regardless, but we do not bother to "connect" days.  Your comment about how this is helping to display "Gaps" in the data also makes more sense now.

So yeah this makes sense...for daily data charts and such the expected cadence can always be hardcoded...for something like this I see where you could have an irregular cadence and want to specify over what time periods you have a "gap" in the graph...k cool

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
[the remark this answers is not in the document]
chrisnowak: Hmm...yeah neither option here feels right to me.

1 as you point out is still parallel arrays

2 is really burying a key required top-level property into sub-sub object so our chart would be like ```<TimeSeriesChart options={{seriesDefinition:{dataKey:"Steps", barColor:"blue"}}}/>```
which feels rough.

I would certainly not break the DailyDataChart interface/props, but I think it'd be ok to have a distinct DailyDataChartLineOptions, DailyDataChartBarChartOptions, etc. interfaces so we have freedom here...just changing types is not enough of a breaking change that I would feel like we'd need to cut a new major.  Or we can name these new options "TimeSeriesChartOptions"

So I think we do that^ and keep those types the same.  And then we can split our "options" from "series" such that we have:

```
export interface TimeSeriesChartProps {
    title?: string
    intervalType?: "Week" | "Month" | "6Month" | "Day",
    intervalStart: Date,
    data: any[] | undefined,
    expectedDataInterval?: Duration,
    series: ChartSeries[]
    chartHasData: boolean,
    tooltip: ({ active, payload, label }: any) => React.JSX.Element | null,
    chartType: "Line" | "Bar" | "Area"
    options?: TimeSeriesLineChartOptions | TimeSeriesBarChartOptions
    innerRef?: React.Ref<HTMLDivElement>
}

export interface ChartSeries {
    dataKey: string
    color: ColorDefinition
} 

export interface AreaChartSeries extends ChartSeries {
    areaColor: ColorDefinition // not sure how many use cases for multi-series area charts we have (or area charts in general), so could go either way on this
} 

export interface TimeSeriesLineChartOptions {
    connectNulls?: boolean,
    domainMin?: number | "Auto"
}

export interface TimeSeriesBarChartOptions {
    thresholds?: BarChartThreshold[]
}

export interface TimeSeriesBarChartThreshold {
    value: number
    referenceLineColor?: ColorDefinition
    overThresholdBarColor?: ColorDefinition
}
```

Something like that?  And now that I look at that ^ we could consider dropping "options" entirely.  ```domainMin ``` probably makes sense for all charts I think...not sure why it was just added to bars.  And sounds like we're adding thresholds to line charts as well...not sure if that is coming here or in @reneefromhold 's PR.  The only kinda oddball one feels like "connectNulls" which i'd be fine just putting on TimeSeriesChartProps even if it doesn't make sense for bar/area charts right now.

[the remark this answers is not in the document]
ajmenca: I've implemented this with the slight adjustment that instead of calling it `TimeSeriesLineChartOptions` I called them `MultipleSeriesLineChartOptions` because then the `SurveyAnswerChart` and `TimeSeriesChart` both share a similar interface because they both support showing multiple series of data.  The `DailyDataChart` is the only one that converts from its own property styles to the interface that `TimeSeriesChart` wants.  I've adjusted all the stories as necessary.

I'd talked to Renee and asked if she would add thresholds to line charts in her PR as this PR was large enough as it was, and adding it in her PR where it would be used the first time seemed appropriate.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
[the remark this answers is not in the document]
ajmenca: This has been adjusted slightly in the last update.  There's still a BarChartThreshold for the DailyDataChart, but there is a more generically named one that the SurveyAnswer and TimeSeries chart use (which eliminates the word Bar from both the interface name and the property inside the interface).  Keeping the existing one around serves to not break the current DailyDataChart interface which would technically require a major version number bump that I'd like to avoid.

basyl-durnan:
Do we anticipate a date yet for the next MDHDUI release and would this be included?

chrisnowak (approved):
Looks good!  I like the latest changes.

PR #251, opened 2024-04-27
Validate language before calling setCurrentLanguage

brandoncharnesky:
## Overview

Previously, calling setCurrentLanguage with the empty string generates a console warning and fails to clear the language.  

Instead, we manually set MyDataHelps.language to the empty string and do not call setCurrentLanguage.  This allows future calls of MyDataHelps.getCurrentLanguage() to return the browser default language.

## Security

No security risks, client side code changes only.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app
- [ ] StoryBook testing only

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -31,6 +31,7 @@ export const decorators = [(story, context) => {
 	if (context.globals.language) {
 		myDataHelps.setCurrentLanguage(context.globals.language);
 	} else {
+		// MDH.js does not currently support clearing current language
 		myDataHelps.language = "";
 	}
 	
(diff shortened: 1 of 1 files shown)


chrisnowak (approved):
Yeah this feels a bit hacky because realistically, MyDataHelps.js shoudl allow setting the language to an empty string or null or something if it lets setting the language explicity.  But overall seems reasonable to get rid of the warning for now.  Maybe just leave a comment above this for why we're doing this?

brandoncharnesky:
> Yeah this feels a bit hacky because realistically, MyDataHelps.js shoudl allow setting the language to an empty string or null or something if it lets setting the language explicity. But overall seems reasonable to get rid of the warning for now. Maybe just leave a comment above this for why we're doing this?

I have added an explanation comment in the code to explain why we are setting language directly.  I have also opened an issue in mydatahelps.js

PR #268, opened 2024-06-18
Fix a couple formatting issues in TimeSeriesChart

chrisnowak:
## Overview

Couple minor things I found when testing out a new branch using the latest version:

- The y-axis formatter wasn't quite doing the right thing; it was *always* displaying a single decimal place, which results in clipping when the number is in the thousands.  It's actually supposed to be```return Number(args.toFixed(1)).toString();``` so that it just displays *up to* one decimal place
- There was some margins added to the charts so there was a large white space to the right of the chart.  I don't think these manually added margins are necessary as far as I can tell?  @ajmenca not sure if i'm missing what these accomplished.  Wondering if they were to prevent clipping of the last x-axis tick, which maybe started happening because of the chart changes (it isn't happening in current view builder branch - tick seems to align itself inside the boundaries).  Either way, I added a quick CSS tweak to make the recharts-surface have overflow:visible which prevents the clipping

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

## Summary by CodeRabbit

- **Style**
  - Updated styling for enhanced tooltip and chart visibility in TimeSeriesChart.

- **Refactor**
  - Improved number formatting logic in TimeSeriesChart for more consistent display.
  - Removed specific margin settings from LineChart, BarChart, and AreaChart components for a cleaner layout.

### DIFF
--- a/src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
+++ b/src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
@@ -15,7 +15,7 @@ export interface TimeSeriesChartProps {
     title?: string
     intervalType?: "Week" | "Month" | "6Month" | "Day",
     intervalStart: Date,
-    data: Record<string,any>[] | undefined,
+    data: Record<string, any>[] | undefined,
     expectedDataInterval?: Duration,
     series: ChartSeries[] | AreaChartSeries[],
     chartHasData: boolean,
@@ -34,14 +34,14 @@ export default function TimeSeriesChart(props: TimeSeriesChartProps) {
         let currentDate = new Date(value);
         if (intervalType === "Month") {
             return <text className={isToday(currentDate) ? "today" : ""} fill="var(--mdhui-text-color-2)" x={x} y={y + 15} textAnchor="middle" fontSize="12">{currentDate.getDate()}</text>;
-        } else if (intervalType === "6Month" ){
+        } else if (intervalType === "6Month") {
             let monthLabel = currentDate.getDate() === 1 ? format(currentDate, "LLL") : "";
             let dayLabel = currentDate.getDate().toString();
             return <>
                 <text className={isToday(currentDate) ? "today" : ""} fill="var(--mdhui-text-color-2)" x={x} y={y + 8} textAnchor="middle" fontSize="11">{monthLabel}</text>
                 <text className={isToday(currentDate) ? "today" : ""} fill="var(--mdhui-text-color-2)" x={x} y={y + 24} textAnchor="middle" fontSize="12">{dayLabel}</text>
             </>;
-        } else if (intervalType === "Week" ){
+        } else if (intervalType === "Week") {
             let dayOfWeek: string = "";
             for (let i = 0; i < 7; i++) {
                 if (currentDate.getTime() === value) {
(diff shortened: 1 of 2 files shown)


src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -225,7 +225,7 @@ export default function TimeSeriesChart(props: TimeSeriesChartProps) {
             }
             {props.chartHasData && props.chartType === "Line" &&
                 <ResponsiveContainer width="100%" height={150}>
-                    <LineChart width={400} height={400} data={dataToDisplay} syncId="DailyDataChart" margin={{left: 5, top: 5, bottom: 5, right: 40}}>
+                    <LineChart width={400} height={400} data={dataToDisplay} syncId="DailyDataChart">
```
ajmenca, line 228: > * There was some margins added to the charts so there was a large white space to the right of the chart.  I don't think these manually added margins are necessary as far as I can tell?  @ajmenca not sure if i'm missing what these accomplished.  Wondering if they were to prevent clipping of the last x-axis tick, which maybe started happening because of the chart changes (it isn't happening in current view builder branch - tick seems to align itself inside the boundaries).  Either way, I added a quick CSS tweak to make the recharts-surface have overflow:visible which prevents the clipping

This margin *is* because specifically the Intraday version of the tick label was being cut off because it's longer than our other tick labels.  There's a bunch of things that contributed to this involving how and why I ended up changing the `interval` setting on the X-Axis and specifically generating X-Axis ticks in order to get consistent and reproducible behavior out of Recharts.  I can go over all that if you want, but the long story short is that I'm confident that I wouldn't want to change any of that code.

That said - your fix doesn't fully work, as shown in the story for Intraday:

image

I agree, in general though, that the right side margin does look a little strange (and probably stranger in context of MDH rather than in the Storybook).

My suggestion is to change line 61 where the label for this is generated to this:

```
                <text fill="var(--mdhui-text-color-2)" x={x} y={y + 8} textAnchor="middle" fontSize="12">{format(currentDate, "hh:mm")}</text>
                <text fill="var(--mdhui-text-color-2)" x={x} y={y + 24} textAnchor="middle" fontSize="12">{format(currentDate, "aaa")}</text>
```

Which, in combination with removing the margin and doing your overflow css looks like this:
image

Those y offsets look a little tight, to be honest, so maybe its worth bumping them both down a little bit?  They're copied from the offsets that we set on the two line label if you're in the Week view, which could also maybe use a bit of extra breathing room?
image

ajmenca (changes requested):

chrisnowak:
Ah @ajmenca actually in the designs pulled together for intraday chart, we just wouldn't show the 12am values - so that's not a big deal.  I will add the code to hide those...I think very much not necessary since context will make it pretty obvious:

e.g 
image

chrisnowak:
I've got that updated so the intraday chart no
Screenshot 2024-06-18 at 12 58 06 PM
w looks correct:

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
chrisnowak: Yeah they are a bit tight, i'll bump them down a couple px

greinard (approved):

PR #219, opened 2024-03-17
Address preview state in ConnectEhrStep and ConnectDeviceAccountStep

bxh:
## Overview

- Display preview state in ConnectEhrStep when in design mode.
- Require `deviceType` in ConnectDeviceAccountStep.

Address https://github.com/CareEvolution/MyDataHelpsDesignerUI/issues/350

## Security

REMINDER: All file contents are public.

- [x] I have ensured no secure credentials or sensitive information remain in code, metadata, comments, etc.
  - [x] Please verify that you double checked that .storybook/preview.js does not contain your participant access key details. 
  - [x] There are no temporary testing changes committed such as API base URLs, access tokens, print/log statements, etc.
- [x] These changes do not introduce any security risks, or any such risks have been properly mitigated.

Describe briefly what security risks you considered, why they don't apply, or how they've been mitigated.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app
- [x] [MyDataHelps website](mydatahelps.org)

### Documentation
- [ ] I have added relevant Storybook updates to this PR as well.
- [ ] If this feature requires a developer doc update, tag @CareEvolution/api-docs.

Consider "Squash and merge" as needed to keep the commit history reasonable on `main`.

## Reviewers

Assign to the appropriate reviewer(s). Minimally, a second set of eyes is needed ensure no non-public information is published. Consider also including:
- Subject-matter experts
- Style/editing reviewers
- Others requested by the content owner

PR #270, opened 2024-06-25
Fix Connect Device Account Props, Step Next Button border, Relative Activity

chrisnowak:
## Overview

Fixes a few bugs:
- Per my comment here: https://github.com/CareEvolution/MyDataHelpsUI/pull/219/files#r1651001440 the step containers aren't supposed to have props, much less required ones, since the point of them is to get the configuration from the SDK and then pass it to the step component itself.  This was blocking upgrading a downstream library
- Border color of buttons in steps wasn't being applied properly, this is fixed here
- There was a bug with having multiple "relative activity" components on the page at once, due to my code to prevent requests from coming back out of order.  This adjusts it so the currentRequestID is not a global but is stored in a ref so it doesn't mean that only one of these can render on the page at once!

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

## Summary by CodeRabbit

- **New Features**
	- Improved request management in `RelativeActivity` to prevent out-of-order processing using `useRef`.
	- Consolidated multiple story exports into a single default export for better maintainability in `ConnectDeviceAccountStepContainer`.

- **Enhancements**
	- Enhanced state management and asynchronous operation handling in `ConnectDeviceAccountStepContainer` by removing `deviceType` prop and initializing state properly.
	- Updated `StepNextButton` to ensure border color consistency based on `backgroundColor` prop.

### DIFF
--- a/src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
+++ b/src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
@@ -4,21 +4,16 @@ import ConnectDeviceAccountStep from "../ConnectDeviceAccountStep";
 import { getFitbitProviderID, getGarminProviderID, getOmronProviderID } from "../../../helpers/providerIDs";
 import { deviceType } from "../../../helpers/deviceType";
 
-export interface ConnectDeviceAccountStepContainerProps {
-    deviceType: deviceType
-}
 
-export default function (props: ConnectDeviceAccountStepContainerProps) {
+export default function () {
     const [title, setTitle] = useState<string>();
     const [text, setText] = useState<string>();
     const [styles, setStyles] = useState<any>({});
-    const [deviceType, setDeviceType] = useState<deviceType>(
-        props.deviceType
-    );
+    const [deviceType, setDeviceType] = useState<deviceType | undefined>(undefined);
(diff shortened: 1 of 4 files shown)


src/components/step/ConnectDeviceAccountStepContainer/ConnectDeviceAccountStepContainer.tsx
```diff
@@ -4,21 +4,16 @@ import ConnectDeviceAccountStep from "../ConnectDeviceAccountStep";
...
 
-export default function (props: ConnectDeviceAccountStepContainerProps) {
+export default function () {
     const [title, setTitle] = useState<string>();
     const [text, setText] = useState<string>();
     const [styles, setStyles] = useState<any>({});
-    const [deviceType, setDeviceType] = useState<deviceType>(
-        props.deviceType
-    );
+    const [deviceType, setDeviceType] = useState<deviceType | undefined>(undefined);
```
greinard, line 12: I think this can just be:

```const [deviceType, setDeviceType] = useState<deviceType>();```

By default it should already be a `deviceType | undefined` with a value of `undefined`.

chrisnowak: Huh, TIL

PR #271, opened 2024-06-26
Add a discrete scale component.

greinard:
## Overview

This branch adds a discrete scale input component.

discrete_scale

## Security

- No security risk.  This PR just adds a new input widget.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Summary by CodeRabbit

- **New Features**
  - Introduced a `DiscreteScale` component that allows users to interact with a discrete scale and select values.
  - Added styling for the `DiscreteScale` component for improved visual consistency.
  - Created a configurable story to showcase the `DiscreteScale` component with customizable properties like color scheme, slider color, tick count, and labels.

- **Documentation**
  - Updated export declarations to include the new `DiscreteScale` component in the presentational components index.

### DIFF
--- a/src/components/presentational/DiscreteScale/DiscreteScale.tsx
+++ b/src/components/presentational/DiscreteScale/DiscreteScale.tsx
@@ -51,11 +51,16 @@ export default function (props: DiscreteScaleProps) {
                     <UnstyledButton
                         className="mdhui-discrete-scale-ticks-section-button"
                         onClick={() => onClick(value)}
+                        style={{
+                            left: index === 0 ? '-10px' : '-50%',
+                            width: index === 0 ? 'calc(50% + 10px)' : '100%'
+                        }}
                     >
                         <div
                             className="mdhui-discrete-scale-ticks-section-marker"
                             style={{
                                 background: value === selectedValue ? sliderColor : 'transparent',
+                                left: index === 0 ? '-1px' : 'calc(50% - 11px)'
                             }}
                         >&nbsp;</div>
                     </UnstyledButton>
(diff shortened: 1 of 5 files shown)


src/components/presentational/DiscreteScale/DiscreteScale.tsx
```diff
@@ -0,0 +1,86 @@
...
+        background: sliderColor
+    } as CSSProperties;
+
+    const onClick = (value: number) => {
+        if (props.onChange) {
+            props.onChange(value);
+        }
+    };
+
+    return <div className="mdhui-discrete-scale" ref={props.innerRef}>
```
ajmenca, line 46: Not quite sure what where the best place to adjust this is, but I think the current layout is going to almost always cause horizontal scroll bars.  Both ends of the chart extend out past where they visually do because of the click targets:
image

I think that adding `overflow: hidden` to the `mdhui-discrete-scale` class is sufficient to fix it.  It does mean that the first and last click target still overflow the bar and you can click slightly *outside* of the bar to select them.  But - honestly I think that's probably fine.  The other alternative would be to do some `first-child`/`last-child` variations to make the `mdhui-discrete-scale-ticks-section-button` smaller.  But that's probably more work and not worth it.

ajmenca (changes requested):
I think the larger click targets are significantly better than the first pass - and functionally it's good that it matches the MDH version that you can get in a Survey.

I think there's one small tweak to still make with horizontal scroll bars showing up, but then it's good to go.

src/components/presentational/DiscreteScale/DiscreteScale.tsx
greinard: Good catch.  I've fixed the issue with the horizontal scroll.  The touch targets are now contained within the bar.

ajmenca (approved):

PR #279, opened 2024-07-12
Some updates to the TimeSeriesChart.

greinard:
## Overview

This branch updates the `TimeSeriesChart` to support overlaying chart components (i.e. children).

image

I also cleaned up some IDE warnings, re-organized the code a little bit, and refactored some code into helper functions.

I will go through and annotate the intended functional changes.  The rest are just part of the re-organization and code cleanup.

A new storybook entry named "Overlay" can be used to view/test the new functionality.

## Security

- No new security risks.  This update just allows us to overlay charting components onto one another when necessary.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Summary by CodeRabbit

- **New Features**
	- Introduced 'overlay' functionality for enhanced chart visualization, allowing users to overlay a bar on the chart with additional step generation logic.

- **Improvements**
	- Enhanced chart customization with new properties for `TimeSeriesChart`, including optional tooltips and flexible data handling.
	- Simplified interval calculations for intraday data generation.

- **Bug Fixes**
	- Corrected object formatting in chart exports to ensure proper rendering and functionality.

- **Refactor**
	- Streamlined chart rendering logic and data processing functions for better maintainability and performance.

### DIFF
--- a/src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
+++ b/src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
@@ -6,7 +6,6 @@ import addDays from "date-fns/addDays";
 import { predictableRandomNumber } from "../../../helpers/predictableRandomNumber";
 import { Meta, StoryObj } from "@storybook/react";
 import { ChartThreshold, MultiSeriesLineChartOptions } from "../../../helpers";
-import { isSameDay, startOfDay } from "date-fns";
 import { Bar } from "recharts";
 
 const meta: Meta<typeof TimeSeriesChart> = {
@@ -342,36 +341,6 @@ export const loading: Story = {
     },
 };
 
-const generateSteps = (date: Date) => {
-    let steps: { timestamp: Date, value: number }[] = [];
-
-    let timestamp = add(startOfDay(date), { minutes: 15 });
-    while (isSameDay(timestamp, date)) {
-        let newValue = Math.round((Math.random() * 200) + 20);
-        if (timestamp.getHours() < 7 || timestamp.getHours() > 22) {
-            newValue = 0;
-        }
-
-        steps.push({
-            timestamp: timestamp,
-            value: newValue
-        });
-
-        timestamp = add(timestamp, { minutes: 30 });
-    }
-
-    return steps;
-};
-
-let steps = generateSteps(new Date());
-let maxSteps = Math.max(...steps.map(r => r.value));
-let stepsScale = 220 / (maxSteps * 0.8);
-let overlaySteps = steps
-    .filter(r => r.value > 0)
-    .map(r => {
-        return { ...r, value: r.value * stepsScale }
-    });
-
 export const overlay: Story = {
     args: {
         title: "Line Chart With Overlay",
@@ -380,12 +349,35 @@ export const overlay: Story = {
         chartHasData: true,
         series: [{ dataKey: 'value' }],
         intervalStart: new Date(),
-        options: { hideDots: true },
-        children: <Bar data={overlaySteps} type="monotone" dataKey="value" fill="#f5b722" opacity={0.3} radius={[2, 2, 0, 0]} />
+        options: { hideDots: true }
     },
     loaders: [
-        async () => ({
-            randomData: await getRandomIntradayData(new Date(), 15)
-        })
-    ]
+        async () => {
+            let steps = await getRandomIntradayData(new Date(), 30);
+
+            steps.forEach(s => {
+                if (s.timestamp.getHours() < 7 || s.timestamp.getHours() > 22) {
+                    s.value = 0;
+                } else {
+                    s.value += 20;
+                }
+            });
+
+            let maxSteps = Math.max(...steps.map(r => r.value));
+            let stepsScale = 220 / (maxSteps * 0.8);
+            let overlaySteps = steps.filter(r => r.value > 0).map(r => {
+                return { ...r, value: r.value * stepsScale }
+            });
+
+            return {
+                randomData: await getRandomIntradayData(new Date(), 15),
+                children: <Bar data={overlaySteps} type="monotone" dataKey="value" fill="#f5b722" opacity={0.3} radius={[2, 2, 0, 0]} />
+            };
+        }
+    ],
+    render: (args: TimeSeriesChartProps, { loaded: { randomData, children } }) => <Layout colorScheme="auto">
+        <Card>
+            <TimeSeriesChart {...args} data={randomData} children={children} />
+        </Card>
+    </Layout>
 };
--- a/src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
+++ b/src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
@@ -6,6 +6,8 @@ import addDays from "date-fns/addDays";
 import { predictableRandomNumber } from "../../../helpers/predictableRandomNumber";
 import { Meta, StoryObj } from "@storybook/react";
 import { ChartThreshold, MultiSeriesLineChartOptions } from "../../../helpers";
+import { isSameDay, startOfDay } from "date-fns";
+import { Bar } from "recharts";
 
 const meta: Meta<typeof TimeSeriesChart> = {
     title: "Presentational/TimeSeriesChart",
@@ -84,7 +86,7 @@ function getRandomIntradayWithinDomainMultipointData(start: Date) {
     return responses;
 }
 
-async function getRandomIntradayData(start: Date) {
+async function getRandomIntradayData(start: Date, intervalMinutes?: number) {
     var responses: { timestamp: Date, value: number }[] = [];
     let currentTime = new Date(start);
     currentTime.setHours(0, 0, 0, 0);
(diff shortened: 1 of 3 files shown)


src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -226,93 +171,117 @@ export default function TimeSeriesChart(props: TimeSeriesChartProps) {
...
         }
         <div className="chart-container">
-            {(!props.chartHasData) &&
-                <div>
-                    {!!dataToDisplay &&
-                        <div className="no-data-label">No Data</div>
+            {!props.chartHasData && !!dataToDisplay && <div className="no-data-label">No Data</div>}
+            {!props.chartHasData && !dataToDisplay && <LoadingIndicator />}
+            <ResponsiveContainer width="100%" height={150}>
+                <ComposedChart data={dataToDisplay} syncId={props.syncId}>
```
greinard, line 194: The `ComposedChart` allows us to overlay different types of chart components (i.e. Lines with Bars, etc).  This is the main functional change in this PR.

Once all of the charts were being created as  `ComposedChart` components, there was no need to repeat the boilerplate container and chart tags, so now those are at the higher level and only the conditional chart-specific stuff is rendered within the conditionals (per chart type) below.

This also meant the standard chart components were only going to be used in one place, so there was no longer a need to have that extracted to a helper function.  Those components are just rendered inline now.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -226,93 +171,117 @@ export default function TimeSeriesChart(props: TimeSeriesChartProps) {
...
+                    {props.chartHasData && props.tooltip &&
```
greinard, line 195: The `tooltip` property is now optional.  If omitted, we will not add the `Tooltip` component to the chart.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -226,93 +171,117 @@ export default function TimeSeriesChart(props: TimeSeriesChartProps) {
...
+                        axisLine={false}
+                        dataKey="timestamp"
+                        tickMargin={0}
+                        minTickGap={0}
+                        tickLine={false}
+                        ticks={xAxisTicks}
+                        includeHidden
+                        interval={0}
+                    />
+                    {props.children}
```
greinard, line 223: Here is where the overlay children are rendered.  This is the reason for switching to the `ComposedChart`.

The children can be things such as references lines, bars, lines, areas, etc.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -226,93 +171,117 @@ export default function TimeSeriesChart(props: TimeSeriesChartProps) {
...
+                                            type="monotone" dataKey={dk}
+                                            dot={!((props.options as MultiSeriesLineChartOptions)?.hideDots ?? false)}
+                                            stroke={`url(#${gradientKey}${i})`}
+                                        />
+                                    )}
+                                </>
+                            }
+                            {props.chartType === "Bar" &&
+                                <>
+                                    {createBarChartDefs(layoutContext, gradientKey, props.series, props.options)}
```
greinard, line 246: The bar and area defs creation was moved to helper functions (like line defs) to keep this code a little smaller/cleaner looking.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
```diff
@@ -96,7 +98,7 @@ async function getRandomIntradayData(start: Date) {
             timestamp: currentTime,
             value: (await predictableRandomNumber(currentTime.toISOString())) % 200
         });
-        currentTime = add(currentTime, { minutes: 5 });
+        currentTime = add(currentTime, { minutes: intervalMinutes ?? 5 });
         console.log(currentTime);
```
reneefromhold, line 102: Clean out console logs

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
reneefromhold: The ComposedChart supports an accessibilityLayer property. Any reason why we wouldn't want to add that now? Info in the [storybook ](https://recharts.org/en-US/storybook)

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -1,57 +1,48 @@
...
+    intervalStart: Date;
+    data?: TimeSeriesDataPoint[];
+    expectedDataInterval?: Duration;
+    series: ChartSeries[] | AreaChartSeries[];
+    chartHasData: boolean;
+    tooltip?: ({ active, payload, label }: any) => React.JSX.Element | null;
+    chartType: "Line" | "Bar" | "Area";
+    options?: MultiSeriesLineChartOptions | MultiSeriesBarChartOptions;
+    syncId?: string;
+    children?: React.ReactNode;
```
reneefromhold, line 33: Nit: not thrilled with this variable name if it will only ever be overlay. I see it named the same in one of the recharts stories, though in that case children could be y/x axis, tooltip, series data.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
```diff
@@ -342,4 +344,52 @@ export const loading: Story = {
...
         tooltip
     },
+};
+
+const generateSteps = (date: Date) => {
+    let steps: { timestamp: Date, value: number }[] = [];
+
+    let timestamp = add(startOfDay(date), { minutes: 15 });
+    while (isSameDay(timestamp, date)) {
+        let newValue = Math.round((Math.random() * 200) + 20);
```
reneefromhold, line 354: Nit: Why not use one of the existing random number generator functions in this same module?

reneefromhold (approved):
The meat of this looks good. Just a few nits and a question about using the accessibility property of ComposedChart

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
ajmenca: This, as far as I know, has to be named this way in order for React to properly pass you child DOM elements: 
 https://react.dev/reference/react/Children.  i.e. it's what allows this structure:

```
<TimeSeriesChart>
    <Bar>...</Bar>
    <Area>....</Area>
</TimeSeriesChart>
```

If you named it something else you'd be forced to do:
```
<TimeSeriesChart overlays={[<Bar>...</Bar>, <Area>...</Area>]}/>
```

@greinard I don't know why I didn't originally ask this when we were on a call last week, but - the whole reason you want this sort of functionality rather than, say, moving the `chartType` property onto `ChartSeries` property (which would allow you to specify that you want to show one series as a Line, one series as a Bar) is that you're basically exposing explicit control over the exact rendering of the overlayed charts, right?  

In some ways, while the second format is ugly, I can kind of see @reneefromhold's point that the mixed usage is slightly confusing.  It does, in some ways, open the door to passing *no* `series` to the chart, but then passing `children` and expecting something sane to happen (which, I think at this point it would not).  Giving it an explicitly named parameter *does* sort of nudge the usage in a particular direction and could be considered slightly more "self documenting."  

I'm not sure I feel super strongly one way or the other - as we sort of discussed in our chat, this wades into the territory of how much good are we doing by "wrapping" the charting library versus explicitly exposing it.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
greinard: I'm not familiar with the `accessibilityLayer` property and I'm unable to find it in any of the API docs.  The only place I see it mentioned is in that storybook, but I'm not sure what it is supposed to be doing.

Do you have any further info you can link me to or can you describe what it would do for us?

src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
greinard: I went through the file and removed all the console logs.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
greinard: Yeah.  As mentioned, the `children` property is just the built-in naming convention to be able to use the nested tags, and I think we want to retain that capability.

I also think that name works okay in that I happen to be using it for what I'm calling an "overlay", but really, it is just allowing a user of the `TimeSeriesChart` to add any supported recharts child component for rendering without always needing to update the `TimeSeriesChart` API to support a specific need.

Moving the chart type to the series may help cover the bars with lines use case that I have right now, but allowing generic children to be added to the `TimeSeriesChart` also allows for adding custom `ReferenceArea`, `ReferenceLine`, `Brush`, `Legend`, `Line`, `Area`, `Bar`, `Customized`, etc tags as needed.

We may want to consider adding first class support for those kinds of things to the `TimeSeriesChart`, but for the time being, having the `children` property will allow for custom chart implementations without the need to expand upon the shared component.  And yes, I do have some reservations about how far we continue down the path of wrapping rechart components rather than just using them directly.  🙂

src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
greinard: I have updated the story to use the existing random intraday data function to generate the steps data.

PR #280, opened 2024-07-16
HtmlStep

ClaysGit:
## Overview

Most of Phase 1 in this (non-public) issue: https://github.com/CareEvolution/MyDataHelpsDesignerUI/issues/631

image

Adds a new step type: HtmlStep. This step writes user-provided content to the step's DOM, much like our WebView step. The difference is that this step should come with baked in support for various common needs, so that the user can focus on functional stuff and have an easier time getting it to look good+consistent with other styles.

The new step has a few pieces to it:
- HtmlStep.tsx: Imports and runs other pieces to host all the content. Also, custom steps can receive user-defined styles. This file is where those styles are used to set CSS variables.
- InnerHtml.tsx: Where the user's content is transformed from string to nodes on a DOM. To support scripting this runs in a couple steps, detailed below.
- elements.css and classes.css: Contains the new default styles for HtmlStep content. I've copied in some common styles, but these need some suggestions to finish.
- HtmlStep.stories.tsx: Some examples of different styling and scripting capabilities
- /HtmlStepContainer directory with simple container element.

Additional changes to support the HtmlStep:
- Add build support for importing files as plain strings. This applies to imports for files ending in `?raw`.
  - Very useful to do work with editor language features on stuff that will ultimately be encoded as a string. In this case 
- Use the `storybook-addon-fetch-mock` package to simulate MDH.js delegated access requests

### InnerHtml.tsx details

This parses a string of HTML content to a [DocumentFragment](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment). This basically creates a tree of nodes, without any processing past that.

The component moves the top-level child trees from the fragment over to the main document one at a time. It waits after each move to let script loading and execution finish.

Waiting for each move is accomplished using a MutationObserver. Before each move, a new Promise is created which only resolves when a mutation observer detects new changes. The observer resolves the promise as soon as changes are detected unless there are <script> tags with src attributes, in which case it resolves only when all of those tags' onload/onerror events are dispatched.

Again, the point of this complexity is to support external scripts (jQuery in my tests/stories) loading in order. The document does not load these synchronously otherwise.

### Regarding CSS choices

I'm not confident in all of the styles I've chosen, and I've also left some blank. I've sourced the current styles from other steps in MDH UI and also from recent prod projects, but I fully expect there are some better ones I'm not aware of. Some direction would be appreciated, or I'll ask around if the rest of this PR is looking good. Others are also welcome to modify the CSS files.

Additional sources for styles
- `"V2" Survey Styles` PR
- `Convert Survey Answer Formats/Questions to React` PR

#### onLoad support

The document.readyState is "complete" throughout the HtmlStep rendering process. This removes one of the easy ways developers are able to detect that a page has finished loading.

Users can attach onload events to elements in their HTML content, but we should probably add a global method for registering event handlers which are dispatched once the step is fully loaded.

#### survey question/answer styles

Include a simplified version of the CSS used to improve survey answer formats/questions. This could be a new .css file alongside the others, injected into Emotion.Global.

## Security

This change mostly adds a new set of content to be rendered in a WebView step. This is already an unsafe environment, and there are adequate existing precautions against malicious actors trying to cross the iframe boundary or similar in native environments.

Beyond that, there are some minor build changes. These changes add special loading rules for files imported with `?raw` at the end. They are done using built in tools, and require no additional packages. This change to the build pipeline does not create additional risk.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

I ran into issues testing this locally, related to window.postMessage from localhost. I can keep working around that, but it'd be more straightforward to test this from .dev viewlibrary.

### Documentation
- [X] If this feature requires a developer doc update, tag @CareEvolution/api-docs.

@CareEvolution/api-docs This adds a new step type to document

## Reviewers

@ashleyrtate @ajmenca @chrisnowak 

## Summary by CodeRabbit

- **New Features**
  - Introduced `HtmlStep` and `HtmlStepContainer` components for rendering HTML-based survey steps.
  - Added support for dynamic HTML content and styles within survey steps.
  - Integrated jQuery for DOM manipulation within the `HtmlStep` component.
  - Storybook stories for `HtmlStep` and `HtmlStepContainer` demonstrating various use cases.

- **Enhancements**
  - Added CSS styles for step-based UI components to ensure consistent visual elements.
  - Customizable visual properties for HTML elements using CSS variables.


chrisnowak:
Awesome - sorry I forgot about the whole scripts loading out of order thing; cool that you got it working but imagine that was a pain?

I'm gonna have a ton of tweaks to the default CSS here more than likely...thinking about how we want to structure that given that step styles differ in some slight ways from MDHUI styles, but I want to move steps more in the direction of MDHUI styles in terms of padding/margins/etc...

ClaysGit:
Closing for now. Likely going to re-implement in separate repo.

PR #281, opened 2024-07-17
TimeSeriesChart Fixes

ajmenca:
## Overview

In discussing some work that @greinard was working on (https://github.com/CareEvolution/MyDataHelpsUI/pull/279) we talked over a few bugs he'd come across that I said I'd tackle.

1) The No Data story wasn't working; this was a quick fix as the story was just set up wrong.

2) The DailyDataChart stories were generating an extra data point.  This was caused by a discrepancy in generating data from the "start of a day" - it was causing the loop which generates fake data to generate too much data, and Recharts was stubbornly trying to display it even though the domain is constrained.  I've fixed both the dates *and* set `allowDataOverflow` on the X-Axis which should (despite how it's named) prevent extra data outside of the range from being displayed (the variable "allows data to overflow" on the graph which means Recharts doesn't attempt to show everything you give it).

3) The newly introduced thresholds for line charts weren't actually working correctly.  They look like they are because in the stories, the data range being generated was almost always covering the fully range of the graph.  However, if you adjusted the story so that it only generated data in the middle of the graph, it would look like this (you can see here that data is red well below the red threshold line, and yellow well below the yellow threshold line):
image

This is because the `<stop>` added to a `<linearGradient>` SVG element is described in percentages - but it's percentages of the *displayed SVG* - and in the case where the line's "height" isn't full range of the graph, doing the stop calculation as percentages of the Y-Axis is incorrect (50% of the SVG's height isn't 50% up the Y-Axis).  This fix for this is to adjust for using the specific data line's min/max values as the domain, and *not* the actual y-domain, which makes the graph properly look like this when the line isn't the full domain:
image

And if it is the full domain, the calculation still works:
image

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

## Summary by CodeRabbit

- **New Features**
  - Introduced a new chart story `multipleLineChartWithThresholdsSmallRange` with custom data loading logic for better range handling in TimeSeriesChart.

- **Enhancements**
  - Improved data generation in TimeSeriesChart stories with new parameters for finer control over the data range.
  - Updated DailyDataChart to start the interval at the beginning of the day for more accurate data representation.
  - Enhanced TimeSeriesChart component with streamlined handling of domain boundaries and improved Y-axis data management.

- **Bug Fixes**
  - Fixed an issue with the `noData` story in TimeSeriesChart by setting `chartHasData` to false for accurate representation.

### DIFF
--- a/src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
+++ b/src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
@@ -67,7 +67,7 @@ function getRandomInt(min: number, max: number) {
     return Math.floor(Math.random() * (max - min + 1)) + min;
 }
 
-function getRandomIntradayWithinDomainMultipointData(start: Date) {
+function getRandomIntradayWithinDomainMultipointData(start: Date, domainMin: number = 0, domainMax: number = 200) {
     var responses = [];
     let currentTime = new Date(start);
     currentTime.setHours(0, 0, 0, 0);
@@ -75,9 +75,8 @@ function getRandomIntradayWithinDomainMultipointData(start: Date) {
     while (currentTime < endTime) {
         responses.push({
             timestamp: currentTime,
-            key1: (getRandomInt(0, 200)),
-            key2: (getRandomInt(0, 200)),
-            key3: (getRandomInt(0, 200)),
+            key1: (getRandomInt(domainMin - 6, domainMax + 15)),
+            key2: (getRandomInt(domainMin - 2, domainMax + 5)),
         });
         currentTime = add(currentTime, { minutes: 5 });
     }
@@ -278,6 +277,25 @@ export const multipleLineChartWithThresholds: Story = {
     loaders: [
         async () => ({
             randomData: await getRandomIntradayWithinDomainMultipointData(new Date())
+        })
+    ]
+};
+
+export const multipleLineChartWithThresholdsSmallRange: Story = {
+    args: {
+        title: "Multiple Line Chart with Thresholds",
+        intervalType: "Day",
+        chartType: "Line",
+        chartHasData: true,
+        series: [{ dataKey: 'key1', color: 'blue' }, { dataKey: 'key2', color: 'purple' }, { dataKey: 'key3', color: 'black' }],
+        data: undefined,
+        intervalStart: new Date(),
+        options: multiSeriesLineOptions,
+        tooltip
+    },
+    loaders: [
+        async () => ({
+            randomData: await getRandomIntradayWithinDomainMultipointData(new Date(), 20, 150)
         })
     ]
 };
--- a/src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
+++ b/src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
@@ -323,12 +323,17 @@ export const noData: Story = {
         title: "No Data Chart",
         intervalType: "Week",
         chartType: "Line",
-        chartHasData: true,
+        chartHasData: false,
         data: [],
         series: [{ dataKey: 'value' }],
         intervalStart: new Date(),
         tooltip
     },
+    loaders: [
+        async () => ({
+            randomData: []
+        })
+    ]
 };
 
 export const loading: Story = {
--- a/src/helpers/chartHelpers.tsx
+++ b/src/helpers/chartHelpers.tsx
@@ -3,38 +3,47 @@ import { isNumber } from "lodash";
 import { ChartThreshold, MultiSeriesLineChartOptions } from "./chartOptions";
 import { resolveColor } from "./colors";
 import { LayoutContext } from "../components";
+import { TimeSeriesDataPoint } from '../components/presentational/TimeSeriesChart/TimeSeriesChart';
 
-export function createLineChartDefs(layoutContext: LayoutContext, gradientKey: string, seriesLineColors: string[], multiSeriesLineChartOptions?: MultiSeriesLineChartOptions) {
-    let nDomainMin: number | undefined = undefined;
-    let nDomainMax: number | undefined = undefined;
+export function createLineChartDefs(
+    layoutContext: LayoutContext, 
+    gradientKey: string, 
+    seriesLineColors: string[], 
+    multiSeriesLineChartOptions: MultiSeriesLineChartOptions | undefined,
+    dataKeys: string[],
+    data: TimeSeriesDataPoint[] ) {
 
-    if (multiSeriesLineChartOptions) {
-        if (isNumber(multiSeriesLineChartOptions.domainMin)) nDomainMin = multiSeriesLineChartOptions.domainMin;
-        nDomainMax = multiSeriesLineChartOptions.domainMax;
-    }
-
-    const getPercent = function (numerator: number, domainMin: number, domainMax: number): number {
-        if ((domainMax - domainMin) > 0){
-            return (numerator - domainMin) / (domainMax - domainMin) * 100;
+    const getPercent = function (numerator: number, min: number, max: number) {
+        if ((max - min) > 0){
+            return (numerator - min) / (max - min) * 100;
         }
         return 0;
     }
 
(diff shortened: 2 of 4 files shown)


src/helpers/chartHelpers.tsx
```diff
@@ -47,20 +56,11 @@ export function createLineChartDefs(layoutContext: LayoutContext, gradientKey: s
...
                 </linearGradient>
-            )
-        }
-
-        return (
-            <linearGradient id={`${gradientKey}${i}`} key={`${gradientKey}${i}`} x1="0" y1="0" x2="0" y2="100%">
-                <stop offset="0%" stopColor={lineColor} />
-                <stop offset="100%" stopColor={lineColor} />
-            </linearGradient>
-        )
```
ajmenca, line 63: I simplified all of this down to a single return statement.  With the domain of the chart not being relevant to stops, it is correct in all cases to just call `createStopsFromThresholds`, which will create the 0% and 100% stops, and then do nothing if no thresholds are defined in between.

src/helpers/chartHelpers.tsx
```diff
@@ -3,38 +3,47 @@ import { isNumber } from "lodash";
...
-export function createLineChartDefs(layoutContext: LayoutContext, gradientKey: string, seriesLineColors: string[], multiSeriesLineChartOptions?: MultiSeriesLineChartOptions) {
-    let nDomainMin: number | undefined = undefined;
-    let nDomainMax: number | undefined = undefined;
+export function createLineChartDefs(
+    layoutContext: LayoutContext, 
+    gradientKey: string, 
+    seriesLineColors: string[], 
+    multiSeriesLineChartOptions: MultiSeriesLineChartOptions | undefined,
+    dataKeys: string[],
+    data: TimeSeriesDataPoint[] ) {
```
ajmenca, line 14: The main change here is that the dataKeys and data need to be passed in so that each line's min/max can be calculated.  I broke this out into single lines because it was getting extremely long.

src/helpers/chartHelpers.tsx
```diff
@@ -3,38 +3,43 @@ import { isNumber } from "lodash";
...
-    const createStopsFromThresholds = function (defaultLineColor: string, domainMin: number, domainMax: number, chartThresholds?: ChartThreshold[]) {
+    const createStopsFromThresholds = function (defaultLineColor: string, chartThresholds: ChartThreshold[] | undefined, index: number) {
         let stops: any[] = [];
         var lineColor: string = defaultLineColor;
         var thresholds = chartThresholds ?? [];
 
         thresholds.sort((a, b) => a.value - b.value);
 
-        if (thresholds.length && domainMin >= thresholds[0].value) {
+        const lineRange = data.reduce((result, dataPoint) => ({
```
reneefromhold, line 30: The domain max was initially passed in so that the chart could continue letting the y max be auto calculated, while still being able to know the range for the sake of calculating gradient stops. 

For single lines where the y domain is passed in, can the re calculation of max/min be avoided? In the case of [ hr intraday  ](https://github.com/CareEvolution/MyDataHelpsUI/blob/029e0e4cb7b9a7f6ee6839e7ce14e62733dfb5ab/src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx#L95) , it already doing the leg work of figuring out what the max should be.

ajmenca: I think that I would argue that in a follow up PR to this one, that we could remove the calculation from the HR graph completely, and let it set the domain max to auto.  But, that somewhat depends on what the exact behavior from the HR graph needs to be.  It's something we can have further conversation on, and would potentially impact the change made here https://github.com/CareEvolution/MyDataHelpsUI/pull/281/files#diff-c7d85721f79b358b3ab7238e978537aa98ffcf42fc45afbb6a1eb497b473892fR225 to be a configurable behavior.

However, from the perspective of the TimeSeriesChart as a reusable widget, I think it's correct for this calculation to be down here in the utility functions; I wouldn't want to push the responsibility of calculating the line min/max to parent components because it's a detail that's very tied to exactly how we're rendering the threshold colors.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
```diff
@@ -68,17 +68,16 @@ function getRandomInt(min: number, max: number) {
...
     let currentTime = new Date(start);
     currentTime.setHours(0, 0, 0, 0);
     let endTime = add(currentTime, { hours: 24 });
     while (currentTime < endTime) {
         responses.push({
             timestamp: currentTime,
-            key1: (getRandomInt(0, 200)),
-            key2: (getRandomInt(0, 200)),
-            key3: (getRandomInt(0, 200)),
+            key1: (getRandomInt(domainMin - 6, domainMax + 15)),
```
greinard, line 79: Are these value adjustments just to add a little variation in the data limits for the two keys?

ajmenca: Yeah - this is to demonstrate that the two lines *don't* have the exact same min and max, and yet the gradient is applied correctly to both.  Without this, with the amount of data generation across one day, the two lines end up basically always being the same.

greinard: Also, probably not a big deal, but you may want to adjust the default values of the `domainMin` and `domainMax` parameters, since this will result in an effective range of `-6` to `215` in the default scenario.

ajmenca: This was actually on purpose, because I wanted to demonstrate that passing a smaller `domainMin` and `domainMax` than the range of the graph actually constrainted the y-axis to what you'd asked (that `allowDataOverflow` flag).

This sort of relates to @reneefromhold's questions above... and predicting what a user would expect from this base widget.  To me, if you pass a domainMin/domainMax it means you're *trying* to enforce that range because you aren't asking for auto.  However, I can see an argument of "I want the Y-Axis to go to *at least* 200, but go higher if the data is out of range".  Right now I'm constraining it and not allowing for that case (which would require passing `allowDataOverflow` to be configurable.

src/helpers/chartHelpers.tsx
```diff
@@ -3,45 +3,47 @@ import { isNumber } from "lodash";
...
         let lineColor: string = defaultLineColor;
         let thresholds = chartThresholds ?? [];
 
         thresholds.sort((a, b) => a.value - b.value);
 
-        if (thresholds.length && domainMin >= thresholds[0].value) {
+        const lineRange = data.reduce((result, dataPoint) => ({
+            min: (dataPoint[dataKeys[index]] < result.min || result.min === 0) ? dataPoint[dataKeys[index]] : result.min,
+            max: (dataPoint[dataKeys[index]] > result.max || result.max === 0) ? dataPoint[dataKeys[index]] : result.max,
+        }), { min: 0, max: 0 });
```
greinard, line 37: I think the `min` and `max` may need to be initialized to `undefined` and then checked for that in the `||` conditions.

Otherwise, I think if the `min` value is ever correctly set to `0` throughout the loop, the next data point will set the new minimum, even if it is greater than the correct minimum of `0`.

An alternative might be to do something like this:
```
let values = data.map(dataPoint => dataPoint[dataKeys[index]]);
let lineRange = {
	min: Math.min(...values),
	max: Math.max(...values)
}
```

src/components/presentational/TimeSeriesChart/TimeSeriesChart.stories.tsx
greinard: Ah, yeah, makes sense.  👍

greinard (approved):
Looks good to me.  I just had the one final question about the min/max computation.

Feel free to :ship: when ready.

src/helpers/chartHelpers.tsx
ajmenca: Wow, great catch.  I've fixed this by removing the explicit check against 0 as an initial value, and replaced the initial value with the first value in the data set.

PR #282, opened 2024-07-19
HtmlStep CSS Tweaks

chrisnowak:
@ClaysGit I figured it would be way easier to do this via PR rather than individual comments, especially since this is top of mind for me as i'm working on the survey styles in designer

Some notes:
- Theres some consistency tweaks with Designer (perhaps even stuff we are inconsistent about in other MDHUI steps); minor things like font sizes and button disabled state
- I wrapped the html step content in a class "mdhui-html-step" as well as most of the classes (and ended up just combining the file)  This was mostly me thinking through a bit of future proofing mostly...e.g. if we wanted a version of this without the default styles....we could add another property to step config for the top-level class to switch what styles we want (or eliminate the boilerplate styles)
  - One could of course use a top level div to denote this as well but that ends up nesting all the html underneath that, which would clutter up the editor later
- Added some new basic styles for inputs
- Shortened some of the class names; there's no reason to have the prefix in this context
- Applied next button styles to all buttons...I figure this is better for brevity...can always be overridden


src/components/step/HtmlStep/classes.css
```diff
@@ -1,101 +1,218 @@
...
 
-.mdhui-step-spinner {
-    border: 10px solid #f3f3f3;
-    border-top: 10px solid #2c62ce;
-    border-radius: 50%;
-    width: 60px;
-    height: 60px;
-    animation: spin 1s linear infinite;
-    margin: 2em auto;
+.mdhui-html-step .mdhui-step-spinner {
```
ClaysGit, line 201: This `.mdhui-step-spinner` could also be shortened

src/components/step/HtmlStep/HtmlStep.tsx
```diff
@@ -33,8 +32,7 @@ export default function (props: HtmlStepProps) {
     const customStyleVars = makeCustomStyleVars(props.styles);
 
     return (
-        <>
-            <EmotionGlobal styles={elementsCSS} />
+        <div className="mdhui-html-step">
```
ClaysGit, line 35: There's already a div in InnerHtml.tsx, and this className should work just as well there. To avoid having unneeded nested divs.

src/components/step/HtmlStep/HtmlStep.stories.tsx
```diff
@@ -51,10 +51,10 @@ export const HtmlStepElementStyles: StoryObj<typeof HtmlStep> = {
...
         html: `
-<div class="mdhui-step-title">title</div>
-<div class="mdhui-step-text">text</div>
-<button class="mdhui-step-next-button">next</button>
-<button class="mdhui-step-next-button" disabled>disabled next</button>
+<div class="title">title</div>
+<p>text</p>
+<button>Next</button>
+<button disabled>Disabled Next</button>
 <div class="mdhui-step-image">
```
ClaysGit, line 58: We can remove the mdhui-step-image stuff from stories, as it's not styled anymore

ClaysGit (approved):
Thanks for taking the lead on this. These look great, and I just have some notes for cleanup.

Regarding alternate versions of this step (no default styles, etc), I do think it makes sense to have a property in the editor to control behavior. Similarly, I've been planning to include a hidden property on the editor, with some arbitrary "version" value hard-coded. The HtmlStep should be able to select the stylesheet (or do more fine-grained selection) based on the version prop. So that we can update our css and existing steps will still use the styles they were built with.

PR #283, opened 2024-07-19
Add some new pass-through options to the TimeSeriesChart.

greinard:
## Overview

This branch adds a number of "pass-through" options to allow for customization of `TimeSeriesChart` rendering.

I've defined our own "options"  interfaces and types rather than using the Recharts ones directly.  However, the property names and types do (and must) match those of the Recharts analogs since I am just passing them along using destructuring.

There were a couple of what could be considered "breaking" changes, but only if `TimeSeriesChart` has been used directly from outside of the MDH-UI library.  The `connectNulls` property was moved and the `hideDots` property was moved and renamed.  The necessary updates have been made to the components within MDH-UI that were using those fields.

## Security

- No security risks.  These are just rendering customization options.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Summary by CodeRabbit

- **New Features**
	- Enhanced configurability for the TimeSeriesChart, allowing for custom rendering options.
	- Introduced a new story `customizedLineChart` showcasing improved data loading and chart rendering.

- **Improvements**
	- Adjusted chart options structure for better organization and clarity across multiple chart components.
	- Dynamic calculation of chart domains and ticks based on the current date for enhanced visual representation.
	- Streamlined control flow and rendering logic in the TimeSeriesChart for improved performance.

- **Bug Fixes**
	- Improved handling of dot visibility in line charts, streamlining user experience.


src/helpers/chartOptions.ts
```diff
@@ -1,50 +1,83 @@
...
-    thresholds?: BarChartThreshold[]
+export interface TimeSeriesChartXAxisOptions {
+    domain?: [number, number];
+    ticks?: (string | number)[];
+    tickFormatter?: (value: any, index: number) => string;
+}
+
+export interface TimeSeriesChartYAxisOptions {
+    width?: number;
+    domain?: [number, number];
```
ajmenca, line 46: It's a little confusing to have this here as well as on the `MultiSeriesLineChartOptions`.  I realize you want it here because you want to be able to customize it for all charts and not just lines.  As you said, there's a few other "breaking" changes in this that I agree are not worth worrying about yet - so I'd probably advocate for removing `domainMin` and `domainMax` on `MultiSeriesLineChartOptions` and updating the handful of places internally that will break (which will likely be mostly a bunch of stories may or may not actually break at compile time).

greinard: Ah, yeah, good call.  I'll do that as well.

greinard: This change has been made.  I believe I have found/updated all of the existing usage of those two fields.

ajmenca:
Code looks fine - the story itself looks a little off though.  The X-Axis is in the middle of the chart and the Y-Axis way overflows.  I realize there's probably a lot of ways to pass weird combinations of arguments, but it doesn't make a great/clear story in this case?
 
image

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
```diff
@@ -99,7 +99,16 @@ export default function (props: IntradayHeartRateChartProps) {
         iHrData = undefined;
     }
 
-    const options: MultiSeriesLineChartOptions = { connectNulls: false, hideDots: true, thresholds: props.thresholds, domainMin: yMinValue, domainMax: yMaxValue };
+    const options: MultiSeriesLineChartOptions = {
+        thresholds: props.thresholds,
+        yAxisOptions: {
+            domain: [yMinValue, yMaxValue]
```
reneefromhold, line 105: yMaxValue can be auto following the most recent changes to TimeSeries

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -224,24 +208,25 @@ export default function TimeSeriesChart(props: TimeSeriesChartProps) {
...
                     {props.children}
                     {props.chartHasData &&
                         <>
                             {props.chartType === "Line" &&
                                 <>
                                     {createLineChartDefs(layoutContext, gradientKey, props.series, props.options, keys, dataToDisplay!)}
-                                    {(props.options as MultiSeriesLineChartOptions)?.thresholds?.filter(t => t.referenceLineColor)?.map(threshold =>
-                                        <ReferenceLine y={threshold.value} stroke={resolveColor(layoutContext.colorScheme, threshold.referenceLineColor)} />
+                                    {(props.options as MultiSeriesLineChartOptions)?.thresholds?.filter(t => t.referenceLineColor)?.map((threshold, index) =>
+                                        <ReferenceLine key={index} y={threshold.value} stroke={resolveColor(layoutContext.colorScheme, threshold.referenceLineColor)} />
```
reneefromhold, line 220: add a prefix to the key (example key={`refLine_${index}`}

greinard:
> Code looks fine - the story itself looks a little off though. The X-Axis is in the middle of the chart and the Y-Axis way overflows. I realize there's probably a lot of ways to pass weird combinations of arguments, but it doesn't make a great/clear story in this case?

Yeah, for sure.  I'll get that fixed up.

I'm actually a little stumped about it, because I know it was working previously.  I specifically added the odd domain range to show that it worked.  🤷‍♂️

I'll see if I can pinpoint why it isn't rendering correctly now.

greinard:
Ah, it has something to do with the `allowDataOverflow` property.  🙂   I must not have checked it again after rebasing the latest from `main`.

Without that property: 
image

I'll see if I can figure out what I need to adjust to get it to work while still using that setting.

greinard:
Ah, it was the min domain in my story setup.  It needed to be adjusted to fit the custom ticks.  It is working correctly now.

src/components/presentational/TimeSeriesChart/TimeSeriesChart.tsx
greinard: I've updated them to be more specific keys.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
greinard: I'm not sure I understand this completely.  Which recent change are you referring to?

reneefromhold (approved):
I pushed a commit for the Intraday component. The domain needed to be 0, auto following the most recent time series changes prior to this PR and the support of 'auto' on the y max in this pr. 

Everything else looked good to me.

src/components/container/IntradayHeartRateChart/IntradayHeartRateChart.tsx
greinard: No concerns.  Sounds good.

PR #285, opened 2024-07-30
MyDataHelps AIAssistant

vladtar:
MyDataHelps Assistant tools definitions, assistant + UI component.

Probably still a little rough but good enough for people to start taking a look at it.

The component can be used with:

<table>
<tr>
<td>

```js
import { MyDataHelpsAssistantTools } from '@careevolution/mydatahelps-ui';

let tools = [
  new MyDataHelpsAssistantTools.GetAllDailyDataTypesTool()
];

<MyDataHelpsAssistant debug={true} additionalInstructions='The password is 7.' tools={tools} appendTools={false} />
```
</td>
<td>

</td>
</tr>
</table>

## Security

The open AI credentials are attached via a proxy lambda.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

image

image

## Summary by CodeRabbit

- **New Features**
	- Introduced the MyDataHelps Assistant component, enabling users to interact with an AI assistant through a chat interface.
	- Added tools for managing participant data and querying health information, enhancing user experience and functionality.
	- Updated dependencies to include new libraries for advanced language processing and validation.
	- Introduced a Chat component for real-time messaging, improving UI interactivity.
	- Added styling for the chat interface to enhance visual presentation and user experience.

- **Bug Fixes**
	- Adjusted bundle settings to improve performance and reduce output size.

- **Documentation**
	- Enhanced module exports for easier access to the MyDataHelpsAssistant and Chat components.


src/helpers/assistant/tools.ts
```diff
@@ -0,0 +1,231 @@
...
+export class QueryDailyDataTool extends StructuredTool {
+  schema = z.object({
+    type: z.string().describe("The type of daily data to query."),
+    startDate: z.string().describe("The start of the date range for the query. This is a datetime in the participant's local timezone."),
+    endDate: z.string().describe("The end of the date range for the query. This is a datetime in the participant's local timezone.")
+  });
+
+  name = "queryDailyData";
+
+  description = "Query daily data for a participant.";
```
chrisnowak, line 212: I suspect we may want to hint this a little better; I'll have to play with it a bit more to see what its like when I don't have the V2 data types enabled, but in many cases I suspect we want the assistant to be using this rather than the V2 endpoints...because the v2 endpoints are only available in certain projects and are more "raw".

How about getting rid of the "GetAllDailyDataTypesTool" and instead embedding all the daily data types in the hints for the tool so it knows to look here for that data?  (I might not fully be understanding how it's working though or how it decides what tools to use)

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { StreamEvent } from '@langchain/core/tracers/log_stream';
+import { AIMessageChunk } from '@langchain/core/messages';
+import { StructuredTool } from '@langchain/core/tools';
+import Markdown from 'react-markdown';
+import { MyDataHelpsAssistant } from '../../../helpers/assistant/assistant';
+
+import '@fortawesome/fontawesome-svg-core/styles.css';
+import "./MyDataHelpsAssistant.css";
+
+export interface MyDataHelpsAssistantProps {
```
chrisnowak, line 18: This is MyDataHelpsUI so the MyDataHelps prefix is redundant; how about "AIAssistant" for the name of this thing

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.tsx
```diff
@@ -0,0 +1,161 @@
...
+    const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        setCurrentUserMessage(event.target.value);
+    }
+
+    const handleExpandClick = async () => {
+        setCollapsed(false);
+    }
+
+    return <>
+        {collapsed && <div className="mdh-assistant-collapsed" onClick={handleExpandClick}><FontAwesomeSvgIcon icon={faFlask} size="lg" /></div>}
```
chrisnowak, line 116: We're mixing a few things into this component; it's nice that it can just be a "drop in" but it's not great to be mixing the presentation logic into the thing itself; it envisions one specific way that this would be invoked via the icon on the bottom right.  Someone might want a different icon on the bottom right or that might not play well with their UI etc.  It's generally inconsistent with how views are presented in MDH; for something like this on mobile one would want to present a full screen chat window anyways presented via MyDataHelps.openApplication.  Generally we want to keep things like buttons here pretty generic

I would remove the button for invoking this entirely; this widget should JUST be the chat window itself.  

Later when we go to incorporate this in view builder, the chat window itself would be the component that would get dropped in.  This would allow for a lot of flexibility:
- Users could invoke the chat window from any button in their view; either modally or as a popup
- Or the chat window could just be a tab in their app

If we want to separately create a "Floating button" or "sticky button" component that could also be added to view builder, we can do that as well, but my 2c is that it is not necessary for this PR.

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.tsx
```diff
@@ -0,0 +1,161 @@
...
+        {!collapsed && <div className="mdh-assistant">
+            <div className="mdh-assistant-header">
+                <div>
+                    <FontAwesomeSvgIcon icon={faFlask} />
+                    MyDataHelps Assistant
```
chrisnowak, line 121: I'd remove the header as well per my other comment

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
+:root {
```
chrisnowak, line 1: Would remove all these as we have CSS vars for various colors already in MyDataHelpsUI: https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/helpers/globalCss.ts < please use those

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+    display: flex;
+    box-shadow: var(--mdhui-box-shadow-1);
+    border-radius: 200px;
+    align-items: center;
+    justify-content: center;
+    color: #fff;
+    cursor: pointer;
+}
+
+.mdh-assistant {
```
chrisnowak, line 59: All classes should be prefixed with "mdhui-" not "mdh-"; would nest everything under the top level style (e.g. .mdhui-assistant .mdhui-assistant-header)

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+
+.mdh-assistant-header {
+    display: flex;
+    align-items: center;
+    height: 44px;
+    padding: 0 12px;
+    justify-content: space-between;
+    border-bottom: var(--border);
+    font-weight: 600;
+    color: var(--ce-grey-55);
```
chrisnowak, line 83: Should be text color var to use here

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+    color: var(--ce-grey-80);
+}
+
+/* form elements */
+.mdh-assistant input,
+.mdh-assistant button {
+    min-height: var(--touch);
+    min-width: var(--touch);
+    border: 0;
+    border-radius: var(--border-radius);
```
chrisnowak, line 117: Should be a default border radius in globalCass

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+    padding: 5px 50px 5px 15px;
+    font-size: 1rem;
+    color: var(--ce-grey-80);
+}
+
+.mdh-assistant button {
+    background: transparent;
+    border: 2px solid transparent;
+    cursor: pointer;
+    color: var(--mdh-blue-55);
```
chrisnowak, line 133: Should probably be the mdhui-color-primary

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+    width: 100%;
+    background: #fff;
+    border-top: var(--border);
+}
+
+/* other */
+.mdh-assistant input {
+    padding: 15px 35px 15px 15px;
+    box-sizing: border-box;
+    margin: 2px 3px;
```
chrisnowak, line 156: Would remove margin, set "outline-offset" to -1px

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+    width: calc(100% - 6px);
```
chrisnowak, line 157: I don't think this width is necessary?

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+    display: flex;
+    align-items: center;
+    width: 100%;
+    background: #fff;
+    border-top: var(--border);
+}
+
+/* other */
+.mdh-assistant input {
+    padding: 15px 35px 15px 15px;
```
chrisnowak, line 154: Padding should be 16px based like the rest of MDHUI; so ```16px 32px 16px 16px```

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+/* form elements */
+.mdh-assistant input,
+.mdh-assistant button {
+    min-height: var(--touch);
+    min-width: var(--touch);
+    border: 0;
+    border-radius: var(--border-radius);
+}
+
+.mdh-assistant input {
```
chrisnowak, line 120: There is another .mdh-assistant input CSS selector below this should be combined with

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+    flex: 1;
+    min-height: calc(var(--touch) + 4px);
```
chrisnowak, line 122: Would remove min height/width; the padding I suggested is good enough

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+    min-width: calc(var(--touch) + 4px);
+    padding: 5px 50px 5px 15px;
+    font-size: 1rem;
```
chrisnowak, line 125: Remove, would default to 1em

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+    padding: 15px 35px 15px 15px;
+    box-sizing: border-box;
+    margin: 2px 3px;
+    width: calc(100% - 6px);
+}
+
+.mdh-assistant .ai-message,
+.mdh-assistant .user-message,
+.mdh-assistant #loading {
+    padding: 4px 8px 4px 20px;
```
chrisnowak, line 163: Let's do 8px 8px 8px 24px

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+.mdh-assistant .user-message {
+    background: var(--ce-grey-3);
+    min-height: 27px;
+}
+
+.mdh-assistant .ai-message .svg-inline--fa,
+.mdh-assistant .user-message .svg-inline--fa,
+.mdh-assistant #loading .svg-inline--fa {
+    color: var(--ce-grey-40);
+    position: absolute;
```
chrisnowak, line 177: I would display:flex this with the message and not position it absolutely actually

rollup.config.js
```diff
@@ -25,12 +25,12 @@ export default [
 			{
 				file: packageJson.main,
 				format: "cjs",
-				sourcemap: true,
+				sourcemap: false
```
chrisnowak, line 28: We should include source maps, why'd you turn them off?

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
chrisnowak: Oh or maybe since it's an input it needs to be explicitly set to 1em

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
chrisnowak: Well really - I think what makes sense is to have this align with the left edge of the gear:
Screenshot 2024-08-05 at 4 18 09 PM

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+    height: 80vh;
+}
+
+/* typography */
+.mdh-assistant p,
+.mdh-assistant ul {
+    margin: .25rem 0;
+    word-break: break-word;
+    line-height: 1.4;
+    font-size: 15px;
```
chrisnowak, line 107: Remove font size override

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.css
```diff
@@ -0,0 +1,190 @@
...
+    overflow-x: hidden;
+    height: 80vh;
+}
+
+/* typography */
+.mdh-assistant p,
+.mdh-assistant ul {
+    margin: .25rem 0;
+    word-break: break-word;
+    line-height: 1.4;
```
chrisnowak, line 106: Remove line-height override, MDH should default to 1.3

package.json
```diff
@@ -63,17 +63,22 @@
...
+    "@langchain/openai": "^0.2.5",
     "@types/lodash": "^4.14.195",
     "date-fns": "^2.28.0",
     "html-react-parser": "^5.0.7",
     "lodash": "^4.17.21",
     "markdown-it": "^13.0.1",
     "react-feather": "^2.0.9",
     "react-fontawesome-svg-icon": "^1.1.2",
-    "recharts": "^2.9.0"
+    "react-markdown": "^9.0.1",
```
chrisnowak, line 76: We already have a markdown component we're using here - I would see if you can use "markdown-it" as we ideally don't want to introduce a redundant dependency

src/components/container/MyDataHelpsAssistant/MyDataHelpsAssistant.tsx
```diff
@@ -0,0 +1,161 @@
...
+    appendTools?: boolean;
+}
+
+export interface MyDataHelpsAssistantMessage {
+    type: string;
+    content: string;
+    runId?: string;
+}
+
+export default function (props: MyDataHelpsAssistantProps) {
```
chrisnowak, line 33: As I look at this I think it'd be ideal to extract a generic generic presentational "mdhui-chat" component; such that the AI bits aren't tied up with the assistant bits.

So here's what I would do:
- Create a "Chat" component under presentational
- As props it would have:
  - messages; each message would have an icon
  - an "onMessageSent" function which captures the user's next message they send

Then the "AIAssistant" container component would just use this "Chat" component to render the various messages it's passing back and forth etc...and could be used for other use cases later

rollup.config.js
vladtar: I was playing with the bundle size when I was trying to include the giant mermaid library, but also sourcemaps weren't working for me and still aren't after I turned them back on. Might need to spend a sec trying to figure out why. I suspect something happened when I updated the various rollup packages maybe?

vladtar:
Should be ready for another look.

I ended up extracting a `Chat` and a `Popup` presentational components.

In my app I now use:

```
<Button onClick={() => setIsCollapsed(false)}>Open AI Assistant</Button>
      {!isCollapsed && <Popup onCollapse={() => setIsCollapsed(true)} title='MyDataHelps Assistant'>
        <AIAssistant debug={true} additionalInstructions='The password is 7.' />
      </Popup>}
``` 
slightly larger than before but not too bad and definitely way more composable.

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,75 @@
+.mdhui-chat {
+    background: var(--mdhui-background-color-0);
+    font-size: 14px;
```
chrisnowak, line 3: use em for font size, e.g. .88em

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,75 @@
...
+}
+
+.mdhui-chat .mdhui-chat-log {
+    overflow-y: auto;
+    overflow-x: hidden;
+    min-height: 20vh;
```
chrisnowak, line 9: I would remove the min-height, not much point?  Wondering how this will work on mobile and if this component should just stretch to fill the remainder of the screen and if the chat box should just be sticky on the bottom

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,75 @@
...
+}
+
+.mdhui-chat p,
```
chrisnowak, line 12: I would remove these styles unless there's a strong need to override them; e.g. just go with the default styles

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,75 @@
...
+
+.mdhui-chat input,
+.mdhui-chat button {
+    min-height: var(--mdhui-touch);
+    min-width: var(--mdhui-touch);
+    border: 0;
+    border-radius: var(--mdhui-card-border-radius);
+}
+
+.mdhui-chat .send-button {
```
chrisnowak, line 27: Should be .mdhui-chat-send-button

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,75 @@
...
+    word-break: break-word;
+    color: var(--mdhui-text-color-1);
+}
+
+.mdhui-chat input,
+.mdhui-chat button {
+    min-height: var(--mdhui-touch);
+    min-width: var(--mdhui-touch);
+    border: 0;
+    border-radius: var(--mdhui-card-border-radius);
```
chrisnowak, line 24: No need for border radius here it seems?

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,75 @@
...
+    font-size: 1em;
+    color: var(--mdhui-text-color-1);
+    padding: var(--mdhui-padding-md) var(--mdhui-touch) var(--mdhui-padding-md) var(--mdhui-padding-md);
+    box-sizing: border-box;
+    outline-offset: -1px;
+}
+
+.mdhui-chat .input-group {
+    position: relative;
+    display: flex;
```
chrisnowak, line 45: Confused, this is display:flex but the button is positioned absolute?  Seems like you should just use flex to lay out the input / button?

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,75 @@
...
+.mdhui-chat input {
+    flex: 1;
+    font-size: 1em;
+    color: var(--mdhui-text-color-1);
+    padding: var(--mdhui-padding-md) var(--mdhui-touch) var(--mdhui-padding-md) var(--mdhui-padding-md);
+    box-sizing: border-box;
+    outline-offset: -1px;
+}
+
+.mdhui-chat .input-group {
```
chrisnowak, line 43: Should be .mdhui-chat-input-group

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,75 @@
...
+    position: relative;
+    display: flex;
+    align-items: center;
+    width: 100%;
+    background: #fff;
+    border-top: 1px solid var(--mdhui-border-color-1);
+}
+
+.mdhui-chat .mdhui-chat-message,
+.mdhui-chat .message-loading {
```
chrisnowak, line 53: .mdhui-chat-message-loading

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,75 @@
...
+    padding: var(--mdhui-padding-xs) var(--mdhui-padding-xs) var(--mdhui-padding-xs) var(--mdhui-padding-lg);
```
chrisnowak, line 54: I would use mdhui-padding-md instead of mdhui-padding-lg here

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,75 @@
...
+    position: relative;
+    border-bottom: 1px solid var(--mdhui-border-color-1);
+}
+
+.mdhui-chat .mdhui-chat-message.sent {
```
chrisnowak, line 59: .mdhui-chat-message-sent

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,75 @@
...
+    background: var(--mdhui-background-color-1);
+}
+
+.mdhui-chat .mdhui-chat-message .svg-inline--fa,
+.mdhui-chat .message-loading .svg-inline--fa {
+    color: var(--mdhui-text-color-3);
+    margin-right: .25rem;
```
chrisnowak, line 66: I would flex in the icon or position it absolutely so it's not just inline with the text, e.g.
Screenshot 2024-08-15 at 2 56 55 PM

src/components/presentational/Chat/Chat.tsx
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    return (
+        <div className="mdhui-chat">
+            <div className="mdhui-chat-log" ref={logRef}>
+                {props.messages.map((message, index) => (
+                    <div key={index} className={"mdhui-chat-message " + message.type}>
+                        {message.icon}{parse(md.render(message.content), {
+                            transform: (reactNode, domNode, index) => {
+                                if (domNode.nodeType === 1 && domNode.name === "p" && index === 0) {
```
chrisnowak, line 52: What's the point of this?

src/components/presentational/Chat/Chat.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export interface ChatMessage {
+    icon: React.JSX.Element;
+    content: string;
+    type: ChatMessageType;
+}
+
+export default function (props: ChatProps) {
+
+    let md = new MarkdownIt();
```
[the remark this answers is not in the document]
chrisnowak: This seems like a good suggestion

src/components/presentational/Popup/Popup.tsx
```diff
@@ -0,0 +1,26 @@
...
+    onCollapse: () => void;
+    title?: string;
+    titleIcon?: React.JSX.Element;
+}
+
+export default function (props: PopupProps) {
+
+    return <div className="mdhui-popup">
+        <div className="mdhui-popup-header">
+            <div>
```
chrisnowak, line 18: I would use an Action component here which supports a left/right icon

src/components/presentational/Popup/Popup.tsx
```diff
@@ -0,0 +1,26 @@
...
+import './Popup.css';
+
+export interface PopupProps {
+    children?: React.ReactNode;
+    onCollapse: () => void;
+    title?: string;
+    titleIcon?: React.JSX.Element;
+}
+
+export default function (props: PopupProps) {
```
chrisnowak, line 14: How is this popup thing supposed to work...I don't see a way to actually expand/collapse it?

src/components/container/AIAssistant/AIAssistant.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+export default function (props: AIAssistantProps) {
+
+    const [messages, setMessages] = useState<AIAssistantMessage[]>([]);
+    const [loading, setLoading] = useState("");
+
+    const assistantRef = useRef<MyDataHelpsAssistant>();
+
+    useEffect(() => {
+        if (assistantRef.current === undefined) {
```
chrisnowak, line 39: Out of curiosity why ref and not state?

src/components/presentational/Chat/Chat.css
chrisnowak: Actually, I would just use "Action" component for the messages.  That component already supports icons on the left etc. and is visually consistent - and, despite it's name, if you do not supply an onclick handler it will not be clickable.  Then you do not need to style the messages at all - you can just put the message itself in the subtitle or children

src/helpers/assistant/tools.ts
chrisnowak: Thoughts on this?  Again like this is going to be the best source of daily data that will match graphs in most cases, so I think we should give it all the data types this can supply

src/components/presentational/Chat/Chat.tsx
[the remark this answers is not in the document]
vladtar: Done.

src/components/presentational/Chat/Chat.css
vladtar: This needs some sort of min-height otherwise the chat window is completely empty at the beginning and it's not really clear that messages will start showing up in it. I thought the 20vh should work fine for both desktop and mobile?

Chat box sticky on the bottom is basically what the Popup is for.

src/components/presentational/Chat/Chat.css
vladtar: Looks fine both with it and without? Are our outlines usually without rounded corners?

src/components/presentational/Chat/Chat.css
vladtar: Removed the text color as that wasn't really needed, the other 2 do change behavior? Less top and bottom margin and different word break behavior than default (normal)?

src/components/presentational/Chat/Chat.css
vladtar: Argh the first time it was positioned absolutely and you said `I would display:flex this with the message and not position it absolutely actually` which I think I also misinterpreted as meaning displaying the icon inline. Will try with flex.

src/components/presentational/Chat/Chat.tsx
vladtar: Crazy ninja code to inline the icon. Will probably end up removing it.

src/components/presentational/Popup/Popup.tsx
vladtar: I don't think `Action` is the right component to use here. If we really wanted to take reusability all the way we can probably write an Icon component and it would have the option to have some text next to the icon as well. But this is small enough and simple enough I would prefer to just leave it as is icon with some text next to it doesn't need to be a component.

src/components/presentational/Popup/Popup.tsx
vladtar: Since the parent sort of controls whether this starts up collapsed or expanded, so basically has to pass in a `prop`, and then the component itself also controls that state in case someone collapses it (`state`) I thought it would be slightly cleaner if the parent kept the `isCollapsed` state. I showed how I'm using it in my app in the top level comment:

```
const [isCollapsed, setIsCollapsed] = useState(true);
// ...
<Button onClick={() => setIsCollapsed(false)}>Open AI Assistant</Button>
      {!isCollapsed && <Popup onCollapse={() => setIsCollapsed(true)} title='MyDataHelps Assistant'>
        <AIAssistant debug={true} additionalInstructions='The password is 7.' />
      </Popup>}
```

src/components/container/AIAssistant/AIAssistant.tsx
vladtar: Conceptually seems slightly different. https://react.dev/reference/react/useRef this is close to the `interval` and `VideoPlayer` examples in the documentation.

src/components/presentational/Popup/Popup.tsx
chrisnowak: Why not?  This seems to perfectly match action because it has the same layout as action?  Since it has a left/right icon which action supports...and again Action has all the correct styles associated with it etc.

src/components/presentational/Chat/Chat.css
chrisnowak: The outline should match the container which is a rectangle, the rounded outline only appears when you focus which feels awkward

src/components/presentational/Chat/Chat.css
chrisnowak: Hmm yeah, this might be worth chatting a bit further about how we would, say, incorporate this as a view builder component; what is the recommended UI paradigm etc.

I was not thinking that we would still necessarily even have the popup component; instead this AI Assistant could be shown on it's own entire view...not necessarily be floating on top of the current view; but instead using MyDataHelps.openApplication to open another view that is the AI Assistant.  (usually that is how you see "chat" applications work for instance)

In which case you'd want the text box anchored to the bottom while the user can scroll back through messages etc.

May be worth chatting live about this at some point if there's a good time though, (this might require some deep testing on mobile to deal with virtual keyboard stuff)

src/helpers/assistant/tools.ts
vladtar: For `queryDeviceDataV2` I included in the description:
```
Before using this tool call the getDeviceDataV2AllDataTypes tool to see which data types are available. If a particular data type is not available use the queryDailyData tool instead.
```

src/components/container/AIAssistant/AIAssistant.tsx
```diff
@@ -0,0 +1,120 @@
...
+            else {
+                return [...prevMessages, { type: 'ai', content: message, runId: runId }];
+            }
+        });
+    }
+
+    return <>
+        {messages && <Chat innerRef={props.innerRef} messages={messages.map((msg) => {
+            return {
+                icon: msg.type === "ai" ? <FontAwesomeSvgIcon icon={faFlask} width={16} /> : undefined,
```
vladtar, line 114: Does the flask icon work for now? I kind of liked the flask with gear icon https://fontawesome.com/icons/flask-gear?s=solid but it's one of the paid fontawesome icons.

I know we discussed getting rid of icons completely but I have to admit things looked strange without the ai icon at all. It becomes a little unclear when the ai started a new response etc since it can talk in paragraphs.

src/components/presentational/Chat/Chat.css
vladtar: If I don't do it like this the outline won't cover the send icon.

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,100 @@
...
+    background: var(--mdhui-background-color-1);
+    max-width: 70%;
+    padding: var(--mdhui-padding-xs) var(--mdhui-padding-sm);
+    border-radius: var(--mdhui-card-border-radius);
+}
+
+.mdhui-chat .mdhui-chat-message .svg-inline--fa,
+.mdhui-chat .mdhui-chat-message-loading .svg-inline--fa {
+    color: var(--mdhui-text-color-3);
+    margin-top: .35rem;
```
chrisnowak, line 90: I would definitely not use rems for this.  You ultimately want the text to be 16, 24, or 32px from the left edge of the screen.

My 2c is to drop the icon at this point?  The little flask icon I don't think works particularly great and it's kinda just jammed in there really close to the text.  And there is decent differentiation between the user messages and the AI messages due to the new border/background of the user messages

Either that or I would give the icon a little more room to breathe if we think we need it

Good rule of thumb is that no icons/text should be in the left or right 16px of the screen though

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,100 @@
...
+    display: flex;
+    flex-direction: row;
+    align-items: flex-start;
+}
+
+.mdhui-chat .mdhui-chat-sent-message-row {
+    display: flex;
+    flex-direction: column;
+    align-items: flex-end;
+    padding-right: var(--mdhui-padding-xs);
```
chrisnowak, line 77: --mdhui-padding-md for this

src/components/container/AIAssistant/AIAssistant.tsx
chrisnowak: Ah I see - you could just go with a dot/bullet maybe if the goal is just to ensure you understand where it started - or an arrow or something

src/components/presentational/Chat/Chat.tsx
```diff
@@ -0,0 +1,89 @@
...
+                        type="text"
+                        value={currentUserMessage}
+                        onChange={(e) => setCurrentUserMessage(e.target.value)}
+                        onKeyDown={(e) => {
+                            if (e.key === 'Enter') {
+                                sendMessage();
+                            }
+                        }}
+                    />
+                    <UnstyledButton onClick={sendMessage} className="mdhui-chat-send-button">
```
chrisnowak, line 82: Would disable the send and also make sure no event fires on enter when there's no message entered (blank string)

src/components/container/AIAssistant/AIAssistant.tsx
chrisnowak: I don't really love the flask because we use that for labs etc. and it doesn't necessarily have any significance to the consumer with an AI chatbot...like it feels like a flag that this is "experimental" perhaps but other than that i'm not sure it makes sense

src/components/container/AIAssistant/AIAssistant.tsx
```diff
@@ -0,0 +1,121 @@
...
+import '@fortawesome/fontawesome-svg-core/styles.css';
+
+export interface AIAssistantProps {
+    innerRef?: React.Ref<HTMLDivElement>;
+    previewState?: "default";
+    debug: boolean;
+    additionalInstructions?: string;
+    tools?: StructuredTool[];
+    appendTools?: boolean;
+    baseUrl?: string;
```
chrisnowak, line 21: Can we just infer this from the base url of MyDataHelps.js?  Is there a reason to override it?

src/components/container/AIAssistant/AIAssistant.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { AIMessageChunk } from '@langchain/core/messages';
+import { StructuredTool } from '@langchain/core/tools';
+import MyDataHelps from '@careevolution/mydatahelps-js';
+
+import { MyDataHelpsAssistant } from '../../../helpers/assistant/assistant';
+import Chat from '../../presentational/Chat';
+
+import '@fortawesome/fontawesome-svg-core/styles.css';
+
+export interface AIAssistantProps {
```
chrisnowak, line 14: I think we should have an initial message string prop that populates in the chat as the first message, so users see something when they land on the page...e.g. "Ask questions about your lab reports" or something

src/components/container/AIAssistant/AIAssistant.tsx
```diff
@@ -0,0 +1,121 @@
...
+                    }
+                    else if (msg.tool_call_chunks[0].name) {
+                        addMessageChunk(streamEvent.run_id, msg.tool_call_chunks[0].name + " ");
+                    }
+                }
+            }
+
+            if (kind === "tool") {
+                if (type === "start") {
+                    setLoading(`Querying your data...`);
```
chrisnowak, line 72: This should be a localized string

src/components/presentational/Layout/Layout.css
```diff
@@ -1,5 +1,11 @@
 ﻿.mdhui-layout {
-    padding-bottom: env(safe-area-inset-bottom);
```
chrisnowak, line 2: Hmm is removing this safe?  IIRC this was added for a reason on iOS

chrisnowak: (Though it seems fine in my testing, so IDK)

src/components/presentational/Layout/Layout.tsx
```diff
@@ -13,6 +13,7 @@ export interface LayoutProps {
 	className?: string;
 	noGlobalStyles?: boolean;
 	colorScheme?: "light" | "dark" | "auto";
+	isFlex?: boolean;
```
chrisnowak, line 16: Would just do "flex" not "isFlex"

src/helpers/assistant/assistant.ts
```diff
@@ -0,0 +1,169 @@
...
+    QueryNotificationsTool,
+    QuerySurveyAnswersTool,
+    QueryDailyDataTool,
+    GetAllDailyDataTypesTool,
+    GetEhrNewsFeedPageTool,
+    GetDeviceDataV2AllDataTypesTool
+} from "./tools";
+
+
+export interface IAssistantState {
```
chrisnowak, line 27: Interfaces in typescript shouldn't start with "I", would call this AIAssistantState

src/helpers/assistant/assistant.ts
```diff
@@ -0,0 +1,169 @@
...
+    messages: BaseMessage[];
+    participantInfo: string;
+}
+
+export class MyDataHelpsAssistant {
```
chrisnowak, line 32: Would call this MyDataHelpsAIAssistant and rename the file / folder to be consistent with "AI Assistant"

src/helpers/assistant/assistant.ts
```diff
@@ -0,0 +1,169 @@
...
+            just objectively share information.
+            
+            You can encourage the user to ask more questions and even suggest additional follow-up questions that might be relevant.
+
+            If the user asks for some data, and you query it with a particular tool, but the returned data does not sufficiently
+            answer the user's question, for example, if the user asks for their last 3 LDL values, and you query the getEhrNewsFeedPage
+            tool and it returns only the last 2 LDL values and a nextPageID, then query the same tool again while passing the nextPageID as the
+            pageID parameter to fetch additional data. Continue this process until you have all the data that the user has asked for, up to 5 iterations.
+
+			User information: {participantInfo}
```
chrisnowak, line 100: I would also consider embedding the projectInfo here (the stuff you get via MyDataHelps.getProjectInfo() or whatever the method is) - as that could be quite useful for things like the support contact number, etc.

chrisnowak: (And is fast and takes no parameters, so doesn't make sense to be a tool)

chrisnowak (approved):
Overall looks great.  I didn't go too deep on the tools themselves since I think those will end up having the most iteration anyways.  Would be very interested to add a tool for symptom shark data to see how that works!

src/components/presentational/Layout/Layout.css
vladtar: If you look at `Layout.tsx` there's `let paddingBottom = props.flex ? "0" : "env(safe-area-inset-bottom)";` so to clarify I did not completely remove this, just made it conditional on the `Layout` not being flex.

src/helpers/assistant/assistant.ts
vladtar: Good idea. Added.

In some future iteration I wonder if we should query upfront `getAllDailyDataTypes` and `getDeviceDataV2AllDataTypes` as well.

src/components/presentational/Chat/Chat.css
vladtar: This only affects the position of the icon with respect to the text on the right.

I added more left and right padding to the assistant messages. Replaced the icon with lightbulb as discussed.

src/components/presentational/Layout/Layout.css
chrisnowak: got it

src/helpers/assistant/assistant.ts
chrisnowak: Likely so, those are very fast calls so there is no downside to doing so

src/components/container/AIAssistant/AIAssistant.tsx
vladtar: I would like to leave this as an upcoming improvement to offer a few rotating suggestions just like chat gpt for interesting prompts:

image

chrisnowak: Interesting.  Yeah it's not super critical since we can throw a text block above it

src/components/presentational/Chat/Chat.css
```diff
@@ -0,0 +1,100 @@
...
+    min-height: 150px;
+}
+
+.mdhui-chat .mdhui-chat-log {
+    overflow: hidden;
+    flex: 1 1 0%;
+}
+
+.mdhui-chat .mdhui-chat-messages {
+    overflow-y: auto;
```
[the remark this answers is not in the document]
vladtar: After some tweaking I was able to get this to work. The only current issue I see with it is now when displaying then hiding the loading message the rest of the messages get moved up and down for a bit which can get a bit jerky if the loading message is displayed only for a very short time.

PR #287, opened 2024-07-31
Add glucose and meal log components.

greinard:
## Overview

This PR adds a number of components related to the display of blood glucose and meals.  Together, they can be used in a view like the one pictured below.

New functionality includes the graphing of continuous glucose data with a scaled steps overlay and self-reported meal indicators.  Meals can be selected from the meal log to view a 2-hour window of CGM data immediately following the selected meal.  Daily steps and sleep counts are displayed along with some glucose stats.  Self reported stress levels can also be recorded for each day.

&nbsp;&nbsp;

&nbsp;

## Security

- No new security risks.  These are just display components for device acquired continuous glucose data and self reported meal and stress levels.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Summary by CodeRabbit

- **New Features**
    - Added support for UUID generation, improving functionality for unique identifier requirements.
    - Launched a new `MealEditor` component for managing meal entries, featuring 'add' and 'edit' modes, along with improved error handling and user feedback.
  
- **Chores**
    - Updated various project dependencies for better compatibility and performance, including enhancements to `@careevolution/mydatahelps-js`.


src/helpers/glucose-and-meals/sleep.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    let settings = await MyDataHelps.getDataCollectionSettings();
+    if (settings.fitbitEnabled) {
+        providers.push(fitbitSleepMinutesProvider(date));
+    }
+    if (settings.queryableDeviceDataTypes.find(s => s.namespace == 'AppleHealth' && s.type == 'SleepAnalysisInterval')) {
+        providers.push(appleHealthSleepMinutesProvider(date));
+    }
+
+    return providers.length > 0 ? await getMaxValue(providers) : undefined;
```
ajmenca, line 33: It looks like this is the only logic thing that is why you're making your own provider here instead of using the provider in `combined-sleep.ts`?  Is there a reason that you made a new one instead of adjusting that provider so that it can have some sort of adjustable "accumulation" strategy?  

I ask partially because I notice that the new provider you're adding here doesn't touch Garmin data, and that might be unexpected.

src/helpers/glucose-and-meals/steps.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    let settings = await MyDataHelps.getDataCollectionSettings();
+    if (settings.fitbitEnabled) {
+        providers.push(fitbitHalfHourStepsDataProvider(date));
+    }
+    if (settings.queryableDeviceDataTypes.find(s => s.namespace == 'AppleHealth' && s.type == 'HalfHourSteps')) {
+        providers.push(appleHealthHalfHourStepsDataProvider(date));
+    }
+
+    return providers.length > 0 ? await getMaxValueReadings(providers) : [];
```
ajmenca, line 58: Similar question here as above regarding not looking at Garmin data?

src/helpers/glucose-and-meals/sleep.ts
greinard: The main reason was that this uses a hybrid of V2 (for Fitbit) and V1 (for Apple Health).

The existing `combined-sleep.ts` uses V1 for both.  I haven't had a chance to examine how the Apple Health sleep data looks in V2, but I could probably take a look at that to create a V2 Apple Health Sleep query and then a new generic `combined-sleep-v2.ts`.

Good catch on Garmin.  Yes, I need to add that back, too.  This query evolved a little bit, since it was originally going to return half hour bucketed sleep values (like steps), and I was going to punt on Garmin for now, but then we reverted back to just a single daily value for the time being.  I'll get Garmin (using V1) added back.

Once a generic `combined-sleep-v2.ts` exists, this file can probably go away.  The `getMaxValue` accumulation strategy is already in use by the V1 combined sleep query, I just refactored it into a more explicitly named helper function.

src/helpers/glucose-and-meals/steps.ts
greinard: I think we do want to support Garmin for this too, I just need to work out how best to do that.  For Fitbit, we are using the V2 API and aggregation.  For Apple Health, we use the special statistics query rule configuration.  I think support Garmin data was just recently added to the V2 real time API, so I just need to look at that a bit and add a "half hour steps" computation.

src/helpers/glucose-and-meals/sleep.ts
greinard: I've added support for V1 Garmin sleep minutes.

I'll take a look at what it would take to create V2 versions of the Apple Health and Garmin sleep queries (and a `combined-sleep-v2.ts`).

src/components/container/GlucoseChart/GlucoseChart.tsx
```diff
@@ -0,0 +1,224 @@
...
+                        width: 24,
+                        domain: [0, 240],
+                        ticks: [40, 80, 120, 160, 200, 240]
+                    }
+                }}
+            >
+                {avgGlucose &&
+                    <ReferenceLine
+                        y={avgGlucose}
+                        stroke="var(--mdhui-color-primary)"
```
reneefromhold, line 186: To make this easier to digest on first sight, consider making the icons for blood glucose (range and avg) different colors, and coloring the reference line for avg glucose to match the icon color. 
Atm, the reference line for avg glucose is colored closer to sleep icon.

src/helpers/glucose-and-meals/sleep.ts
greinard: I went ahead and created the V2 Apple Health and Garmin sleep queries, as well as a `combined-sleep-v2.ts`, which this file now references.

src/helpers/glucose-and-meals/steps.ts
greinard: I've added the Garmin half hour steps implementation.

src/components/presentational/DiscreteScale/DiscreteScale.tsx
```diff
@@ -1,15 +1,15 @@
 import React, { CSSProperties, useContext } from 'react';
-import { ColorDefinition, resolveColor } from '../../../helpers';
+import { ColorDefinition, language, resolveColor } from '../../../helpers';
```
reneefromhold, line 2: Why no index.js file for this component?

src/components/view/MealEditorView/MealEditorView.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
```
reneefromhold, line 1: Can I see a demo of how this works when it is dismissed? Trying to imagine how it would be configured in vb.
Appearance wise, It looks identical to MealEditor

reneefromhold (commented):
@greinard Left some comments. No code issues, mostly questions on how some components and changes to the data providers. If you have some time on Monday, can we chat?

src/helpers/daily-data-providers/garmin-sleep-v2.ts
```diff
@@ -0,0 +1,19 @@
+﻿import { querySleepMinutesV2 } from './common-sleep-v2';
+
+// TODO: Remove namespace "as any" once MDH-JS gets updated.
```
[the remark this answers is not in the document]
greinard: Done.

src/components/view/MealEditorView/MealEditorView.tsx
greinard: Yes, for sure, I can show you a demo.  The modal dialog just closes, though.

This view component is lightweight wrapper around the editor component, so it doesn't add any functionality other than defining the actions to take when things like save and cancel occur.  In view builder, these handlers are configured as action definitions on the editor component node, so the view component is not necessary.

I'm planning to remove this view and the `GlucoseChartView` now that view builder support has been implemented.  I just left them here for the time being so that folks could see how the new components are meant to fit together.

src/components/container/GlucoseChart/GlucoseChart.tsx
greinard: I think we'll want to keep the two blood icons the same color since they both refer to the same data.

The average line is currently using the primary color (which is configurable), but I will add a configuration option for that in case someone wants to override the primary color with something else.

greinard: I've added an optional property on the `GlucoseChart` for the average glucose reference line color.

greinard:
I have resolved the merge conflicts from `main`, reset the version number, and removed the unnecessary view components.

Please let me know if there are any further concerns.  I would like to get this merged in by tomorrow morning, if possible.

Thanks!

ajmenca (approved):
In general, this is all really cool.  I have a few UI feedback comments, which you can feel free to take or leave with whatever size grain of salt you'd like.  I know that there have been plenty of design discussions about this and I don't want to swoop in last minute.

- I don't think that it's immediately obvious that the bar graph behind the glucose chart is steps.  I know the idea is that the step icon is at the bottom of the graph, and it's color coded to be the same as the bar graph and the step count in the stats - but I personally found it's location confusing because it looks like its part of the Y-Axis representation (since it's where the 0 would be) but it's actually not relevant to the Y-Axis at all since that's glucose levels and not steps.  I think it would be great if there were actually a second Y-Axis on the right to indicate steps, and the step icon could be over there instead (https://recharts.org/en-US/examples/BiaxialLineChart).

- The interaction between the Meal Log and the Glucose Graph is *really cool* but I don't think it's discoverable.  The log entries don't look like buttons, and they have an actual edit button on them which looks like the only clickable thing.  I think this is a really great feature that most participants will not notice or find.  I hate giving this feedback because I don't actually have a great suggestion for making this click interaction more obvious; the only idea I really have is putting an empty circle where the checkbox would show up so that they have a sort of "radio button" vibe?  But I don't really love that idea.

- I know you're not optimizing for Desktop on any of these widgets, but they all look a litle weird with the exception of the Glucose Graph.  The Meal Editor is probably the one that's the strangest, as the super wide time input puts the clickable "clock" widget to open the native browser time editor a really far distance from the text itself.

greinard:
Thanks for all the great feedback @ajmenca.  I think we are going to proceed with the current design for the time being and see what sort of feedback we get from end users.  I'll tinker with the items you suggested and see if I can come up with some improvements in a follow up PR.

PR #293, opened 2024-08-21
Update the AsthmaRecommendedArticle component to support localization.

greinard:
## Overview

This branch updates the `AsthmaRecommendedArticle` component so that it supports localization of the recommended article titles and subtitles.

Translations were generated using ChatGPT.

I will revert the snapshot version prior to merging.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation


greinard:
I would like to merge this within the next couple of days if someone has a few minutes for a review.

Thank you!

ajmenca (commented):
Just based on code review, this looks ok; but I don't see any real easy way to test or verify this via Story, because both the article chosen relies on live data (Story shows placeholder data) and we also don't have a very easy way to switch languages via Story.  

So maybe a loaded question - but how did you verify all of these changes?

greinard:
> Just based on code review, this looks ok; but I don't see any real easy way to test or verify this via Story, because both the article chosen relies on live data (Story shows placeholder data) and we also don't have a very easy way to switch languages via Story.
> 
> So maybe a loaded question - but how did you verify all of these changes?

The conditional logic has thus far been tested using the DryRun Asthma Tool environment.  I think we'll probably test a couple of those conditions again after this update to ensure things are still working as expected, but no logical changes were intended with this PR, just the translations.  That said, I will add an item to my personal to-do list to come back and add some unit testing for this logic.

For the translations, I had just commented out a few lines in the story/component to simulate a "live" state within the storybook and then adjusted the selected language.  I didn't exhaustively test all the translations.  I just verified that the translation was taking place.  The Spanish translations themselves are undergoing a separate review within the Asthma Tool library feature (same translations are duplicated here).  However, I don't have a good way to verify the translations to the other languages short of the few spot checks I ran back through Google translate.   All were originally generated using ChatGPT.

I have reworked the component a tad and updated the storybook such that it now uses a real example for the default preview state and supports a live state (if you have an Asthma Tool participant access token).

reneefromhold (approved):

PR #298, opened 2024-09-11
Add ability to ensure sequential completion of survey tasks

felideon:
## Overview

Paves way for https://github.com/CareEvolution/MyDataHelpsViewBuilder/issues/111.

## Security

No security risks come to mind.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `sequential` property to enhance task presentation in the survey task components.
	- Added a `surveyBlocked` property to manage user interactions based on survey availability.
	- Created a new story for sequential task presentation in the storybook.

- **Bug Fixes**
	- Improved control flow to prevent user interactions when a survey task is blocked, enhancing overall user experience.

- **Style**
	- Updated cursor behavior for interactive elements to improve user interface clarity.

### DIFF
--- a/src/components/container/SurveyTaskList/SurveyTaskList.stories.tsx
+++ b/src/components/container/SurveyTaskList/SurveyTaskList.stories.tsx
@@ -22,7 +22,8 @@ Incomplete.args = {
 	status: 'incomplete',
 	title: 'Incomplete Tasks',
 	previewState: "IncompleteTasks",
-	variant: "singleCard"
+	variant: "singleCard",
+	sequential: false
 }
 
 export const Complete = Template.bind({});
@@ -40,7 +41,8 @@ Multicard.args = {
 	status: 'incomplete',
 	title: 'Incomplete Tasks',
 	previewState: "IncompleteTasks",
-	variant: "multiCard"
+	variant: "multiCard",
+	sequential: false
 }
 
 export const CustomStyle = Template.bind({});
@@ -56,5 +58,6 @@ CustomStyle.args = {
 		boxShadow:"none"
 	},
 	buttonVariant: "default",
-	buttonColor: "blue"
+	buttonColor: "blue",
+	sequential: false
 }
--- a/src/components/container/SurveyTaskList/SurveyTaskList.stories.tsx
+++ b/src/components/container/SurveyTaskList/SurveyTaskList.stories.tsx
@@ -26,23 +26,16 @@ Incomplete.args = {
 	sequential: false
 }
 
-export const Complete = Template.bind({});
-Complete.args = {
-	limit: 3,
-	status: 'complete',
-	title: 'Completed Tasks',
-	previewState: "CompleteTasks",
-	variant: "singleCard"
+export const Sequential = Template.bind();
+Sequential.args = {
+	...Incomplete.args,
+	sequential: true
 }
 
 export const Multicard = Template.bind({});
 Multicard.args = {
-	limit: 3,
-	status: 'incomplete',
-	title: 'Incomplete Tasks',
-	previewState: "IncompleteTasks",
-	variant: "multiCard",
-	sequential: false
+	...Incomplete.args,
+	variant: "multiCard"
 }
 
 export const CustomStyle = Template.bind({});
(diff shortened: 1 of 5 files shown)


felideon:
Storybook screenshots:

Screenshot 2024-09-11 at 3 09 49 PM

Screenshot 2024-09-11 at 3 11 07 PM

src/components/container/SurveyTaskList/SurveyTaskList.stories.tsx
ajmenca, line 1: There should be a new story in this file which demonstrates the new `sequential` flag.

src/components/container/SurveyTaskList/SurveyTaskList.tsx
```diff
@@ -51,13 +52,31 @@ export default function (props: SurveyTaskListProps) {
 		}
 	};
 
+	const isSurveyBlocked = (task: SurveyTask): boolean => {
+		let blocked = false;
+
+		if (tasks && props.sequential && tasks?.length > 1) {
+			if (tasks.at(0)?.id != task.id) {
```
ajmenca, line 59: This is a nitpick, but I think style-wise I wouldn't use `at` here.  Since you're always indexing `0`, it doesn't seem like there's a benefit over doing `tasks[0]?`.id` here.

ajmenca, line 59: Actually - is all of this checking necessary in practice?  This function is called on a specific `Task` and... you can't *have* a `Task` in any of this code unless `tasks` is non-null and there's at least one thing in it?

It seems like this function could just `return tasks![0].id === task.id`?

felideon: Ah, yes, thanks. Didn't know about the non-null assertion operator. Pushed a change as suggested, except the equality needed to be flipped and the sequential prop still needed to be checked.

src/components/container/SurveyTaskList/SurveyTaskList.stories.tsx
felideon: I added it as an arg to the existing stories but can add a new story where it's the default. Would you still keep the args in the other stories?

ajmenca: I would - I think both are helpful.  Having a story that specifically calls out the behavior helps, in my opinion, to highlight that the feature is available, while having the arg on the other stories helps test the combinations of features (since having *every* combination of features as a story is certainly not sustainable!).

src/components/container/SurveyTaskList/SurveyTaskList.stories.tsx
```diff
@@ -56,5 +51,15 @@ CustomStyle.args = {
 		boxShadow:"none"
 	},
 	buttonVariant: "default",
-	buttonColor: "blue"
-}
\ No newline at end of file
+	buttonColor: "blue",
+	sequential: false
```
reneefromhold, line 55: Remove setting sequential to false. Default is false

felideon: I added so it shows in those stories in the Storybook UI, which is nice to be able to play around with. Is there another way to get it to show if it's the default value? (I'm new to Storybook.)

reneefromhold: It is part of the Sequential story

reneefromhold:
@felideon The way it is implemented, a new view would be needed if there was ever a desire to complete non sequentially. Is that going to be ok?

src/components/container/SurveyTaskList/SurveyTaskList.stories.tsx
ajmenca: > It is part of the Sequential story

Yeah, but as I noted in one of my comments, having it available to toggle in *each* story helps test compatibility with all the other settings.  I don't think it's bad to have it available on all of the stories.

reneefromhold:
I connected with @felideon on my previous question about how to not make this binary. In the near future, there will be support for filtering tasks in the task list by category. Having options like that will make it easier to support a mix of non seq and sequential on the same view

reneefromhold (approved):

ajmenca (approved):

PR #309, opened 2024-09-18
Some glucose related component updates.

greinard:
## Overview

This branch contains a number of glucose related component updates.

- The `DateRangeCoordinator` was updated to support having a custom navigator within its `children`.
- The `RelativeActivityDayNavigator` was renamed/refactored into a reusable `WeeklyDayNavigator` component that can be configured with a use-case-specific data loader and day renderer.
- The `RelativeActivityDayCoordinator` was updated to use a `DateRangeCoordinator` with a custom `WeeklyDayNavigator`.
- A new `SparkRangeChart` component has been added for rendering collections of min/max/average information.
- A `GlucoseDayCoordinator` component was created that uses a `DateRangeCoordinator` with a custom `WeeklyDayNavigator` to render glucose daily data ranges above the days.
- The `GlucoseChart` stories have been updated to use a `GlucoseDayCoordinator`.
- The `GlucoseChart` was updated to use already-loaded data from the `GlucoseContext`, when available.
- The `GlucoseChart` reference line was updated to show the recent (7 day) average, when available from the `GlucoseContext`, rather than the current daily average.
- Glucose sample data is now generated using predictable random numbers.

**Glucose Chart w/ GlucoseDayCoordinator (w/ SparkRangeChart's)**

image

I will revert the snapshot version prior to merging.

## Security

- No new security risk.  This update just adjusts the rendering of data that is already available to the user.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Release Notes

- **New Features**
  - Introduced a custom navigator component for the `DateRangeCoordinator`, enhancing date navigation with interactive features.
  - Added `DateRangeTitle` for improved display of date ranges.
  - Launched the `GlucoseDayCoordinator` component to manage and visualize glucose readings over a specified date range.
  - Enhanced the `RelativeActivityDayCoordinator` with improved data loading and visualization features.

- **Improvements**
  - Enhanced the `DateRangeCoordinator` functionality with a visually engaging weekly day navigator that includes custom day rendering.
  - Improved glucose data visualization with a structured approach to display readings and averages.
  - Refined state management and rendering logic in the `RelativeActivityDayCoordinator` for better clarity and user feedback.

- **Bug Fixes**
  - Removed unused imports to simplify the codebase.

- **Documentation**
  - Expanded Storybook configurations for new components to facilitate testing and visualization.


src/components/container/GlucoseChart/GlucoseChart.previewData.ts
```diff
@@ -8,16 +8,17 @@ export interface GlucoseChartPreviewData {
     sleepMinutes: number | undefined;
 }
 
-export const previewData = (previewState: GlucoseChartPreviewState, date: Date): GlucoseChartPreviewData => {
+export const previewData = async (previewState: GlucoseChartPreviewState, date: Date): Promise<GlucoseChartPreviewData> => {
```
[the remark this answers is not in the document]
greinard: The `previewData` functions used by the MealEditor and MealCoordinator do not return promises.

src/components/container/GlucoseChart/GlucoseChart.tsx
```diff
@@ -39,14 +43,16 @@ export default function (props: GlucoseChartProps) {
...
+                setGlucose(glucose)
+                setSteps(steps);
+                setSleepMinutes(sleepMinutes);
+                setLoading(false);
+            });
             return;
         }
 
-        let glucoseReadingLoader = getGlucoseReadings(selectedDate);
+        let glucoseReadingLoader = glucoseContext?.readings ? getGlucoseReadingsFromContext() : getGlucoseReadings(selectedDate);
```
ajmenca, line 55: I think we discussed this a bit on Slack - this is an attempt to prevent multiple coordinated widgets from all requesting the same data, right?  

I'm a little confused on exactly how this ends up working because it looks like `readings` isn't defaulted to an empty array in the `GlucoseDayCoordinator`?  So it seems like on initial load, `glucoseContext?.readings` would return false even if you're inside of a valid `GlucoseContext`?  So it looks like you'd still get a double load - what am I missing?

src/components/container/GlucoseDayCoordinator/GlucoseDayCoordinator.tsx
```diff
@@ -0,0 +1,81 @@
...
+    </div>;
+}
+
+interface CustomNavigatorProps {
+    loadData: (startDate: Date, endDate: Date) => Promise<void>;
+    dayRenderer: (dayKey: string) => React.JSX.Element | null;
+    dependencies?: DependencyList;
+}
+
+function CustomNavigator(props: CustomNavigatorProps) {
```
ajmenca, line 72: Can these two names be more specific?  I had to double check that they weren't generic interfaces for the customizable navigators that were living in the wrong file (I didn't immediately notice that they weren't exported).

src/components/presentational/DateRangeCoordinator/DateRangeCoordinator.tsx
```diff
@@ -12,16 +11,22 @@ export interface DateRangeCoordinatorProps {
     children?: React.ReactNode;
     innerRef?: React.Ref<HTMLDivElement>;
     sticky?: boolean;
+    useCustomNavigator?: boolean;
```
ajmenca, line 14: This customization is really awesome - my only complaint about it is the usage of this boolean.  It feels redundant to have to pass this flag *in addition* to passing the actual navigator as a child.

I think it might be cleaner (or at least, maybe more misuse-proof) if this was simply `customNavigator?: React.ReactNode`.  Then line 58 below becomes `{props.customNavigator || <DateRangeNavigator... `.

Thoughts?

ajmenca: Sort of an attached thought to this is that another sort of potential for mis-use here is that calling the `DateRangeContext.update` function is the single expectation of what a custom navigator must do at some point - but that fact isn't exactly *conveyed* anywhere.  It's maybe possible to convey that via types/interfaces if this isn't a boolean but a specific expected type that assures that it calls `update`?

src/components/container/GlucoseChart/GlucoseChart.tsx
greinard: It works because the `GlucoseDayCoordinator` does not render its children until it has finished loading its readings.

src/components/presentational/DateRangeCoordinator/DateRangeCoordinator.tsx
greinard: I like both of these suggestions.  I'll tinker with it a bit and see what I can come up with.

greinard: Hmm.  I think we may need to leave things as-is for the highest level of customizability.

If we use the `{props.customNavigator || <DateRangeNavigator... ` approach, we force the custom navigator to be above the children.  Having it as part of the children means the navigator can be placed above, within, or below the child components.  We might be able to dynamically look for a specific type of child/descendent component in order to hide the default navigator, but having the boolean flag to disable the default navigator seems like a less fragile approach.

I also can't think of a good way to ensure the navigators call the update function on the context.  However, without calling it, they don't really do anything, so I think it would quickly become apparent to someone creating a custom navigator that they need to call something to adjust the date, which would be the update function on the context.

src/components/container/GlucoseDayCoordinator/GlucoseDayCoordinator.tsx
greinard: Done.

src/components/container/GlucoseDayCoordinator/GlucoseDayCoordinator.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    const loadData = async (startDate: Date, endDate: Date): Promise<void> => {
+        const readings = props.previewState ?
+            await previewData(props.previewState, startDate, endDate) :
+            await getGlucoseReadings(startDate, endDate);
+
+        setGlucoseReadings(readings);
+        setGlucoseRanges(computeGlucoseReadingRanges(readings));
+        setRecentAverage(computeGlucoseReadingRecentAverage(readings, add(startDate, { days: 14 })));
+    };
```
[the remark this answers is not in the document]
greinard: I believe React schedules and batches state updates in order to avoid unnecessary re-renders already.

greinard:
I would like to merge this sometime in the next couple of days if there are no further revision requests or requests for delay. 

Thank you!

ajmenca (approved):

PR #318, opened 2024-09-27
Fix DailyDataGoal not updating when date changes

reneefromhold:
## Overview

When using the DailyDataGoal within a DateRangeCoordinator (already supported), when the user navigates to a new day

**Current Behavior:**
The data does not update
The data is searched for across a span of 3 days

**Expected Behavior:**
The data should reflect the current date
The data is searched for should be limited to one day

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

## Summary by CodeRabbit

- **New Features**
	- Introduced a new story for `LiveGarminSleepScore` with updated goal settings and messages.
	- Added a new story for `LiveSleepMinutesWithDateRangeCoordinator` to enhance component rendering.

- **Improvements**
	- Enhanced date handling logic for better accuracy in daily data queries.

- **Refactor**
	- Renamed existing story from `SleepGoalLive` to `LiveSleepMinutes` and updated rendering method for consistency.
	- Removed the `WearTimeGoalLive` story to streamline options.

### DIFF
--- a/src/components/container/DailyDataGoal/DailyDataGoal.stories.tsx
+++ b/src/components/container/DailyDataGoal/DailyDataGoal.stories.tsx
@@ -4,6 +4,7 @@ import Layout from "../../presentational/Layout"
 import { DailyDataType } from "../../../helpers";
 import DailyDataGoal, { DailyDataGoalProps } from "./DailyDataGoal";
 import { add, startOfDay } from "date-fns";
+import { DateRangeCoordinator } from "../../presentational";
 
 export default {
     title: "Container/DailyDataGoal",
@@ -94,7 +95,7 @@ export const Zero = {
 };
 
 
-export const SleepGoalLive = {
+export const FitbitSleepMinutesGoalLive = {
     args: {
         goal: 1,
         dailyDataType: DailyDataType.FitbitSleepMinutes,
@@ -114,34 +115,40 @@ export const SleepGoalLive = {
     render: render
 };
 
-export const WearTimeGoalLive = {
+const renderWithDRC = (args: DailyDataGoalProps) => <Layout colorScheme='auto'>
+    <DateRangeCoordinator intervalType="Day">
+        <DailyDataGoal {...args} />
+    </DateRangeCoordinator>
+</Layout>;
+
+export const LiveGarminSleepScore = {
     args: {
-        goal: 10000,
-        dailyDataType: DailyDataType.FitbitSteps, // TODO Wear time 
-        title: "Worn 10 Hours",
-        subtitle: "100 points",
+        goal: 100,
+        dailyDataType: DailyDataType.GarminSleepScore,
+        title: "Garmin Sleep Score",
+        subtitle: "100 is the perfect score",
         messages: [
             {
                 threshold: 0,
                 message: "No points yet"
             },
             {
-                threshold: 2500,
-                message: "Keep going!"
+                threshold: 70,
+                message: "Min Score 70!"
             },
             {
-                threshold: 5000,
-                message: "Halfway there!"
+                threshold: 85,
+                message: "Min Score 85"
             },
             {
-                threshold: 7500,
-                message: "Almost there!"
+                threshold: 91,
+                message: "Min Score 95"
             },
             {
-                threshold: 10000,
+                threshold: 100,
                 message: "Complete!"
             }
         ]
     },
-    render: render
+    render: renderWithDRC
 };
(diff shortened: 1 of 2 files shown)


src/components/container/DailyDataGoal/DailyDataGoal.tsx
```diff
@@ -33,13 +33,13 @@ export default function (props: DailyDataGoalProps) {
     const date = props.date ?? dateRangeContext?.intervalStart ?? startOfDay(new Date());
 
     useInitializeView(() => {
-        const startDate = add(date, { days: -1 });
-        const endDate = add(date, { days: 2 });
+        const startDate = startOfDay(date);
+        const endDate = endOfDay(date);
```
ajmenca, line 37: I'm curious - was this actually causing misbehaviors?  I'm pretty sure that, somewhere a couple of layers below this in the actual queries we already "pad" the date range requested for a variety of timezone/sync related reasons, but this code still only looks at the result of a single day.  So was the only "bad effect" of this that we're getting a larger data set back than necessary?

src/components/container/DailyDataGoal/DailyDataGoal.tsx
```diff
@@ -33,13 +33,13 @@ export default function (props: DailyDataGoalProps) {
...
 
         queryDailyData(props.dailyDataType, startDate, endDate, props.previewState === "Default").then((results) => {
             setDailyValue(results[getDayKey(date)] || 0);
         });
-    }, ["externalAccountSyncComplete"], []);
+    }, ["externalAccountSyncComplete"], [dateRangeContext?.intervalStart]);
```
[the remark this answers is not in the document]
ajmenca: This is incorrect, as this is not a dependency array being passed to `useEffect`, but an array being passed to `useInitializeView`.  There are purposefully two array parameters to this function, the first is used for registering for MyDataHelps events, and the second is used for `useEffect` dependencies.

src/components/container/DailyDataGoal/DailyDataGoal.stories.tsx
```diff
@@ -114,34 +115,40 @@ export const SleepGoalLive = {
...
 };
 
-export const WearTimeGoalLive = {
+const renderWithDRC = (args: DailyDataGoalProps) => <Layout colorScheme='auto'>
+    <DateRangeCoordinator intervalType="Day">
+        <DailyDataGoal {...args} />
+    </DateRangeCoordinator>
+</Layout>;
+
+export const LiveGarminSleepScore = {
```
ajmenca, line 124: Minor suggestion - I think the story names are a little confusing here.  It sounds like the only difference between the story you renamed and the story you edited is the type of data, but one also uses the `DateRangeCoordinator` and the other doesn't.  My suggestion would actually be 4 stories in this case; two for each data type with and without the `DateRangeCoordinator`.

src/components/container/DailyDataGoal/DailyDataGoal.tsx
reneefromhold: This is intended for looking at one day of data, what is the value of padding extra days back/fwd? Is there a use case or something with the returned data that isn't apparent here?

src/components/container/DailyDataGoal/DailyDataGoal.stories.tsx
reneefromhold: There is a unified sleep dailyDataType, but in this story it wasn't used. I think it makes more sense to update to use the unified vs manf specific sleep data type and revert the name back to imply it is not manf specific data. 

I'll update the new story I added with the Date Range Coordinator to use the same unified data type. There are 65ish DailyDataTypes. Adding a live story for each one is out of scope for this change.

src/components/container/DailyDataGoal/DailyDataGoal.tsx
ajmenca: I wasn't suggesting changing it back, I was asking if it was *actually* causing a participant facing bad behavior, or just extra data load.  As I said, in the underlying code it buffers the timestamps slightly to ensure you actually get the data you're looking for, so this double buffering certainly isn't necessary.  I was just curious.

reneefromhold: No, it wasn't showing the wrong data. I came across it while troubleshooting and found it misleading

ajmenca (approved):

PR #300, opened 2024-09-13
Add category filtering to SurveyTaskList

felideon:
## Overview

Paves way for https://github.com/CareEvolution/MyDataHelpsViewBuilder/issues/112

## Security

REMINDER: All file contents are public.

- [x] I have ensured no secure credentials or sensitive information remain in code, metadata, comments, etc.
  - [x] Please verify that you double checked that .storybook/preview.js does not contain your participant access key details. 
  - [x] There are no temporary testing changes committed such as API base URLs, access tokens, print/log statements, etc.
- [x] These changes do not introduce any security risks, or any such risks have been properly mitigated.

No security risks come to mind.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app
- [ ] [MyDataHelps website](mydatahelps.org)

### Documentation
- [x] I have added relevant Storybook updates to this PR as well.

Consider "Squash and merge" as needed to keep the commit history reasonable on `main`.

## Reviewers

Assign to the appropriate reviewer(s). Minimally, a second set of eyes is needed ensure no non-public information is published. Consider also including:
- Subject-matter experts
- Style/editing reviewers
- Others requested by the content owner

## Summary by CodeRabbit

- **New Features**
	- Introduced a `surveyCategory` property to enhance organization of survey tasks.
	- Added a new story variant for the `SurveyTaskList` component focused on the 'Pregnancy Study' category.
	- Implemented an optional `category` property in the `SurveyTaskList` component to filter tasks by survey category.

- **Bug Fixes**
	- Enhanced task filtering logic to ensure proper display of tasks based on the provided category.

PR #322, opened 2024-10-03
Add ability to hide SurveyTaskList if empty

felideon:
## Overview

Adds `hideIfEmpty` and `emptyText` props to the SurveyTaskList component.

- For `incomplete` task lists, `hideIfEmpty` helps users not have to implement their own conditionals using schedules/custom fields, and makes it nice to use multiple task lists on the same view, which the [surveyCategory](https://github.com/CareEvolution/MyDataHelpsUI/pull/300) property now makes more likely to occur.
- Previously, `complete` task lists were always hidden if empty. The new property allows users to display an empty complete task list, but, by default (i.e. if `hideIfEmpty` is undefined) will keep the previous behavior.
- `emptyText` provides a way for users to override the default string for empty lists, which was also changed (and re-translated). Previously, only `incomplete` task lists had a message when empty, but now there's also a default message for empty `complete` task lists.

Also includes a refactor to make the SurveyTaskList stories code consistent with more recent components, fixing a [type deprecation](https://github.com/CareEvolution/MyDataHelpsUI/pull/322/commits/fb684039a61a04c27012af7ca24cb9e25a9e9df8).

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation
- [ ] If this feature requires a developer doc update

## Summary by CodeRabbit

- **New Features**
	- Enhanced `SurveyTaskList` component to manage and display empty states with customizable messages.
	- Added new stories in Storybook to demonstrate various states of the `SurveyTaskList`, including `EmptyIncomplete` and `EmptyComplete`.
	- Updated localization for task completion messages across multiple languages to provide clearer feedback on task statuses.

- **Bug Fixes**
	- Improved handling of loading states and rendering logic when no tasks are available.


chrisnowak (approved):

src/components/container/SurveyTaskList/SurveyTaskList.tsx
```diff
@@ -136,6 +144,10 @@ export default function (props: SurveyTaskListProps) {
 		return null;
 	}
 
+	if (props.status == "incomplete" && !tasks?.length && props.hideIfEmpty) {
+		return null;
+	}
+
```
ajmenca, line 150: The combination here made me double check the `SurveyTaskStatus` type, which then begs the question of what the proper behavior is if `status` is `closed`.  Admittedly, it's probably an infrequent case - but I feel like `complete` is a somewhat rare case too?  And the fact that the list *always* hides if you're using `complete` is probably a largely unknown functionality?  I think it's confusing that 2 of the 3 statuses would basically ignore this property though.  This sort of interaction is largely hidden and hard to document.

Maybe it's simply worth collapsing both of these cases down to remove `status` as a variable at all with the addition of the `hideIfEmpty` property?  There would be a slight chance of behavior change if someone upgraded and was using `complete` because `hideIfEmpty` is technically `false` if it isn't set.  We could either accept that risk, or we could swap this to `showIfEmpty` (though I think that's a little more confusing, potentially).

felideon: In terms of rendering the task list, `closed` is not really a valid [List Type](https://github.com/CareEvolution/MyDataHelpsViewBuilder/blob/main/src/nodes/SurveyTaskListNode.tsx#L71-L74), as far as View Builder is concerned [1]. So maybe `status` should not be typed to `SurveyTaskStatus` since it includes `closed`. The problem is that it's used as a [task query parameter](https://github.com/CareEvolution/MyDataHelpsUI/pull/322/files#diff-f70c12f89bdafafbba64c0a348a060fd6d05ce450e3f3dec972822eb63026264R102). Otherwise I'd be tempted to rename/redefine that property as `listType: "incomplete" | "complete";`.

I would say `complete` is commonly used. For EMPOWER, we started without a completed tasks list (I think it adds a bit of noise to the UI), but then site coordinators that did not have access to MDHD had no way of confirming whether the participant had completed their ePROs (surveys) before, e.g., getting their lab work done. AoU also had a completed tasks section, of course.

Your point on the asymmetry between `incomplete` and `complete` having different hidden-or-not semantics still stands, though. I wouldn't say "the fact that the list always hides if you're using complete is probably a largely unknown functionality", as that should (in theory) become readily apparent the moment you test your view (if building with View Builder). It is certainly opinionated to ignore the new `hideIfEmpty` prop if your list type is `complete`, so we could arguably honor the prop, if set, as long as the default behavior is to hide if empty, to ensure backwards compatibility. I guess if you put a completed task list in its own view, having a message that says "No tasks have been completed." makes sense. But if you have an incomplete task list next to it, it's kind of unnecessary since you have the incomplete tasks staring at you.

Incidentally, there is a [`SurveyTasksView`](https://ui.mydatahelps.org/?path=/story/view-surveytasksview--hide-incomplete) that implements its own hiding mechanism, which I should probably update.

[1] Which makes sense in terms of participant-facing use cases, at least.

felideon: Regarding the `closed` status, I guess making this change wouldn't be too bad:
```
listType: "incomplete" | "complete";
...
var parameters: SurveyTaskQueryParameters = { status: props.listType }
```
But then we'd really be saying "do not use this component for closed tasks."

felideon: Ok, see latest commit.

ajmenca (approved):

PR #234, opened 2024-03-27
Docs: Update Storybook to v8

dylanbox:
## Overview

Updates the storybook install to the new version, v8, and updates stories to the new format. This is a prerequisite for implementing the design-system, as we want to use the new storybook inheritance and visual testing suite.

## Security

REMINDER: All file contents are public.

- [X] I have ensured no secure credentials or sensitive information remain in code, metadata, comments, etc.
  - [X] Please verify that you double checked that .storybook/preview.js does not contain your participant access key details. 
  - [X] There are no temporary testing changes committed such as API base URLs, access tokens, print/log statements, etc.
- [X] These changes do not introduce any security risks, or any such risks have been properly mitigated.

Describe briefly what security risks you considered, why they don't apply, or how they've been mitigated.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app
- [ ] [MyDataHelps website](mydatahelps.org)

### Documentation
- [X] I have added relevant Storybook updates to this PR as well.
- [ ] If this feature requires a developer doc update, tag @CareEvolution/api-docs.

Consider "Squash and merge" as needed to keep the commit history reasonable on `main`.

## Reviewers

Assign to the appropriate reviewer(s). Minimally, a second set of eyes is needed ensure no non-public information is published. Consider also including:
- Subject-matter experts
- Style/editing reviewers
- Others requested by the content owner

PR #334, opened 2024-10-18
Update Storybook & Start using jest

reneefromhold:
In short, this repo would benefit from more testing. 
-Update storybook from 7.0 to 8.5
-Introduce Jest
-Add unit tests for Custom and ExternalAccount activities for use in BasicPointsForBadges
-Add jsdoc comment at the component level to BasicPointsForBadges
-Render the previously mentioned jsdoc comment as the description of the Default story for BasicPointsForBadges. These descriptions will export as part of the npm package

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

## Summary by CodeRabbit

- **New Features**
	- Enhanced Storybook configuration for various components, improving documentation and story definitions.
	- Introduced unit tests for awarding points based on external account connections and custom activities.
	- Added a new mock data structure for participant information to facilitate testing.
	- Updated TypeScript configuration to support modern JavaScript features and React.
	- Added a new testing configuration for unit tests in the development environment.
	- Improved styling and organization of the `Introduction.mdx` file for better clarity.

- **Bug Fixes**
	- Updated logic for awarding points based on connected accounts and custom activities to ensure accuracy.

- **Documentation**
	- Improved JSDoc comments for multiple components, clarifying their purpose and functionality.

- **Chores**
	- Updated dependencies and scripts in `package.json` to enhance functionality and compatibility.
	- Modified build process in `amplify.yml` to include testing before building Storybook artifacts.


src/helpers/BasicPointsAndBadges/CustomActivity.test.ts
```diff
@@ -0,0 +1,60 @@
+import { ParticipantInfo } from "@careevolution/mydatahelps-js";
+import { CustomActivity } from "../../../dist";
+import { awardCustomActivityPoints } from "./CustomActivity";
```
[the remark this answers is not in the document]
zwkohn: this is a valid point

zwkohn (commented):
Maybe others feel differently, but I find having tests in a separate `tests` folder to be a less cluttered

reneefromhold:
> Maybe others feel differently, but I find having tests in a separate `tests` folder to be a less cluttered

I appreciate that feedback. I do like the tests within the same folder as the component bc it encourages creating it as part of component/helper completion and makes the tests easy to find/reference.  I am interested in hearing others thoughts too.

greinard:
> > Maybe others feel differently, but I find having tests in a separate `tests` folder to be a less cluttered
> 
> I appreciate that feedback. I do like the tests within the same folder as the component bc it encourages creating it as part of component/helper completion and makes the tests easy to find/reference. I am interested in hearing others thoughts too.

I tend to prefer a separate `tests` folder as well.  I agree that the co-location can make it easier to find the test for a given component, but it can also make it more difficult to tell which files in the folder are test related and which are not.  The `.test.ts` suffix helps with the actual test files, but if there are other `.ts` files to support the `.test.ts` files, then it can be trickier.

We'll also want to make sure that the tests and other test related files are not included in the published builds.  That'll probably be easiest if we can just exclude (or not include) the `tests` folder vs using a regex or wildcard string and naming conventions for tests (and related) files that are part of the `src` directory tree.

greinard:
Just putting this here so we don't forget, but the existing PR about upgrading the Storybook can probably go away once this one gets merged.

https://github.com/CareEvolution/MyDataHelpsUI/pull/234

.npmignore
greinard, line 1: I am not super familiar with this stuff, but I think the `.npmignore` file may not help us with this.

We already [whitelist the `dist` directory](https://github.com/CareEvolution/MyDataHelpsUI/blob/main/package.json#L102) as the only thing to include in the `npm publish`.  We just need to make sure the `dist` folder doesn't contain the test files.

I think we can modify [this line](https://github.com/CareEvolution/MyDataHelpsUI/blob/main/rollup.config.mjs#L28) of the rollup config to be something like:

`typescript({ tsconfig: "./tsconfig.json", exclude: ["**/__tests__", "**/*.test.ts"] }),`

I haven't tested that, though, so we may need to iterate on it to get the correct exclusion patterns.

All that said, now that the tests are in their own directory outside of the source tree, maybe we don't even need to add an exclusion at all?  That would probably be worth testing.

__tests__/fixtures/participantInfo.ts
greinard, line 1: I'm curious about the use of `__tests__` for the folder name instead of just `tests`.   Is that a naming convention for jest or something?

.storybook/main.js
```diff
@@ -1,11 +1,22 @@
 module.exports = {
-  "stories": ["../src/**/*.stories.mdx", "../src/**/*.stories.@(js|jsx|ts|tsx)"],
-  "addons": ["@storybook/addon-links", "@storybook/addon-essentials", "@storybook/addon-interactions"],
+  "stories": ["../src/**/*.mdx", "../src/**/*.stories.@(js|jsx|ts|tsx)"],
```
greinard, line 2: Should this first entry still be: `"../src/**/*.stories.mdx"`?

__tests__/helpers/BasicPointsAndBadges/ConnectExternalAccountActivity.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        default: {
+            token: { access_token: '1' },
+            getCurrentLanguage: () => jest.fn(),
+            getExternalAccounts: jest.fn(() => {
+                return Promise.resolve(externalAccounts);
+            }),
+            on: jest.fn(),
+            off: jest.fn(),
+        },
+        MyDataHelps: {
```
greinard, line 29: I believe you can remove this `MyDataHelps` entry.  I'm pretty sure the MDH-JS module doesn't export anything named `MyDataHelps`.  It just exports it as the `default`.

__tests__/helpers/BasicPointsAndBadges/ConnectExternalAccountActivity.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    return {
+        __esModule: true,
+        default: {
+            token: { access_token: '1' },
+            getCurrentLanguage: () => jest.fn(),
+            getExternalAccounts: jest.fn(() => {
+                return Promise.resolve(externalAccounts);
+            }),
+            on: jest.fn(),
+            off: jest.fn(),
```
greinard, line 27: Minor: Extra comma at the end of the line.

__tests__/helpers/BasicPointsAndBadges/CustomActivity.test.ts
```diff
@@ -0,0 +1,47 @@
...
+        };
+        const activity: CustomActivity = {
+            key: "CustomGoals",
+            type: "custom",
+            points: 8,
+            customField: "DailyGoals"
+        };
+
+        const newActivityState = await awardCustomActivityPoints(activity, pptInfoWithDecimal);
+        expect(newActivityState.pointsAwarded).toBe(9.6); // Assuming rounding down
```
greinard, line 45: You can probably remove this comment.  If you want to assert that a particular rounding strategy takes place, it could be done with some additional test cases.

jest.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+    // Other Jest config
+    testPathIgnorePatterns: ['/fixtures/'],
```
greinard, line 3: Minor:  No need for the comment here and there is an unnecessary comma.

__tests__/fixtures/participantInfo.ts
reneefromhold: @greinard That is what I am finding in most examples for setting up jest. While it does catch my eye, I do not feel strongly about keeping it with the underscores

src/components/container/BasicPointsForBadges/BasicPointsForBadges.stories.tsx
```diff
@@ -8,22 +8,31 @@ import DailyDataGoal from "../DailyDataGoal";
 import { Grid, Title } from "../../presentational";
 import BasicBadges from "../BasicBadges";
 import Divider from "../../presentational/Divider";
+import { Description } from "@storybook/blocks"
+import { Meta, StoryObj } from "@storybook/react";
+import { ConnectExternalAccountActivity, CustomActivity, SurveyCompletionActivity } from "../../../../dist";
```
greinard, line 13: These should be imported from `../../../helpers`.

The `DailyDataActivity` import can also be simplified to `../../../helpers`.

All of them can be combined with the existing import for `DailyDataType`.

.storybook/main.js
reneefromhold: The .mdx extension itself is enough to indicate that the file contains both stories and Markdown content. During the upgrade, the one file - Introduction.stories.mdx - was auto converted to Introduction.mdx.

__tests__/fixtures/participantInfo.ts
reneefromhold: @greinard Keeping the __tests__ name will put the folder at the top of the list in the ide

src/helpers/BasicPointsAndBadges/ConnectExternalAccountActivity.ts
```diff
@@ -1,18 +1,26 @@
...
+    providerCategories: ConnectExternalActivityType[];
 }
 
 interface ConnectExternalAccountActivityState extends BasicPointsForBadgesActivityState {
     providersConnected?: number[];
 }
 
 export async function awardConnectExternalAccountActivityPoints(activity: ConnectExternalAccountActivity, activityState: ConnectExternalAccountActivityState) {
+    //convert to array to protect against new types added in consumers but not yet here
+    const activityProviderCategories = activity.providerCategories as string[];
```
greinard, line 17: I think you can just use `account.provider.category as ConnectExternalActivityType` below to avoid the typescript error.  If the value from the server is not one of the currently defined values, it just won't match anything when `includes` gets called.

src/helpers/BasicPointsAndBadges/ConnectExternalAccountActivity.ts
```diff
@@ -1,18 +1,26 @@
...
 }
 
 export async function awardConnectExternalAccountActivityPoints(activity: ConnectExternalAccountActivity, activityState: ConnectExternalAccountActivityState) {
+    //convert to array to protect against new types added in consumers but not yet here
+    const activityProviderCategories = activity.providerCategories as string[]; 
+
+    //Get the external accounts
     const connectedAccounts = await MyDataHelps.getExternalAccounts();
-    const allProviders = connectedAccounts.filter(account => activity.providerCategories.includes(account.provider.category)).map(t => t.id);
+    //Filter for the accounts that match this activity. Return the external account id
```
greinard, line 21: I think these are probably simple enough operations that the comments probably aren't necessary.  I also tend to feel like comments get out of date fairly easily when adjustments or refactorings occur, so they often can't be relied upon.

If some form of inline documentation is desirable, I think a slightly less brittle way to do it is to create helper functions that clearly describe what is happening with the function name.  For example, something like:

```
function filterAccountsByActivity(connectedAccounts: ExternalAccount[], activity: ConnectExternalAccountActivity) : number[] {
    return connectedAccounts.filter(account => activity.providerCategories.includes(account.provider.category as ConnectExternalActivityType)).map(account => account.id);
}
```

It is probably splitting hairs, but I think the function name is more likely to be updated if the functionality changes than the comment.

Of course, the best way to document what is/should be happening is with the unit tests you've created.  😃

tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "allowJs": true,
     "sourceMap": true,
     "moduleResolution": "node",
-    "allowSyntheticDefaultImports": true
+    "allowSyntheticDefaultImports": true,
+    "outDir": "./dist",
```
greinard, line 17: Was something not working prior to this?  I'm curious about why this was needed now, but not before.

.storybook/main.js
greinard: Sounds good.  Thanks for the explanation.

src/helpers/BasicPointsAndBadges/ConnectExternalAccountActivity.ts
```diff
@@ -1,18 +1,26 @@
...
     providersConnected?: number[];
 }
 
 export async function awardConnectExternalAccountActivityPoints(activity: ConnectExternalAccountActivity, activityState: ConnectExternalAccountActivityState) {
+    //convert to array to protect against new types added in consumers but not yet here
+    const activityProviderCategories = activity.providerCategories as string[]; 
+
+    //Get the external accounts
     const connectedAccounts = await MyDataHelps.getExternalAccounts();
-    const allProviders = connectedAccounts.filter(account => activity.providerCategories.includes(account.provider.category)).map(t => t.id);
```
greinard, line 15: As we discussed, I believe this should be using the `t.provider.id` instead of `t.id` since the external account connection ids can change if disconnected and reconnected.  The provider ids should remain the same.  Probably worth renaming `t` to `account` as well, for better understanding of what is being mapped at a glance.

Also, as discussed, the state for this activity should track all provider ids that contributed to the points, not just the current list of connected ones that still match the criteria.  Therefore, the `providersConnected` array should be being  appended to rather than replaced when new provider connections are found.

.npmignore
reneefromhold: Notes from yesterday's chat, 
-changes in .npmignore will be rolled back. Most of what was updated here is already handled by the dist setting in package.json
-tests have been relocated external to src folder and are not packaged

src/helpers/BasicPointsAndBadges/ConnectExternalAccountActivity.ts
reneefromhold: I agree. Especially with the change to map provider id and ut around it.

tsconfig.json
reneefromhold: It's not needed, and was removed.

src/components/step/ConnectEhrStepContainer/ConnectEhrStepContainer.stories.tsx
```diff
@@ -1,20 +1,26 @@
 import React from "react";
-import { ComponentStory, ComponentMeta } from "@storybook/react";
-import ConnectEhrStepContainer, {
-    ConnectEhrStepContainerProps,
-} from "./ConnectEhrStepContainer";
+import { Meta, StoryObj } from "@storybook/react";
```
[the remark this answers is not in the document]
reneefromhold: TODO for another PR

__tests__/helpers/BasicPointsAndBadges/ConnectExternalAccountActivity.test.ts
```diff
@@ -0,0 +1,115 @@
...
+        status: "fetchComplete"
+    },
+    {
+        "id" : 2,
+        "provider": { "id": 22, "name": "Fitbit", "category": "Device Manufacturer"},
+        status: "fetchComplete"
+    },
+    {
+        "id" : 3,
+        "provider": { "id": 11, "name": "FitBit", "category": "Device Manufacturer"} ,
```
greinard, line 17: Maybe use an `id` of 33 here to avoid any potential collisions with the Garmin provider.

greinard: Actually, I just realized this is a second Fitbit provider.  Was that intentional?  A don't think a user should be able to have two Fitbit connections, but if they did, the provider would have the same id.

__tests__/helpers/BasicPointsAndBadges/ConnectExternalAccountActivity.test.ts
```diff
@@ -0,0 +1,115 @@
...
+   });
+
+   it("should append points and update providers despite prior providers no longer being connected", async () => {
+        const activity = {
+            type: "connectExternalAccount",
+            points: 10,
+            providerCategories: ["Health Plan"]
+        } as ConnectExternalAccountActivity;
+        const activityState = {
+            pointsAwarded: 10,
```
greinard, line 93: Minor:  Given the rest of the test setups, I think this would actually be 20 points here to start out if they already had two successful connections to providers 1 and 2.

tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "allowJs": true,
     "sourceMap": true,
     "moduleResolution": "node",
-    "allowSyntheticDefaultImports": true
+    "allowSyntheticDefaultImports": true,
+    "noEmit": true
```
greinard, line 17: Is this because we are using `rollup` to build the modules?

reneefromhold: Correct [-more info here](https://www.typescriptlang.org/tsconfig/#noEmit)

__tests__/helpers/BasicPointsAndBadges/ConnectExternalAccountActivity.test.ts
reneefromhold: I updated the 3rd external accounts provider id to reflect the desired mock data - multiple external accounts connected to the same providers (Fitbit/Garmin/Etc). 
I have updated the provider id to match the fitbit provider

src/components/container/ConnectFitbit/ConnectFitbit.stories.tsx
```diff
@@ -1,48 +1,103 @@
...
 
-export default {
+
+const meta: Meta<typeof ConnectFitbit> = {
 	title: "Container/ConnectFitbit",
 	component: ConnectFitbit,
 	parameters: {
 		layout: 'fullscreen',
+		docs: {
+			//TODO: Description: <Description />
```
greinard, line 14: Since this is optional and probably applies to all of our stories, can we remove the TODO comment for now?  Or go ahead and implement the description?

src/components/container/ConnectFitbit/ConnectFitbit.stories.tsx
```diff
@@ -1,48 +1,103 @@
...
+		<ConnectFitbit {...args} />
+	</Card>
+</Layout>;
 
-export const NotConnected = Template.bind({});
-NotConnected.args = { previewState: "notConnected", title: "Fitbit" };
+export const NotConnected : Story = {
+    args: {
+        previewState: "notConnected", 
+		title: "Fitbit"
```
greinard, line 31: Minor:  The formatting looks a little off on these.

src/components/step/ConnectEhrStepContainer/ConnectEhrStepContainer.stories.tsx
```diff
@@ -1,20 +1,26 @@
...
+const meta: Meta<typeof ConnectEhrStepContainer> = {
     title: "SurveyStep/ConnectEhrStepContainer",
     component: ConnectEhrStepContainer,
     parameters: {
-        layout: "fullscreen",
-    },
-} as ComponentMeta<typeof ConnectEhrStepContainer>;
+        layout: 'fullscreen',
+        docs: {
+            //TODO: Description: <Description />
```
greinard, line 13: Same question here regarding the TODO comment.  Can we remove it or implement the description?

__tests__/helpers/BasicPointsAndBadges/ConnectExternalAccountActivity.test.ts
reneefromhold: This demonstrates that there is no recalculating of points done based on the providers that had been connected in the past. The description has been updated to reflect the intent of the test

src/components/container/ConnectFitbit/ConnectFitbit.stories.tsx
reneefromhold: I'm removing them for now. I will clean up the rollup in a new PR once this one is merged

__tests__/helpers/BasicPointsAndBadges/ConnectExternalAccountActivity.test.ts
greinard: Right, I think the description is fine.  What I mean is, just to avoid confusion, the initial state here could be 20 points awarded, since they already have/had providers 1 and 2 connected.  The results below would then be `30` and `[1, 2, 44]`.

__tests__/helpers/BasicPointsAndBadges/ConnectExternalAccountActivity.test.ts
```diff
@@ -0,0 +1,115 @@
...
+        const activityState = {
+            pointsAwarded: 0,
+            providersConnected: []
+        };
+        const newActivityState = await awardConnectExternalAccountActivityPoints(activity, activityState);
+        expect(newActivityState.pointsAwarded).toBe(20);
+        expect(newActivityState.providersConnected).toEqual([11, 22]);
+   });
+
+   it("should recaluclate points across all providers. Awarding points should be based on newly connected", async () => {
```
greinard, line 86: Minor:  Typo in recalculate.  I think the prior description was fine too, since we aren't actually recomputing the value, we are just appending points for the newly connected ones, but I don't feel that strongly about it either way.

greinard (approved):

PR #352, opened 2024-11-07
Add a new InsightMatrix component (formerly OverviewTable).

greinard:
## Overview

This branch adds a new `InsightMatrix` component that can be used to compare multiple daily survey and daily data values in a tabular format.

The group-by data is first grouped by value threshold.  The comparison data are then grouped by the same days in which the group-by data values met each threshold.  This is intended to illustrate what the values from each of the comparison data looked like on the days when the values from the group-by data were within a given threshold (i.e. What did my step count look like when my mood rating was high?).

Thresholds, labels, colors, value calculators, value formatters, and value evaluators (good vs bad) can be customized on the group-by and comparison data types.

image
image

I will revert the snapshot version number prior to merging.

## Security

- No new security risks.  This component just facilitates displaying already available data in a tabular format.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Release Notes

- **New Features**
  - Introduced the `InsightMatrix` component for displaying comparative data in a matrix format.
  - Added a preview data provider for the `InsightMatrix` component, enabling mock data generation.
  - New translations for "No comparison data types configured" in multiple languages.

- **Improvements**
  - Enhanced Storybook configuration for better testing and visualization of the `InsightMatrix` component.
  - Added new CSS styles for improved layout and presentation of the `InsightMatrix`.

- **Version Update**
  - Updated package version to `2.49.1-OverviewTable.3`. 

These changes enhance the functionality and user experience of the application, particularly around data visualization and internationalization.


src/components/container/OverviewTable/OverviewTable.tsx
```diff
@@ -0,0 +1,146 @@
...
+import { add, startOfToday } from 'date-fns';
+import { createPreviewDataProvider } from './OverviewTable.previewData';
+import { IconDefinition } from '@fortawesome/fontawesome-svg-core';
+import { FontAwesomeSvgIcon } from 'react-fontawesome-svg-icon';
+import language from '../../../helpers/language';
+
+export interface OverviewTableProps {
+    preview?: boolean;
+    primaryDataType: PrimaryOverviewDataType;
+    secondaryDataTypes: SecondaryOverviewDataType[];
```
ajmenca, line 14: Usage-wise, I'll admit that I just spent a surprisingly long time understanding the fact that data *only shows up in the table* if you pass it in as a Secondary type.  This is... kind of confusing?  I was lost for a little bit that setting the Primary type to "Mood" doesn't cause the mood *values* to show up in the table.  I think these two inputs could use some naming thought, because Primary and Secondary don't really convey their usage (to me, at least).  It's also a little confusing for me because you're not actually defining a "data type" (there's a data type *in* this type, accompanied by customization parameters).

Alternatively, maybe it's just the setup of the story that lead me down the confusing path, as the default state passes *all* of the types as secondary data types.  Because of this, the primary is set to Mood, but Mood is also the leftmost column of the table, so it made me think that was... purposeful?  

I could be way off base here, this was just my impression coming in without a frame of reference.  I think I'd go with something like `OverviewTableGroupByDataParameters` and `OverviewTableComparisonDataParameters`.

src/components/container/OverviewTable/OverviewTable.tsx
```diff
@@ -0,0 +1,146 @@
...
+        const primaryDataProvider = getPrimaryDataProvider();
+        primaryDataProvider.dataProvider(startDate, today).then(result => {
+            return { type: primaryDataProvider.type, queryResult: result };
+        }).then(setPrimaryData);
+
+        const secondaryDataProviders = getSecondaryDataProviders();
+        Promise.all(secondaryDataProviders.map(secondaryDataProvider => secondaryDataProvider.dataProvider(startDate, today).then(result => {
+            return { type: secondaryDataProvider.type, queryResult: result };
+        }))).then(setSecondaryData);
+    }, [], [props.preview, props.primaryDataType, props.secondaryDataTypes]);
```
ajmenca, line 84: Along with my confusion about whether or not the Primary data shows up in the table... if you *do* want it to show up in the table, it looks like we'll query the exact same data twice?

src/components/container/OverviewTable/OverviewTable.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+    if (!primaryData || !secondaryData) {
+        return <div className="mdhui-overview-table" ref={props.innerRef}><LoadingIndicator /></div>;
+    }
```
ajmenca, line 88: I'm having trouble hitting this state in the storybook to make sure that it works.  I think it's because `useInitializeView` doesn't reset `primaryData` and `secondaryData` to null/undefined values when it goes to re-load data?  So the storybook starts with preview data, and then if you switch it to live, you don't see this loading indicator because the "old" data hangs around in widget's state while the calls are pending.

src/components/container/OverviewTable/OverviewTable.tsx
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/**
+ * The overview table component can be used to compare multiple daily survey and daily data values in a tabular format.
+ *
+ * The primary data type is first grouped by value threshold.  The secondary data types are then grouped by the same
+ * days in which the primary values met each threshold.  This is intended to illustrate what the values from each of
+ * the secondary data types looked like on the days when the values from the primary data type were within a given
+ * threshold (i.e. What did my step count look like when my mood rating was high?).
+ */
```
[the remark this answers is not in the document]
greinard: Credit to @reneefromhold.  She told me about that feature the other day.

src/components/container/OverviewTable/OverviewTable.tsx
```diff
@@ -0,0 +1,146 @@
...
+    primaryDataType: PrimaryOverviewDataType;
+    secondaryDataTypes: SecondaryOverviewDataType[];
+    daysToCompute?: number;
+    primaryHeaderBackgroundColor?: ColorDefinition;
+    primaryHeaderTextColor?: ColorDefinition;
+    goodValueBackgroundColor?: ColorDefinition;
+    goodValueTextColor?: ColorDefinition;
+    goodValueIndicator?: IconDefinition;
+    notGoodValueBackgroundColor?: ColorDefinition;
+    notGoodValueTextColor?: ColorDefinition;
```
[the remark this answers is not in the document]
greinard: Both great questions.

> 1. Why _must_ a value be either good or not good?  I realize you can set "not good" to be a neutral color, and then your chart is only highlighting "good" and "normal" - but it feels like most of the time these sorts of thresholds are ternary instead of binary?  If I get under 6 hours of sleep it's not good, if I get 6-7 it's normal, and if I get over 7 it's good?  That sort of situation?

The original/existing reason for these colors was/is to be able to highlight and provide positive feedback when values are in some "good" range.  For my use case, that means having a minimum number of steps (i.e. 6k+).  For sleep, it means having enough sleep, but not too much (i.e. 7-10 hours).  Thus far, this has always been used in a binary fashion.  Things are either "good" and are highlighted as such, or they are put into the other bucket (currently referred to as "not good").  That isn't to say that those other values are necessarily "bad".  It just means they are not being highlighted as "good".  I'm certainly open to naming suggestions for the "values which are not considered good" category.  Maybe just "other"? 🙂

> 2. Similarly, it feels like we're avoiding using the term "bad" here - notably that there's not a "notGoodValueIndicator."  Are we really avoiding letting someone highlight "problem" areas with this graph?

The original use case for this widget was as part of a mental health study.  I think to some degree, there was purpose in not focusing on the negatives, but rather, just highlighting the positives and allowing the negatives (or non-positives) to be inferred.  That said, I wouldn't be opposed to adding support for negative highlighting as well, in case some future use case calls for that.

The `goodValueIndicator` is new and has not been used previously.  It is my attempt at making this table a tad more "accessible" for the color averse folks in the population.  Regardless of the color palette chosen, it facilitates marking the "positive" values with a non-color based indication.  The ones without the indication are the "non-positives".  Since the categorization is currently binary (i.e. good, other), I didn't think a second indicator was necessary.

I think if we do add support for negative highlighting, which creates a third categorization (i.e. good, bad, other), then I do think it would be appropriate to add a `badValueIndicator` option as well.

src/components/container/OverviewTable/OverviewTable.tsx
greinard: I like that suggestion.  The "type in a type" thing did feel a tad off to me.

How about `OverviewTableGroupByDataConfiguration` and `OverviewTableComparisonDataConfiguration`?  "Parameters" doesn't feel like quite the right word to me, but I don't feel that strongly about it either way.

RE: the story setup - I went back and forth a couple of times on that, but I wanted to make all of the data types available to be either the primary or a secondary, for demonstration purposes.  I could adjust the default to not include mood rating in the comparison data types, if you think that would help.

src/components/container/OverviewTable/OverviewTable.tsx
greinard: Yes, we would.

I think in practice it will be rare to have the same data type in both the primary and secondary configuration.

It wouldn't provide much value to see that when my steps are categorized as high (primary column), they are also high numerically (secondary column).  Likewise for other data types.

I think it might be a tad complex to try to share data providers across the primary/secondary boundary, so I'm not sure the juice is worth the squeeze given the expected usage of this component.

src/components/container/OverviewTable/OverviewTable.tsx
greinard: Yeah.  This was intentional to avoid having the table flash and resize each time the screen gets foregrounded.

You can test the loading behavior by hitting the "Remount component" button at the top of the Storybook render panel.

image

You can use this while in preview mode or after switching to live to see the loading indicator.

chrisnowak:
@greinard I would consider a less broad name for this - OverviewTable is quite broad.

How about "InsightMatrix"?  (Credit to ChatGPT). Feels like it a well distinguished name that captures the idea behind the component (participants having insights about their health) in a matrix

greinard:
> How about "InsightMatrix"?

I like that.

src/helpers/overview-table/survey-data-type.ts
```diff
@@ -0,0 +1,50 @@
...
+export function isSurveyDataType(dataType: DailyDataType | SurveyDataType): dataType is SurveyDataType {
+    return dataType && (dataType as any).surveyName !== undefined;
+}
+
+export function getSurveyDataProvider(dataType: SurveyDataType): DailyDataProvider {
+    return async (startDate: Date, endDate: Date) => {
+        const query: SurveyAnswersQuery = {
+            surveyName: dataType.surveyName,
+            stepIdentifier: dataType.stepIdentifier,
+            after: add(startDate, { days: -1 }).toISOString(),
```
reneefromhold, line 23: Why are days being added to?

greinard: I usually do that just to ensure that we get all the data we want back from the query, without having to worry about any cross-platform (client/server/db) timezone offset nuances around the day boundaries.  It probably isn't necessary, but, at least in this case, the extra days only add 1 or 2 more survey answers to the returned result, so it is relatively harmless for the assurance it provides.

I think it is a pretty common pattern in use by the majority of our other daily data providers, too, possibly for a similar reason.

src/helpers/overview-table/value-threshold.ts
```diff
@@ -0,0 +1,49 @@
...
+    minimumValue?: number;
+    maximumValue?: number;
+}
+
+export type OverviewThresholdDaysLookup = { [key: string]: string[] };
+
+export const NotEnteredThreshold = 'Not Entered';
+
+export function computeThresholdDays(startDate: Date, endDate: Date, overviewData: OverviewData<PrimaryOverviewDataType>): OverviewThresholdDaysLookup {
+    const lookup: OverviewThresholdDaysLookup = {};
```
reneefromhold, line 17: Can unit tests be added for this?

greinard: Yes.  Will do.  I'll also work on adding them to a number of other bits of code as well.

src/components/container/OverviewTable/OverviewTable.tsx
```diff
@@ -0,0 +1,146 @@
...
+    const primaryHeaderBackgroundColor = resolveColor(layoutContext.colorScheme, props.primaryHeaderBackgroundColor ?? { lightMode: '#e2f1ff', darkMode: '#000' });
+    const primaryHeaderTextColor = resolveColor(layoutContext.colorScheme, props.primaryHeaderTextColor ?? { lightMode: '#000', darkMode: '#8bcbff' });
+
+    const goodValueBackgroundColor = resolveColor(layoutContext.colorScheme, props.goodValueBackgroundColor ?? { lightMode: '#c8ffc2', darkMode: '#000' });
+    const goodValueTextColor = resolveColor(layoutContext.colorScheme, props.goodValueTextColor ?? { lightMode: '#000', darkMode: '#96ff85' });
+
+    const notGoodValueBackgroundColor = resolveColor(layoutContext.colorScheme, props.notGoodValueBackgroundColor ?? { lightMode: '#ffedc2', darkMode: '#000' });
+    const notGoodValueTextColor = resolveColor(layoutContext.colorScheme, props.notGoodValueTextColor ?? { lightMode: '#000', darkMode: '#ffc847' });
+
+    const daysToCompute = props.daysToCompute ?? 28;
```
reneefromhold, line 49: Is 28 used bc of days in month? If yes, can getDaysInMonth be used?

src/components/container/OverviewTable/OverviewTable.tsx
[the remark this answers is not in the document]
greinard: I have renamed "notGood" to "other" for now.  I will also see about adding support for highlighting "bad" values.

src/components/container/OverviewTable/OverviewTable.tsx
greinard: I have renamed these to `InsightMatrixGroupByDataConfiguration` and `InsightMatrixComparisonDataConfiguration`.

I have also removed "mood" from the default comparison types on the storybook.

greinard:
I have renamed the component to `InsightMatrix`.  I'll update the PR title/description momentarily.

greinard:
@coderabbitai pause

src/components/container/OverviewTable/OverviewTable.tsx
greinard: The 28 day default is because the original use case for this component was to compute data from the past 4 weeks.  We can pick a different default, but I think it should be a fixed number for consistency rather than something variable like the number returned by `getDaysInMonth`.

src/components/container/InsightMatrix/InsightMatrix.tsx
```diff
@@ -0,0 +1,146 @@
...
+    };
+
+    return <div className="mdhui-insight-matrix" style={{ gridTemplateColumns: '112px 1fr ' + Array(Math.max(0, comparisonData.length - 1)).fill('1fr').join(' ') }} ref={props.innerRef}>
+        <div className="mdhui-insight-matrix-header-group-by" style={{ background: groupByHeaderBackgroundColor, color: groupByHeaderTextColor }}>
+            {groupByData.configuration.label ?? getDefaultDataLabel(groupByData.configuration)}
+        </div>
+        {comparisonData.length === 0 &&
+            <div className="mdhui-insight-matrix-no-comparison-data">{language('insight-matrix-no-comparison-data')}</div>
+        }
+        {comparisonData.map(data => data.configuration).map((configuration: InsightMatrixComparisonDataConfiguration, index: number) => {
```
reneefromhold, line 110: Can the map/looping here and on line 120 be consolidated into one?

src/components/container/InsightMatrix/InsightMatrix.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+    const getDefaultDataLabel = (configuration: InsightMatrixDataConfiguration): string => {
+        if (isSurveyDataType(configuration.rawDataType)) {
+            return configuration.rawDataType.stepIdentifier;
+        }
+        const labelKey = getDailyDataTypeDefinition(configuration.rawDataType).labelKey;
+        return labelKey ? language(labelKey) : configuration.rawDataType;
+    };
+
+    return <div className="mdhui-insight-matrix" style={{ gridTemplateColumns: '112px 1fr ' + Array(Math.max(0, comparisonData.length - 1)).fill('1fr').join(' ') }} ref={props.innerRef}>
```
reneefromhold, line 103: Is a LoadingIndicator needed?

src/components/container/InsightMatrix/InsightMatrix.tsx
greinard: It is possible to invert the grid to fill column by column instead of row by row, but the code doesn't really get any more simplified (perhaps less).  You still end up looping something in two places, either the data types to build the header row or the thresholds to build the first column.

I think the row by row rendering is a little easier for me to understand.  The first loop here builds the header row, and then the loop on 120 building each threshold's row.

src/components/container/InsightMatrix/InsightMatrix.tsx
greinard: The loading indicator is up on line 87: https://github.com/CareEvolution/MyDataHelpsUI/pull/352/files#diff-37c4e9f5691934f943b8da5a8f6e08abbb815de42c9da6481bf3b55a0c6abe98R87

src/helpers/overview-table/value-threshold.ts
greinard: I have added a bunch of unit tests for the helper functions.

greinard:
I think I have addressed or responded to all of the comments thus far.  Please let me know if there is anything further that you would like me to revisit.

Thanks!

src/helpers/insight-matrix/value-calculator.ts
```diff
@@ -0,0 +1,23 @@
+export interface InsightMatrixValueCalculator {
+    calculate: (thresholdDays: string[], values: number[]) => number | undefined;
+}
+
+export const createAverageValueInsightMatrixValueCalculator = (): InsightMatrixValueCalculator => {
+    return {
+        calculate: (_: string[], values: number[]) => {
+            const nonZeroValues = values.filter(v => v > 0);
+            return nonZeroValues.length > 0 ? nonZeroValues.reduce((total, current) => total + current, 0) / nonZeroValues.length : undefined;
```
reneefromhold, line 9: Something about filtering out zero values feels off.  bc in normal circumstances, the count of all values makes up the denominator. Can this be renamed to include the fact that it ignores non zeros? thinking about consuming this downstream, it could create confusion.

greinard: The reason the non-zero values are being removed here is because some device types like Fitbit report a value whether you wear your device or not.  Steps, for example, gets reported as zero for each day you didn't wear it.  It doesn't mean that you walked zero steps that day.  It just means it didn't record any.  If we include the zero values, it can greatly affect the averages and not accurately reflect how much you actually walk each day.

I think zero is probably not a valid value for the majority of device data types, since most devices don't report a value at all when it is zero, but I also don't have a problem with renaming this implementation to be very clear that it will filter out the zero values when computing the average.

greinard: Done.  🙂

reneefromhold (approved):

greinard:
Thanks @reneefromhold.

Issue #366, opened 2024-11-27
Combined daily data type availability checkers should use the data collection settings.


greinard:
Currently, most/all combined daily data type availability checkers are not looking at the data collection settings to see if a given data source (such as Garmin) is enabled for the project.  This can result in a bunch of server side errors being logged even while under normal operating conditions.

For example, if you use a `DailyDataChart` with "combined steps", but don't have Garmin enabled for your project, because you are only interested in Apple Health and Fitbit step data, then you will cause errors to be logged every time it loads and the Garmin availability check runs.

I think the non-combined availability checkers are probably okay as is.  They could check the data collection settings as well, but I think it is less of an issue, because in the case of device specific data types, you probably wouldn't ever select them unless you also have them enabled for your project.

jjyutm4532:
Hello, I tried to solve the issue.

This is what I did:

Modified the combined daily data type availability checkers in the `src/helpers/daily-data-types/combined.tsx` file to check the data collection settings before performing availability checks for specific data sources. This change prevents unnecessary server-side errors when certain data sources are not enabled for the project.

You can review changes in this commit: https://github.com/jjyutm4532/CareEvolution-MyDataHelpsUI-366/commit/5268ecc8a15579ba9e550debaca3c0768e02e6a3.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

PR #371, opened 2024-12-09
Update the combined daily data provider availability checkers to use data collection settings.

greinard:
## Overview

Fixes: https://github.com/CareEvolution/MyDataHelpsUI/issues/366

This branch updates the combined daily data provider availability checkers such that they use data collection settings to avoid querying data sources that are not enabled.

## Security

- No new security concerns.  This update just makes the availability checkers perform the same checks that the actual data providers already use to avoid querying data sources which are not enabled.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

n/a

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the availability checks for various daily data types, enhancing reliability.
  
- **Refactor**
	- Updated the `availabilityCheck` functions to utilize `async/await` for better readability and control flow.

### DIFF
--- a/src/helpers/daily-data-types/combined.tsx
+++ b/src/helpers/daily-data-types/combined.tsx
@@ -6,28 +6,26 @@ import { defaultFormatter, heartRateFormatter, minutesFormatter, sleepYAxisConve
 import combinedRestingHeartRate from "../daily-data-providers/combined-resting-heart-rate";
 import { combinedSleepDataProvider, combinedStepsDataProvider, combinedMindfulMinutesDataProvider, combinedTherapyMinutesDataProvider } from "../daily-data-providers";
 import { simpleAvailabilityCheck } from "./availability-check";
+import MyDataHelps from "@careevolution/mydatahelps-js";
 
 let combinedTypeDefinitions: DailyDataTypeDefinition[] = [
     {
         dataSource: "Unified",
         type: DailyDataType.RestingHeartRate,
         dataProvider: combinedRestingHeartRate,
-        availabilityCheck: function (modifiedAfter?: Date) {
-            return simpleAvailabilityCheck("AppleHealth", ["RestingHeartRate"])(modifiedAfter).then(function (result) {
-                if (!result) {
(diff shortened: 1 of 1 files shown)


src/helpers/daily-data-types/combined.tsx
```diff
@@ -92,12 +84,15 @@ let combinedTypeDefinitions: DailyDataTypeDefinition[] = [
...
-            } else {
-                return result;
+            const settings = await MyDataHelps.getDataCollectionSettings();
+
+            if (settings.queryableDeviceDataTypes.find(dt => dt.namespace == "AppleHealth" && dt.type == "MindfulSession")
+                && await simpleAvailabilityCheck("AppleHealth", "MindfulSession")(modifiedAfter)) {
+                return true;
             }
+
+            return !!settings.queryableDeviceDataTypes.find(dt => dt.namespace == "GoogleFit" && dt.type == "ActivitySegment")
```
ajmenca, line 94: Here and on line 114 you did an explicit boolean coercion, but nowhere else that you're doing the same type of `find` call.  Is there something different about these two spots?

greinard: Yes.  This confused me at first, too (when the IDE suggested it).  It has to do with how the value from the `&&` evaluation works with ifs vs returns.

The resulting value from `obj && bool` is:
- `obj` when `obj` is falsy  (this is the important part)
- `bool` when `obj` is truthy

Since `obj` will only be falsy when it is `undefined`, the returned data type from the `&&` condition is `undefined | boolean`.  This works fine within the `if` conditions, because that `undefined | boolean` value gets re-evaluated for truthy-ness.  However, when it is being used as a return value that needs to be typed as a `boolean`, it doesn't match.  The `!!` operator forces it to be a `bool && bool` condition, which has the correct data type for the return statements.

ajmenca: Ah! Appreciate the explanation.  Makes sense - I didn't think to delete the `!!` locally in my IDE and see what happened because I figured it was a style choice and not a compilation one!

PR #376, opened 2024-12-19
NotificationList: Add identifier filtering, display limit, and hide when empty.

greinard:
## Overview

This branch updates the `NotificationList` component with some additional optional configuration properties.

- `notificationIdentifierPattern`: An optional regex used to filter the queried notifications by identifier.
- `displayLimit`: An optional limit on the number of displayed notifications.
- `hideWhenEmpty`: An optional flag to indicate that the list should be hidden when empty instead of showing the empty text.

## Security

- No new security risk.  These new properties just adjust the rendering behavior of an existing component.  No new data is being made available.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

### DIFF
--- a/src/components/container/NotificationList/NotificationList.tsx
+++ b/src/components/container/NotificationList/NotificationList.tsx
@@ -1,91 +1,102 @@
-import React, { useState, useEffect, useRef } from 'react'
-import MyDataHelps,  { Guid, Notification, NotificationQueryParameters, NotificationType } from "@careevolution/mydatahelps-js"
-import { LoadingIndicator, SingleNotification, Card, TextBlock } from '../../presentational'
-import { previewNotifications } from './NotificationList.previewdata'
+import React, { useState } from 'react';
+import MyDataHelps, { Guid, Notification, NotificationQueryParameters, NotificationType } from '@careevolution/mydatahelps-js';
+import { Card, LoadingIndicator, SingleNotification, TextBlock } from '../../presentational';
+import { previewNotifications } from './NotificationList.previewdata';
 import language from '../../../helpers/language';
-import OnVisibleTrigger from '../../presentational/OnVisibleTrigger/OnVisibleTrigger'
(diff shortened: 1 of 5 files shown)


src/components/container/NotificationList/NotificationList.tsx
```diff
@@ -1,91 +1,102 @@
...
+    hideWhenEmpty?: boolean;
+    innerRef?: React.Ref<HTMLDivElement>;
 }
 
-export type NotificationListPreviewState = "Default" | "NoData";
+/**
+ * A list of participant notifications.
+ *
+ * <ul>
+ * <li>The list can be filtered by type (i.e. Email/Push/SMS) and/or identifier pattern.</li>
```
ajmenca, line 24: Just double checking - does "pattern" imply regular expression in this type of usage?  I don't have a good sense as to how interchangable the phrase "pattern" and "regular expression" are.

ajmenca (approved):
Functionally, everything looks great - just one opinion question.

src/components/container/NotificationList/NotificationList.tsx
greinard: I think `pattern` is a fairly common terminology to use when referring to regular expressions.

Some examples here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/RegExp#return_value

However, I'm also fine with renaming to something like `notificationIdentifierRegex`, if you think that would be better understood.  Originally, I was going to have that property be a `string`, but then the component would have needed to handle invalid regex strings.  I figured it was better to leave that to the calling code so it could choose how to handle invalid regexes.  🙂

greinard: I went ahead and renamed it to `notificationIdentifierRegex` to avoid any confusion.

PR #379, opened 2025-01-03
Add a new combined steps daily data provider that includes Google Fit steps.

greinard:
## Overview

This branch adds an option to include Google Fit steps in the combined steps daily data provider.  Google Fit steps were not originally included in the combined provider for performance reasons since Google Fit only supports granular steps rather than hourly or daily aggregates like the other sources.

I have added a separate combined daily data type definition that can be used when Google Fit steps should also be included.

I also fixed a couple of the combined daily data provider availability checkers that were querying for the wrong types.

## Security

- No new security risk. Google Fit steps can already be queried.  This update just allows for including that data in the combined provider.

### Testing

- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

- No updates necessary.

### DIFF
--- a/src/helpers/strings-de.ts
+++ b/src/helpers/strings-de.ts
@@ -56,6 +56,7 @@ let strings: { [key: string]: string } = {
     "support": "Support",
     "all-notifications": "Alle Benachrichtigungen",
     "steps": "Schritte",
+    "steps-with-google-fit": "Schritte (inkl. Google Fit)",
     "resting-heart-rate": "Ruhepuls",
     "distance-traveled": "Zurückgelegte Entfernung",
     "google-fit-share": "Tippen Sie hier, um Fitnessdaten zu teilen",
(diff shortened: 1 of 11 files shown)


src/helpers/strings-de.ts
```diff
@@ -56,6 +56,7 @@ let strings: { [key: string]: string } = {
     "support": "Support",
     "all-notifications": "Alle Benachrichtigungen",
     "steps": "Schritte",
+    "steps-with-google-fit": "Schritte (inkl. Google Fit)",
```
chrisnowak, line 59: This is a participant-facing label actually, so it should still just be "Steps" actually...since they do not care

chrisnowak (approved):
Looks good, just one comment about the label

src/helpers/strings-de.ts
greinard: Ah.  I forgot about that.  We also use this label in the VB data type select dropdown, which is why I was making it different.

I'll update this to use the existing "Steps" label and then add an override of some kind in VB to handle that situation.

PR #386, opened 2025-01-15
Add support for AppleHealthStepsWhileWearingDevice and Custom (combined) data types.

greinard:
## Overview

This branch adds a new daily data type for Apple Health steps that has some special filtering logic.  It also adds support for a new "Custom" daily data type mechanism.

Changes include:
- Updated the `InsightMatrix` component (and some of its supporting code) to use string daily data types and the shared `queryDailyData` helper function.
- Added a new Apple Health steps daily data provider that returns steps only from days when a watch was also worn.
- Added support a new "Custom" daily data type.
  - This new data type will allow us to configure prioritized combined daily data types via the View Builder without having to hard code their implementations.
  - The new data type string has the following format: `Custom:["FitbitSteps","AppleHealthSteps"]`.
  - The subtypes are queried in priority order to find data for each day.  That is, for each day in the query date range, the value is used from the first subtype that has one.
  - The formatter, label key, preview data range, etc, are used from the first subtype only.
  - Care should be taken when using the custom type to ensure that all subtypes return the same kind of data.  There is nothing to prevent you from using `Custom:["FitbitSteps","FitbitSleep"]`, but it will attempt to render the sleep data as steps (based on the spec from the first subtype).
  - I have updated the `checkDailyDataAvailability`, `queryDailyData`, `queryPreviewDailyData`, and `getDailyDataTypeDefinition` helper functions to support this new functionality.
  - Going forward, while daily data types can (and usually are) set using the `DailyDataType` enum, the properties that hold those values should be of type `string` to allow for use of this new custom type.  This was already the case for most of the existing code, and the rest were updated as part of this PR.

## Security

- No new security risk.  This PR just adds some new ways to query the existing daily data API's for data that was already available.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

- You can adjust the daily data types being used in a number of existing stories to try out the new custom data type.


src/helpers/query-daily-data.tsx
```diff
@@ -32,29 +32,50 @@ export function registerDailyDataTypeDefinition(typeDefinition: DailyDataTypeDef
...
-export function checkDailyDataAvailability(type: string, modifiedAfter?: Date) {
-	var dailyDataType = dailyDataTypes.get(type);
-	if (!dailyDataType) { throw "Unknown data type:" + type; }
-	return dailyDataType.availabilityCheck(modifiedAfter);
+function parseDailyDataTypes(type: string) : DailyDataTypeDefinition[] {
+	const findDailyDataType = (type: string) : DailyDataTypeDefinition => {
+		const dailyDataType = dailyDataTypes.get(type);
+		if (!dailyDataType) {
+			throw "Unknown data type: " + type;
+		}
```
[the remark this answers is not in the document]
greinard: Done.

src/helpers/daily-data-types.tsx
```diff
@@ -16,6 +16,7 @@ export enum DailyDataType {
 	AppleHealthInBedMinutes = "AppleHealthInBedMinutes",
 	AppleHealthStandMinutes = "AppleHealthStandMinutes",
 	AppleHealthSteps = "AppleHealthSteps",
+	AppleHealthStepsWhileWearingWatch = "AppleHealthStepsWhileWearingWatch",
```
kneumei, line 19: Maybe `AppleHealthStepsWhileWearingDevice` just because it could be some other kind of wearable that triggers it to count.

greinard: Good idea.  Done.

kneumei (approved):

PR #388, opened 2025-01-17
Fix Recharts Bar graph edge case

ajmenca:
## Overview

Recharts has trouble rendering bar graphs on number based x-axis lines when there's only one data point.  It fails to properly calculate the width of the bar (https://github.com/recharts/recharts/issues/3640) as well as the gap between the x-axis and the first tick.  

While Recharts has provided some theoretical tools to help (letting you manually set the width, or manually set the padding), it turns out that for this edge case a trivial solution is to pad the data with a null data point at an expected interval.  

image
image

There's a very minor visual artifact of this if you're on desktop and can hover the chart; you get a line where it attempts to put a tooltip for the data point, but no tooltip displays.  This is the same artifact that you get on a line chart where we need to enforce a "gap" in the data and not have the line connect.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/container/SurveyAnswerChart/SurveyAnswerChart.stories.tsx
+++ b/src/components/container/SurveyAnswerChart/SurveyAnswerChart.stories.tsx
@@ -126,6 +126,59 @@ export const ffwelBarChart = {
     render: render
 };
 
+export const ffwelBarChartWithOneDataPoint = {
+    args: {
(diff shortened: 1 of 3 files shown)


greinard (approved):
LGTM :ship:

PR #391, opened 2025-01-20
Fix Error with Undefined Immunization MedicationName

chrisnowak:
## Overview

The current EHR news feed API will apparently return immunizations with an empty "MedicationName", and this was causing an error because we were filtering/searching on MedicationName, assuming it would always be there.

This just hides any immunizations with an empty MedicationName because there would be nothing to display.  Further investigation is required to figure out whether there is something additional we need to be displaying here.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/view/DeviceActivityView/DeviceActivityView.tsx
```diff
@@ -24,7 +24,7 @@ export default function DeviceActivityView(props: DeviceActivityViewProps) {
             showCloseButton={props.presentation == "Modal"}
             colorScheme={props.colorScheme}>
             <SegmentedControl
-                style={{ marginLeft: "16px", marginRight: "16px", marginBottom: selectedView == "daily" ? "16px" : "0px" }}
+                style={{ marginTop: !props.presentation ? "16px" : "0px", marginLeft: "16px", marginRight: "16px", marginBottom: selectedView == "daily" ? "16px" : "0px" }}
```
chrisnowak, line 27: This also adds a small tweak to DeviceActivityView so the segmented control isn't pushed up against the top of the screen if there's no "presentation" arg

brandoncharnesky (approved):
Tested locally, looks good.  Immunization is not showing if up MedicationName is missing or blank.

PR #392, opened 2025-01-23
Provider Search - Request to add new provider or health plan

reneefromhold:
## Overview

If you are at the end of your search results (i.e. there are no more pages, including if there are 0 results), there should be an action to request your provider if one is not found. The action should say "Request to add provider or health plan" and selecting it should go to this web page:
https://support.mydatahelps.org/hc/en-us/requests/new?ticket_form_id=34288897775635

Reqs:

- [ ] The link should open embedded (e.g. via openembeddedurl), not leaving the app
- [ ] There should be a flag to disable this behavior if someone doesn't want it, but by default it should be enabled
- [ ] We should be careful to only show it when we are finished loading the last page - it shouldn't show in any "loading" state for instance

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/container/ProviderSearch/ProviderSearch.tsx
```diff
@@ -163,6 +173,9 @@ export default function (props: ProviderSearchProps) {
                         </div>
                     </UnstyledButton>
                 )}
+                {!searching &&
+                    !props.hideAddProviderOrHealthPlanButton &&
+                    <Action onClick={addNewProviderAction} title={language("request-add-provider")} />}
```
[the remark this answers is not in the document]
chrisnowak: Actually - i'm not sure I even see that rule being used - maybe it is just old and can be removed

src/helpers/strings-en.ts
```diff
@@ -25,6 +25,7 @@
     "connect-ehr-text-connected": "Connect another provider to share more data or manage your EHR connections.",
     "connect-ehr-not-enabled": "EHR linking is not enabled for this project.",
     "search-for-provider": "Search for Provider or Health Plan",
+    "request-add-provider": "Request to add provider or health plan",
```
chrisnowak, line 28: We should likely make this title dynamic based on whatever categories are configured for the provider search - e.g. "Request to add health plan" if it's just health plan, "Request to add provider" if it's just provider.  I think there's an example of us constructing a string like that in the ExternalAccountsPreview component

src/components/container/ProviderSearch/ProviderSearch.tsx
```diff
@@ -14,20 +14,24 @@ export interface ProviderSearchProps {
     providerCategories?: string[];
     onProviderConnected?: (provider: ExternalAccountProvider) => void;
     innerRef?: React.Ref<HTMLDivElement>
-    connectExternalAccountOptions?: ConnectExternalAccountOptions
+    connectExternalAccountOptions?: ConnectExternalAccountOptions,
+    hideAddProviderOrHealthPlanButton?: boolean
```
chrisnowak, line 18: Prop name should probably be "hideRequestProviderButton" - the button is to request us adding one to our list - could be confusing to use "add" as one could conflate that with the user adding the account they found

chrisnowak:
Looks good overall - just those 3 issues I commented on

src/components/container/ProviderSearch/ProviderSearch.tsx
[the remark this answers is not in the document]
reneefromhold: That cleaned things up nicely, thanks for finding that!

src/components/container/ProviderSearch/ProviderSearch.tsx
reneefromhold: @chrisnowak I used a separator vs "or".  
I'm concerned concatenating the strings together vs translating the 3 potential values will create an output that doesn't align with how it would be naturally spoken
Do you have any thoughts with keeping it as is with the "/" or using "or" making 3 dedicated translations for
1. Request to add Providers
2. Request to add Health Plans 
3. Request to add Providers or Health Plans 

Not sure if this has come up before

src/helpers/strings-en.ts
reneefromhold: @chrisnowak I used a separator vs "or".
I'm concerned concatenating the strings together vs translating the 3 potential values will create an output that doesn't align with how it would be naturally spoken
Do you have any thoughts with keeping it as is with the "/" or using "or" making 3 dedicated translations for

Request to add Providers
Request to add Health Plans
Request to add Providers or Health Plans
Not sure if this has come up before

chrisnowak: I think "/" is fine, "Request to add Providers / Health Plans" seems fine to me

chrisnowak (approved):

PR #397, opened 2025-01-28
Added a couple more options to ProviderSearch

vladtar:
Added:
* `onProviderSelected()` property which can override default behavior (start the connection process) when you select a provider
* `publicProviderSearchApiUrl` property which can be used to query our public provider search endpoint (instead of the default authenticated one)

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@careevolution/mydatahelps-ui",
-  "version": "2.61.0",
+  "version": "2.61.1-WelcomeScreenProviderSearch.0",
   "description": "MyDataHelps UI Library",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
--- a/src/components/container/ProviderSearch/ProviderSearch.stories.tsx
+++ b/src/components/container/ProviderSearch/ProviderSearch.stories.tsx
@@ -49,7 +49,23 @@ export const HealthPlanOnly: Story = {
 export const onProviderSelected: Story = {
 	args: {
 		previewState: "Default",
-		onProviderConnected: (provider: ExternalAccountProvider) => alert(`You selected ${provider.name}`)
+		onProviderSelected: (provider: ExternalAccountProvider) => alert(`You selected ${provider.name}`)
+	},
+	render: render
+}
+
+export const onProviderConnected: Story = {
+	args: {
+		previewState: "Default",
+		onProviderConnected: (provider: ExternalAccountProvider) => alert(`You connected to ${provider.name}`)
+	},
+	render: render
+}
+
+export const publicEndpoint: Story = {
+	args: {
+		publicProviderSearchApiUrl: "https://designer.mydatahelps.dev/api/fhirpublicproviders",
+		onProviderSelected: (provider: ExternalAccountProvider) => alert(`You selected ${provider.name}`)
 	},
 	render: render
 }
@@ -72,7 +88,7 @@ export const Searching: Story = {
 export const Live: Story = {
 	args: {
 		previewState: "Default",
-		onProviderConnected: (provider: ExternalAccountProvider) => alert(`You selected ${provider.name}`)
+		onProviderConnected: (provider: ExternalAccountProvider) => alert(`You connected to ${provider.name}`)
 	},
 	render: render
 }
--- a/src/components/container/ProviderSearch/ProviderSearch.tsx
+++ b/src/components/container/ProviderSearch/ProviderSearch.tsx
@@ -12,10 +12,12 @@ import { FontAwesomeSvgIcon } from 'react-fontawesome-svg-icon';
 export interface ProviderSearchProps {
     previewState?: ProviderSearchPreviewState;
     providerCategories?: string[];
+    onProviderSelected?: (provider: ExternalAccountProvider) => void;
     onProviderConnected?: (provider: ExternalAccountProvider) => void;
-    innerRef?: React.Ref<HTMLDivElement>
-    connectExternalAccountOptions?: ConnectExternalAccountOptions,
-    hideRequestProviderButton?: boolean
+    innerRef?: React.Ref<HTMLDivElement>;
+    connectExternalAccountOptions?: ConnectExternalAccountOptions;
+    hideRequestProviderButton?: boolean;
+    publicProviderSearchApiUrl?: string;
 }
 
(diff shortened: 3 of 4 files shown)


src/components/container/ProviderSearch/ProviderSearch.tsx
```diff
@@ -12,10 +12,12 @@ import { FontAwesomeSvgIcon } from 'react-fontawesome-svg-icon';
...
     previewState?: ProviderSearchPreviewState;
     providerCategories?: string[];
+    onProviderSelected?: (provider: ExternalAccountProvider) => void;
     onProviderConnected?: (provider: ExternalAccountProvider) => void;
-    innerRef?: React.Ref<HTMLDivElement>
-    connectExternalAccountOptions?: ConnectExternalAccountOptions,
-    hideRequestProviderButton?: boolean
+    innerRef?: React.Ref<HTMLDivElement>;
+    connectExternalAccountOptions?: ConnectExternalAccountOptions;
+    hideRequestProviderButton?: boolean;
```
reneefromhold, line 19: @vladtar How are the providers here different from what is returned from getExternalAccountProviders? 
Is this something we anticipate component users to be customizing regularly?

src/components/container/ProviderSearch/ProviderSearch.tsx
```diff
@@ -12,10 +12,12 @@ import { FontAwesomeSvgIcon } from 'react-fontawesome-svg-icon';
 export interface ProviderSearchProps {
     previewState?: ProviderSearchPreviewState;
     providerCategories?: string[];
+    onProviderSelected?: (provider: ExternalAccountProvider) => void;
```
reneefromhold, line 15: Can the new and existing parameter have some documentation added? It's not apparent that one will be ignored when the other is supplied. Adding documentation is a good middle ground. Storybook will pull the documentation in automatically and show on the main docs page

export interface ProviderSearchProps {
    previewState?: ProviderSearchPreviewState;
    providerCategories?: string[];
    /** @param props.onProviderSelected - Callback function triggered when a provider is selected. If this function is defined, onProviderConnected will not be called.*/ 
    onProviderSelected?: (provider: ExternalAccountProvider) => void;
    /** @param props.onProviderConnected - Callback function triggered after a provider is connected. This will be ignored when onProviderSelected is supplied. */
    onProviderConnected?: (provider: ExternalAccountProvider) => void;
    innerRef?: React.Ref<HTMLDivElement>;
    connectExternalAccountOptions?: ConnectExternalAccountOptions;
    hideRequestProviderButton?: boolean;
    publicProviderSearchApiUrl?: string;
}

src/components/container/ProviderSearch/ProviderSearch.stories.tsx
```diff
@@ -72,7 +88,7 @@ export const Searching: Story = {
 export const Live: Story = {
 	args: {
 		previewState: "Default",
```
reneefromhold, line 90: Not something you added, but Live has no preview state/data. Please delete this arg

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@careevolution/mydatahelps-ui",
-  "version": "2.61.0",
+  "version": "2.61.1-WelcomeScreenProviderSearch.0",
```
reneefromhold, line 3: Note to revert before merge

src/components/container/ProviderSearch/ProviderSearch.tsx
```diff
@@ -69,13 +71,31 @@ export default function ProviderSearch(props: ProviderSearchProps) {
...
+            MyDataHelps.getExternalAccountProviders(search, props.providerCategories?.length == 1 ? props.providerCategories[0] : null, pageSize, currentPage).then(function (searchResultsResponse) {
+                if (requestID == currentRequestID) {
+                    updateSearchResults(searchResultsResponse.externalAccountProviders);
+                    setTotalResults(searchResultsResponse.totalExternalAccountProviders);
+                    setSearching(false);
+                }
+            });
+        }
+        else {
+            fetch(`${props.publicProviderSearchApiUrl}?keyword=${search}&pageSize=${pageSize}&pageNumber=${currentPage}`, {
```
bxh, line 84: Not only it is more readable, but also it handles parameter encoding for you.
```suggestion
      const url = new URL(props.publicProviderSearchApiUrl);
      url.searchParams.append('keyword', search);
      url.searchParams.append('pageSize', String(pageSize));
      url.searchParams.append('pageNumber', String(page));
      fetch(url.toString(), {
        method: "GET",
        headers: { "Accept": "application/json" },
      });
```

src/components/container/ProviderSearch/ProviderSearch.tsx
chrisnowak: @reneefromhold added this...it is probably more accurate to say that onProviderSelected overrides the normal action taken when a provider is selected.

src/components/container/ProviderSearch/ProviderSearch.tsx
chrisnowak: I don't think component users will really be customizing this much; it has a fairly narrow use case of just the new welcome screen.

We might add more to the existing API or customize it based on user data in the future, so I don't think it makes sense to have it always use the public one

src/components/container/ProviderSearch/ProviderSearch.tsx
chrisnowak: Done

chrisnowak:
I took over this PR from @vladtar and added/tweaked a few things:
- I added the concept of "Featured Providers" which you can pass in to populate search results when the user has not entered a search term; this is primarily to be able to populate very popular providers at the top.  
  - This is a bit of a hack; ideally the search endpoint itself would return these providers first, but this works for now and we can revisit later
- I slightly adjusted the CSS of the providers to give the logo and each row just a *bit* more room; which I think makes a nice aesthetic difference
- Unrelated to provider search, I added override handling to the navigation bar for onBack() and onCancel().  While I could have added custom left/right nav items, that's actually kind of painful and requires styling those, and this is a pretty easy/straightforward option

chrisnowak (approved):

vladtar (commented):
Looks good to me. 🚢

PR #409, opened 2025-02-11
"days with data" single value provider

kneumei:
## Overview

This adds a new single value data provider called "Days with Data". The intention is to use it for the "Goal" component in order to show how many days a participant has had wearable activity (steps, etc) in the last x days. 

## Security

- [ x These changes do not introduce any security risks, or any such risks have been properly mitigated.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/helpers/single-value-providers.ts
+++ b/src/helpers/single-value-providers.ts
@@ -76,3 +76,7 @@ export const createRandomIntegerValueProvider = (maxValue: number): RandomIntege
 export const createCustomFieldIntegerValueProvider = (customField: string): CustomFieldIntegerValueProvider => {
     return new CustomFieldIntegerValueProvider(customField);
 };
+
+export const createDaysWithDataValueProvider = (dataType: string, daysInPast: number): DaysWithDataValueProvider => {
+    return new DaysWithDataValueProvider(dataType, daysInPast);
+};
--- a/src/helpers/single-value-providers.ts
+++ b/src/helpers/single-value-providers.ts
@@ -54,6 +54,9 @@ export class CustomFieldIntegerValueProvider extends SingleValueProvider<number>
 
 export class DaysWithDataValueProvider extends SingleValueProvider<number> {
     constructor(public dataType: string, public daysInPast: number) {
+        if(daysInPast < 0) {
+            throw new Error('daysInPast must be a non-negative number');
+        }
         super('days with data');
     }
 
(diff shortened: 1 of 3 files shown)


src/helpers/single-value-providers.ts
```diff
@@ -50,6 +52,25 @@ export class CustomFieldIntegerValueProvider extends SingleValueProvider<number>
...
+    constructor(public dataType: string, public daysInPast: number) {
+        if(daysInPast < 0) {
+            throw new Error('daysInPast must be a non-negative number');
+        }
+        super('days with data');
+    }
+
+    async getValue(): Promise<number | undefined> {
+        const today = new Date();
+        const startDate = startOfDay(subDays(today, this.daysInPast));
```
greinard, line 65: I'm not sure it is 100% necessary, but you may want to ensure that both the start and end dates are "start of day" dates.  Maybe something like this:

```
const today = startOfToday();
const startDate = subDays(today, this.daysInPast);
```

src/helpers/single-value-providers.ts
```diff
@@ -50,6 +52,25 @@ export class CustomFieldIntegerValueProvider extends SingleValueProvider<number>
...
+        const data = await queryDailyData(this.dataType, startDate, today);
+        if(!data || data.length === 0) {
+            return 0;
+        }
+        return new Set(Object.keys(data)).size;
```
greinard, line 70: I think you will want to count the non-zero values here.  The key count will just be the days it computed values for, which will probably be all non-zero values, but I'm not sure we enforce that.  Some daily data types could return `0` for unvalued days.

`return Object.keys(data).reduce((daysWithData, dayKey) => daysWithData + (data[dayKey] > 0 ? 1 : 0), 0);`

kneumei: ah good catch!

kneumei:
@greinard changes made. LMK if it looks good

src/helpers/single-value-providers.ts
```diff
@@ -50,6 +52,25 @@ export class CustomFieldIntegerValueProvider extends SingleValueProvider<number>
...
+export class DaysWithDataValueProvider extends SingleValueProvider<number> {
+    constructor(public dataType: string, public daysInPast: number) {
+        if(daysInPast < 0) {
+            throw new Error('daysInPast must be a non-negative number');
+        }
+        super('days with data');
+    }
+
+    async getValue(): Promise<number | undefined> {
+        const today = startOfDay(new Date());
```
greinard, line 64: This is fine, but just wanted to mention that `startOfToday()` returns the same date as this does.

Issue #404, opened 2025-02-05
Add support for de, fr, pt-PT, it, pl


basyl-durnan:
# About

https://github.com/CareEvolution/deploy-myfhr/issues/4836

# MDH UI and View Utils - Locale Definition

This creates the translations for the MDH UI components and view builder.

## Code Changes

In the [**MyDataHelpsUI** repo](https://github.com/CareEvolution/MyDataHelpsUI):

1. Copy the [English language file](https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/helpers/strings-en.ts) to make one for the new language.
2. Send out for translations.
3. Once merged, deploy a new NPM package with the updated strings.
4. Once the new MDH UI version is published, update [MDH View Builder](https://github.com/CareEvolution/MyDataHelpsViewBuilder) to reference it.
5. Ditto for the [MDH View Library](https://github.com/CareEvolution/MyDataHelpsViewLibrary).

Note: The view builder main branch is deployed to QA automatically. Once tested, another PR is necessary to merge to the `prod` branch and deploy to prod. This can be done outside a release cycle.

## Testing

From the [Localization Test Plan](https://docs.google.com/document/d/1blcMJfHzE3PJ7iuL_jBA7nkoafRsjhZAT5JB7pPetww/edit?usp=sharing):

* TBD (need new test cases for this)

basyl-durnan:
Dev PR: https://github.com/CareEvolution/MyDataHelpsUI/pull/412

Incomplete however—we don't seem to currently support regional variations of a language.

basyl-durnan:
@chrisnowak Do you think we are okay to skip "5. Ditto for the [MDH View Library](https://github.com/CareEvolution/MyDataHelpsViewLibrary)." if we are deprecating the View Library?

basyl-durnan:
Noting the conclusion of Slack discussions: We _will_ translate the View Library.

lynnfaraday:
These languages are in MDH UI and ViewBuilder in prod, and have been tested by Dan C and others as part of the 2025.3 localization testing: https://docs.google.com/spreadsheets/d/1uiXl-SkvtxgDFGacp3jfkmVSbxpIbddPUl2nYpDyy5E/edit?usp=sharing

PR #412, opened 2025-02-12
Add pt-PT

basyl-durnan:
## Overview

See https://github.com/CareEvolution/MyDataHelpsUI/issues/404

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


basyl-durnan:
@nathia wondering if you might be able to assist. The `language.ts` file only seems to look at the first 2 letters of a language, so it ignores the region. How can we adjust it so that we support and deliver pt-PT strings?

mmertsock:
> @nathia wondering if you might be able to assist. The language.ts file only seems to look at the first 2 letters of a language, so it ignores the region. How can we adjust it so that we support and deliver pt-PT strings?

I noticed the same thing in the related [MyDataHelpsWelcomeScreen PR](https://github.com/CareEvolution/MyDataHelpsWelcomeScreen/pull/52#discussion_r1953214853), and made a suggested change, though language.ts here take a different approach so my suggestion isn't directly portable.

lynnfaraday:
@nathia Let's sort out the discussion on the welcome screen PR about what the approach should be, and then figure out how to  do the same thing here.

lynnfaraday:
Actually the fix needs to be made in MDH.JS - this is just using MyDataHelps.getCurrentLanguage(). Even if we set the SDK's language to pt-PT, it gives it back as just "pt"

src/helpers/language.ts
```diff
@@ -4,32 +4,41 @@ import spanishStrings from "./strings-es"
...
-	if (currentLanguage == "fr") resolvedString = frenchStrings[key];
-	if (currentLanguage == "pt") resolvedString = portugueseStrings[key];
-	if (currentLanguage == "it") resolvedString = italianStrings[key];
-	if (currentLanguage == "pl") resolvedString = polishStrings[key];
-	if (resolvedString != null) return format(resolvedString, args);
-
-	return format(englishStrings[key], args);
+export function language(key: string, specifiedLocale?: string, args?: { [key: string]: string }): string {
+	// Normalize formats like pt-PT or pt_PT to a consistent pt-pt for lookup.
+	const currentLocale = (specifiedLocale || MyDataHelps.getCurrentLanguage() || "").replace("_", "-").toLowerCase();
```
lynnfaraday, line 21: Need to clean this up a little and add some unit tests, but this should generally make pt-PT and any future country-specific strings tables work.

lynnfaraday:
@nathia / @mmertsock / @basyl-durnan. I updated the code here to handle specific country codes, if someone could take a look.

mmertsock (approved):
Latest changes LGTM and will properly handle the current bug in the iOS app until we deploy a new app release with the bug fixed

.storybook/preview.js
```diff
@@ -15,7 +15,8 @@ export const globalTypes = {
 				{ value: 'nl', title: 'Dutch' },
 				{ value: 'de', title: 'German' },
 				{ value: 'fr', title: 'French' },
-				{ value: 'pt', title: 'Portugese' },
+				{ value: 'pt', title: 'Portuguese (Brasil)' },
```
bxh, line 18: It is a bit counter intuitive to use `pt` to represent Brazilian Portuguese.

lynnfaraday: That's just because it's the default. It's this way across all the MDH apps.

`pt-BR` is Brazilian Portuguese and `pt-PT` is Portugal Portuguese. There are a few others.  `pt` is just the default Portuguese when the user doesn't specify a country code. Since most of our anticipated customers are in North America, we made Brazilian Portuguese the default.

This is similar to Spanish - Mexican Spanish would be `es-MX`, US Spanish `es-US`, Spain Spanish `es-ES`, etc. We chose Latin American Spanish as the default (`es`) so when we add Spain Spanish the mapping will be:

* es - Spanish (Latin American)
* es-es - Spanish (Spain)

basyl-durnan:
Based upon conversations over here, https://github.com/CareEvolution/MyDataHelpsWelcomeScreen/pull/52/files#r1960239036, sounds like we should be good to proceed after https://github.com/CareEvolution/MyDataHelpsUI/pull/412/commits/8dfc7cdddfc3cf3ddb066c5ab0794e1058309eed?

lynnfaraday:
@basyl-durnan Could you please revert your change - properly handling variants in case and hyphens/underscores is important, and you changed some of the test cases that made sure it worked right. It's fine for us internally to have a preferred style that WE use, but we've seen cases where different devices/browsers sometimes give us different inputs.

basyl-durnan:
Sure—just reverted.

nathia (approved):
@basyl-durnan,

This looks good. The fallback is the primary language when a region-specific local is not found. If the primary language is not found then we fall back to English. This is exactly how we intended it to work.

image

basyl-durnan:
Thanks for checking that. That suggests it is working as we would expect—we fallback to the language if we don't have region-specific and otherwise do English if we don't have the language.

basyl-durnan:
@lynnfaraday is this PR good to merge? 

And sounds like we will just make the same changes to the Welcome Screen (https://github.com/CareEvolution/MyDataHelpsWelcomeScreen/pull/52), then that one is good to go too?

lynnfaraday:
@basyl-durnan Codewise yeah, but I wasn't sure what the testing plan was or whether we were trying to coordinate the langauge changes.

basyl-durnan:
Ah, my intention with this PR was only to update dev, but I see there is no `prod` branch. Do we only use storybook for testing or how can we update dev and not prod? It wasn't immediately clear to me from the readme.

PR #413, opened 2025-02-13
Add Health Connect support to daily data types

bxh:
## Overview

This set of changes adds Health Connect support throughout the application for displaying daily data (specifically **steps**, **resting heart rate**, and various **sleep** metrics). It also includes several Storybook story updates and an .editorconfig file introduction. More data types can be added, but I’d like to gather some early feedback now.

- Data Providers
  - Added `health-connect-steps.ts` (for querying daily step counts)
  - Added `health-connect-resting-heart-rate.ts` (for querying average resting heart rate)
  - Added `health-connect-sleep.ts` (for querying total, light, deep, and REM sleep minutes)
  - Updated `combined-resting-heart-rate.ts`, `combined-steps.ts`, and `combined-sleep-v2.ts` to include Health Connect data if enabled in user settings.
- Availability Checks:
  - Introduced simpleAvailabilityCheckV2 to check for data presence in Health Connect, which requires V2 DeviceData endpoints.
- UI Components
  - HealthConnectMonthCharts: New container component that uses MonthCharts to display multiple Health Connect charts (steps, resting heart rate, sleep metrics) together.
  - RestingHeartRateCalendar: Updated to allow HealthConnect as a valid data source type, querying average resting heart rate from Health Connect.
- Relative Activity Logic
  - Modified queryRelativeActivity to handle potential errors (using Promise.allSettled) so a response error (e.g. 401) from one provider does not block others.

**Note:** I did not include `exercise` data due to lack of server-side support.

Screenshot 2025-02-13 135919

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

Tested using data generated by the[ Android Health Connect toolbox](https://developer.android.com/health-and-fitness/guides/health-connect/test/health-connect-toolbox) and synced to the QA environment.

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/helpers/daily-data-providers/health-connect-steps.ts
```diff
@@ -0,0 +1,25 @@
+﻿import { add } from "date-fns";
+import { DailyDataQueryResult } from "../query-daily-data";
+import queryAllDeviceDataV2Aggregates from "../query-all-device-data-v2-aggregates";
+
+export async function stepsProvider(
+    startDate: Date,
+    endDate: Date,
+): Promise<DailyDataQueryResult> {
+    const aggregates = await queryAllDeviceDataV2Aggregates({
```
chrisnowak, line 9: Don't think this is right; which is why I said we should skip steps for now until we have aggregate support on the Android side and can establish a "Daily Steps" rule

Multiple devices can write steps into Health Connect so you will get overlapping intervals; e.g. you can have 20 steps written from your phone from 10:15 - 10:30 and then 100 steps written from your smartwatch from 10:00-11:00

The correct number of steps in that instance is 100, not 120 (which is what this would provide I believe); and Health Connect aggregates (like statistics on iOS) do that magic to dedupe intervals appropriately.

src/helpers/daily-data-providers/combined-sleep-v2.ts
```diff
@@ -2,6 +2,7 @@ import { add } from 'date-fns';
 import { totalSleepMinutes as fitbitTotalSleepMinutesDataProvider } from './fitbit-sleep-v2'
 import { totalSleepMinutes as garminTotalSleepMinutesDataProvider } from './garmin-sleep-v2'
 import { totalSleepMinutes as appleHealthSleepMinutesDataProvider } from './apple-health-sleep-v2';
+import { totalSleepMinutes as healthConnectTotalSleepMinutesDataProvider } from './health-connect-sleep';
```
chrisnowak, line 5: I'm a bit confused why we have a v2 sleep data provider actually...this isn't exposed in any of the graphs etc nor does it have a DailyDataTypeDefinition

@greinard I think this might have been used for some glucose stuff?  And I imagine the difference is that it queries V2 data for Apple Health sleep vs. regular device data.  

We likely should not have 2 different ones here.  Seems like we could make this pick up on whether V2 Sleep data is available (there's an API for that I believe) and make the V1 provider smart enough to use that?

Either way I think we would want to incorporate health connect sleep into the combined v1 provider though, because that's the ones that are used in all the graphs and there is no way to access health connect data in v1.  And projects may not have V2 sleep enabled for Apple Health etc.

src/components/container/HealthConnectMonthCharts/HealthConnectMonthCharts.tsx
```diff
@@ -0,0 +1,91 @@
+import React from "react";
```
chrisnowak, line 1: Would drop this component; no reason to have a whole component targeted just to Health Connect.

src/helpers/relative-activity.tsx
```diff
@@ -27,30 +27,31 @@ export function queryRelativeActivity(startDate: Date, endDate: Date, dataTypes:
             add(startDate, { days: dataType.threshold == undefined ? -31 : -1 }),
             add(endDate, { days: 1 }),
             preview));
-    return Promise.all(promises).then((results) => {
+    return Promise.allSettled(promises).then((results) => {
```
chrisnowak, line 30: What's this change about?

src/helpers/daily-data-providers/combined-resting-heart-rate.ts
```diff
@@ -16,6 +21,9 @@ export default function (startDate: Date, endDate: Date) {
 		if (settings.queryableDeviceDataTypes.find(s => s.namespace == "AppleHealth" && s.type == "RestingHeartRate")) {
 			providers.push(appleHealthRestingHeartRateDataProvider(startDate, endDate));
 		}
+		if (settings.healthConnectEnabled) {
```
chrisnowak, line 24: We should also query MyDataHelps.getDeviceDataV2AllDataTypes() and check to make sure that resting heart rate is enabled for Health Connect before including it here

src/helpers/daily-data-providers/combined-sleep-v2.ts
```diff
@@ -18,6 +19,10 @@ export default async function (startDate: Date, endDate: Date): Promise<{ [key:
     if (settings.queryableDeviceDataTypes.find(s => s.namespace == "AppleHealth" && s.type == "SleepAnalysisInterval")) {
         providers.push(appleHealthSleepMinutesDataProvider(startDate, endDate));
     }
+    if (settings.healthConnectEnabled) {
```
chrisnowak, line 22: Should query MyDataHelps.getDeviceDataV2AllDataTypes() and check to make sure that sleep is enabled for Health Connect before including it here

src/helpers/daily-data-providers/combined-resting-heart-rate.ts
```diff
@@ -1,5 +1,10 @@
 ﻿import { add } from "date-fns";
-import { appleHealthRestingHeartRateDataProvider, fitbitRestingHeartRateDataProvider, garminRestingHeartRateDataProvider } from ".";
+import {
```
chrisnowak, line 2: We'll need to update the availability check for RHR/sleep combined as well: https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/helpers/daily-data-types/combined.tsx#L17

chrisnowak: Which should also incorporate getDeviceDataV2AllDataTypes()

chrisnowak:
I would definitely test this against a project that doesn't have V2 storage of sleep/RHR data enabled to make sure we have the appropriate guards in place to prevent actual errors

src/helpers/daily-data-providers/combined-sleep-v2.ts
chrisnowak: > Now that we can determine when the v2 sources are enabled, I think we could update the v1 providers to query the v2 data as well. I guess we'd just have them prefer v2 sources when enabled, and use v1 sources as the backup? The only catch here might be when v2 data types are first enabled. They don't get a historic backfill, so the v1 source might have access to much more data than v2. This shouldn't be a problem for new participants, but for existing ones, I think it could potentially be an issue.

I think this is fine as we're going to be setting this up when the project first starts anyways most of the time.  Maybe we can revisit if / when it becomes an issue

> Also, tangential to this, I think the querySleepMinutesV2(...) function in the common-sleep-v2.ts file should probably be updated to use the recently refactored de-duplication and day-boundary-splitting logic that was originally part of the v1 Apple Health sleep data provider. https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/helpers/daily-data-providers/apple-health-sleep.ts#L18-L19

Ah I was wondering about that, agreed...should be updated

lynnfaraday:
@bxh It doesn't seem there's any API doc impact to this change, since it's contained within MDHUI? Though maybe I'm missing something.

src/helpers/daily-data-providers/health-connect-steps.ts
chrisnowak: @bxh max heart rate would be nice, I was thinking about adding that to the main dashbaord.  But for MVP RHR/Sleep would be the key ones to focus on; and once we support client-side aggregates we can do steps

src/helpers/relative-activity.tsx
bxh: `allSettled` allows partial success. Failure from one request doesn't block another. There are some daily data providers that don't check data collection settings prior to requesting data.

src/helpers/daily-data-providers/combined-sleep-v2.ts
chrisnowak: Yes, I would add the v1 DailyDataProvider for HC sleep data per the above.  We may want to actually get rid of the v2 sleep data provider, we should ideally not have different v1 vs. v2 sleep data providers, as it complicates the story of what to pick

bxh: I don't like the separation between v1 and v2 either, as it is a burden on the users. I'm going to add HC to V1 DeviceDataNamespace then.

chrisnowak: To be clear it's just that we should have the normal combined sleep provider incorporate the V2 data.  On the back-end it's still split between v1 and v2, but the "combined sleep" data type would query both V1 and V2 and the Health Connect "Sleep" data type would query V2 still...all just hidden from users

bxh: Ah, that makes sense. There is no LDS entries for HC data.

bxh:
Replaced by https://github.com/CareEvolution/MyDataHelpsUI/pull/429

For some reason, this PR couldn't receive any updates.

PR #414, opened 2025-02-15
Update mindful and therapy minutes daily data providers to handle duplicated and/or overlapping time ranges.

greinard:
## Overview

This branch updates the Apple Health and Google Fit mindful and therapy minutes daily data providers such that they handle duplicated and overlapping time ranges.  Also, minutes are now split across day boundaries, when such a case exists (i.e. 10pm-2am gets split in half).

The code for this change was largely (almost completely) lifted from the existing `apple-health-sleep.ts` daily data provider, which had already implemented this de-duplication and day-boundary-splitting logic.  I just refactored that code into a couple of reusable helper functions, added a bunch of UTs, and updated the relevant daily data providers to use those new helper functions.

No functional changes were intended with regard to the Apple Health sleep data provider, with one exception.  The prior implementation used to round down to the minute for each range within a day before they were added together.  Now, the code waits until all of the ranges for a given day have been summed together before rounding down to the minute.  This keeps partial minute loss to a minimum:

```
Prior Logic = 3.9 mins + 2.9 mins + 4.9 mins = 3 mins + 2 mins + 4 mins = 9 mins
New Logic = 3.9 mins + 2.9 mins + 4.9 mins = 11.7 mins = 11 mins
```

## Security

- No new security risk.  This branch just updates existing data provider computations to avoid counting duplicate minutes.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

- N/A


src/helpers/date-range.ts
```diff
@@ -0,0 +1,106 @@
...
+    while (dayCutoff < observationDate) {
+        const nextDayCutoff = add(dayCutoff, { days: 1 });
+        if (nextDayCutoff > startDate) {
+            dateRanges.push({
+                startDate: (startDate >= dayCutoff) ? startDate : dayCutoff,
+                endDate: (observationDate < nextDayCutoff) ? observationDate : nextDayCutoff
+            });
+        }
+        dayCutoff = nextDayCutoff;
+    }
```
bxh:
```suggestion
    let left = startDate;
    const dateRanges: DateRange[] = [];
    while (left < observationDate) {
        const right = add(startOfDay(current), { days: 1, hours: offsetHours }); // Note that offsetHours is set here instead of by the caller.
        dateRanges.push({
            startDate: left,
            endDate:  observationDate < right ? observationDate : right
        }
        left = right;
    }
```

src/helpers/date-range.ts
```diff
@@ -0,0 +1,106 @@
...
+        const ranges = splitSampleIntoRanges(dataPoint, offsetHours);
+        for (const range of ranges) {
+            const anchorDate = add(range.startDate, { hours: -(offsetHours ?? 0) });
+            const dayKey = getDayKey(anchorDate);
+            if (!dayRanges[dayKey]) {
+                dayRanges[dayKey] = [];
+            }
+            dayRanges[dayKey].push(range);
+        }
+    });
```
bxh: You can instead sort data points once and merge in a single scan, reducing time complexity from O(n^2) to O(nlogn)

src/helpers/time-range.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    return ranges;
+}
+
+function rangesHaveOverlap(range1: TimeRange, range2: TimeRange): boolean {
+    return (range2.startTime >= range1.startTime && range2.startTime <= range1.endTime) ||
+        (range2.endTime >= range1.startTime && range2.endTime <= range1.endTime) ||
+        (range1.startTime >= range2.startTime && range1.startTime <= range2.endTime) ||
+        (range1.endTime >= range2.startTime && range1.endTime <= range2.endTime);
```
bxh, line 100:
```suggestion
  return range1.endDate > range2.startDate && range1.startDate < range2.endDate;
```

bxh (changes requested):
Plus for centralizing date range processing logic, but the implementation can be simplified for both performance and readability.

ajmenca (approved):
The renaming tweaks we discussed on Slack all look good to me.  Makes what the functions are doing more clear.

src/helpers/date-range.ts
```diff
@@ -0,0 +1,106 @@
...
+import getDayKey from './get-day-key';
+import { DailyDataQueryResult } from "./query-daily-data";
+
+export interface DateRange {
+    startDate: Date;
+    endDate: Date;
+}
+
+export function computeDayRanges(dataPoints: DeviceDataPoint[], offsetHours?: number): { [key: string]: DateRange[] } {
+    const dayRanges: { [key: string]: DateRange[] } = {};
```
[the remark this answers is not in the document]
greinard: This has been updated to the type `DailyTimeRanges`, but the definition of that type still uses `[key: string]`.

`export type DailyTimeRanges = { [key: string]: TimeRange[] };`

Is there a benefit to using `Record` here vs just the string key?

src/helpers/date-range.ts
```diff
@@ -0,0 +1,106 @@
...
+import { add, differenceInSeconds, parseISO, startOfDay } from 'date-fns';
+import getDayKey from './get-day-key';
+import { DailyDataQueryResult } from "./query-daily-data";
+
+export interface DateRange {
+    startDate: Date;
+    endDate: Date;
+}
+
+export function computeDayRanges(dataPoints: DeviceDataPoint[], offsetHours?: number): { [key: string]: DateRange[] } {
```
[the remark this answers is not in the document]
greinard: Done.

src/helpers/date-range.ts
greinard: I don't think this suggested code is correct.  It has a couple of syntax errors (i.e. closing parens of the `push(...)` call and use of a non-existent `current` variable).  The logic does not appear to be correct either.

For example, if `current` should have been set to `left` or the start date, which I think it would have to be on the first iteration, then `right` would be set to a value that is less than `left` if the start date is after the cutoff time when an offset is used.  For example, if start date is `2025-02-16T19:00:00-05:00`, and of the offset being used is -6 hours, then `left` would be set to `2025-02-16T19:00:00-05:00` (the start date) and `right` would be set to `2025-02-16T18:00:00-05:00` (the beginning of the start date, plus a day, minus 6 hours).  This also causes an infinite loop because `left` gets set to `right`, and then `right` keeps getting set to the same value over and over.  The `current` value would need to be incremented here somewhere.

bxh: I didn't have a compiler. :) `current` is actually `left.

src/helpers/date-range.ts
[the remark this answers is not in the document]
bxh: It is typically used for dynamic object keys with a specific value type, which fits this use case. https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type

src/helpers/date-range.ts
greinard: I'm not sure I agree with this.

The `ranges` here are the ranges from each single data point.  I think we want to accumulate all of the ranges for a given day (which can come from multiple data points) before sorting them.  We could have 100 data points across a 7 day window.  If we perform a per-data-point sort, we'd have a minimum of 100 sorts.  If we instead perform the sorts on the accumulated daily ranges, we'd have a maximum of 7 sorts.

If we wanted to incorporate the merging into a single pass loop here, I think we could.  We'd probably avoid the sorting altogether, since it would only apply across a single data point anyways, and instead, just check the existing collection of ranges from each day to decide if the new range should be merged with an existing one or added.  That would logically work fine, I think, but it would also mean searching the daily accumulated list of ranges for every range from every data point.  We'd also need to re-check all of the existing daily ranges any time we performed a merge, because a merge of existing range A with new range B could mean that merged range A' now overlaps with existing range C.  I think that would generally be less performant on average than the two pass approach.

I also think the two pass "identify then merge" approach is a tad bit easier to understand at a glance than the "identify and merge" approach would be.

src/helpers/time-range.ts
greinard: This is a nice simplification. 👍 

I did make one small adjustment to this, though, to work as it did previously (with inclusive start and end times):
`return range1.endTime >= range2.startTime && range1.startTime <= range2.endTime;`

I also had to draw this out on my whiteboard to really understand it.  All the UTs were passing, though, so I was fairly confident it was correct.  😃

greinard:
@coderabbitai pause

src/helpers/date-range.ts
greinard: This still results in an infinite loop in the scenario described above.  If start date is `2025-02-16T19:00:00-05:00` and the offset is `-6` because `right` (and subsequently, `left`) keeps getting set to `2025-02-16T18:00:00-05:00`.

I am open to alternative naming, but I feel like `left` and `right` are a little two generic for what is going on here.  It is an interval, but it isn't the resulting interval.  The `dayCutoff` and `nextDayCutoff` variables are meant to define the bounds of the current "day" (which can be offset from midnight) that we are evaluating data point dates against.  Perhaps something like `dayWindowStart` and `dayWindowEnd` could be used, but I think `left` and `right` just wouldn't clearly communicate the intent of these vars.

bxh: Got it. I didn't expect that `offsetHours` can be negative. Your current implementation is correct.

src/helpers/date-range.ts
bxh: OK, I see that you need to not only merge overlapping intervals, but also group by the day. So, it makes sense to sort within each group, like what you did.

But given that `computeDayRanges` and `computeMinutesResultFromDayRanges` are always coupled, I wonder if we can merge the them as follows:
```
export function computeDayDurationsInSeconds(
  dataPoints: DeviceDataPoint[],
  offsetHours: number = 0
): Record<string, number> {
  // Group ranges by dayKey.
  const dayMap: Record<string, DateRange[]> = {};

  for (const dp of dataPoints) {
    if (!dp.startDate || !dp.observationDate) continue;

    const ranges = splitSampleIntoRanges(dp, offsetHours);
    for (const range of ranges) {
      const dayKey = getDayKey(add(range.startDate, { hours: -offsetHours }));

      if (!dayMap[dayKey]) {
        dayMap[dayKey] = [];
      }
      dayMap[dayKey].push(range);
    }
  }

  //  For each dayKey, sort and merge overlapping ranges.
  for (const dayKey of Object.keys(dayMap)) {
    const sorted = dayMap[dayKey].sort(
      (a, b) => a.startDate.getTime() - b.startDate.getTime()
    );

    const merged: DateRange[] = [];
    for (const range of sorted) {
      const last = merged[merged.length - 1];
      if (!last || last.endDate < range.startDate) {
        merged.push(range);
      } else {
        last.endDate = new Date(
          Math.max(last.endDate.getTime(), range.endDate.getTime())
        );
      }
    }
    dayMap[dayKey] = merged;
  }

  // Sum each day's merged ranges.
  const result: DailyDataQueryResult = {};

  for (const dayKey of Object.keys(dayMap)) {
    const totalSeconds = dayMap[dayKey].reduce(
      (sum, range) => sum + differenceInSeconds(range.endDate, range.startDate),
      0
    );
    result[dayKey] = Math.floor(totalSeconds / 60);
  }

  return result;
} 
```

src/helpers/date-range.ts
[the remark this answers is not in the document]
ajmenca: I manage to somehow forget about `Record` all the time because I (like many people) learned the `{ [...] : ...}` technique first.  As far as I know, they're virtually identical, and generally the argument is simply that `Record` is a lot easier to read (and remember how to write out, frankly).  This also isn't a change that will bubble out anywhere because as far as actual type safety goes, they're functionally equivalent and nothing will break (see this [proof](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAkgJhAdsAlgYwIYBsA8BBAGgCEA+KAXijyggA9gk4BnKIqAflZvsZeo6jAATgFdoALigAzbEwnTZEANwBYAFDr1Aei2DR0auSNc8AOQAiXI5TzrQkQRCbAKsBMnTYcAbwDaAa0lnIRREAHMAXUkRRH9EAHsAd0QAXwIoACUINHihOBxg0LD0mLikxBISVQ01HWkRIWAACwghKDAhePipdRzEZygMST9AqELwqKhShOSU128U6r6BgCNJLJy8guEiktiZitcMJSg6hJohTqFq3vj+lwwhqEzs3Pzx4qn98rJKBaW7qs1lARkEdhNot9ZkcTmd4hcrkogA) I found that someone cooked up).  

While there is a fine detail in the very, very slight difference in definition with `[key: string]` versus `[key in string]`, I don't think the way `Record` is defined is meant to imply it should *only* be used for mapped types, just that it *can* be used for mapped types.

I think this just boils down to a readability argument, and I'd agree that `Record` is a little better/simpler.

src/helpers/time-range.ts
ajmenca: > I also had to draw this out on my whiteboard to really understand it. All the UTs were passing, though, so I was fairly confident it was correct. 😃

If it makes you feel any better, I had to draw it too, and then read it a handful of times because I kept *reading it wrong*.  Reading it too fast I just kept assuming that it was comparing end dates against each other, and not end dates against start dates.

src/helpers/date-range.ts
greinard: I think I'd prefer to keep them separate for the time being.  They are coupled in their current usages, but they could be used independently.

The `buildMinutesResultFromDailyTimeRanges` function is specifically for daily data providers that output "minutes" as the resulting value.  The `computeDailyTimeRanges` function could be used separately by something other than a daily data provider, or by a daily data provider that wants to build a different output result (i.e. hours, days with x number of minutes, any other computation).

I'm open to adding an additional helper function that calls both of the current functions for this common use case, if you think that would be helpful, but I think these two operations are independent enough to remain available as separate functions, too.

src/helpers/date-range.ts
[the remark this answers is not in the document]
greinard: I have updated the type definition to use `Record`:

`export type DailyTimeRanges = Record<string, TimeRange[]>;`

src/helpers/time-range.ts
```diff
@@ -0,0 +1,102 @@
...
+    });
+
+    return dailyTimeRanges;
+}
+
+export function buildMinutesResultFromDailyTimeRanges(startDate: Date, endDate: Date, dailyTimeRanges: DailyTimeRanges): DailyDataQueryResult {
+    const result: DailyDataQueryResult = {};
+
+    let currentDate = startDate;
+    while (currentDate < endDate) {
```
bxh, line 55: This is an inefficient implementation, because it is proportional to the date range. In fact, you just need to iterate over keys in `dailyTimeRanges`.

src/helpers/time-range.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    Object.keys(dailyTimeRanges).forEach(key => {
+        const ranges = dailyTimeRanges[key].sort(
+            (range1, range2) => range1.startTime.getTime() - range2.startTime.getTime(),
+        );
+
+        const mergedRanges: TimeRange[] = [];
+        let currentMergedRange: TimeRange = ranges[0];
+        for (const range of ranges.slice(1)) {
+            if (rangesHaveOverlap(currentMergedRange, range)) {
```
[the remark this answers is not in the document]
greinard: Good catch. 👍 

I've updated:
`if (rangesHaveOverlap(currentMergedRange, range)) {`
to be:
`if (currentMergedRange.endTime >= range.startTime) {`

and removed the `rangesHaveOverlap` function, since it is no longer being used.

src/helpers/time-range.ts
greinard: We can't just iterate over the keys because `dailyTimeRanges` can include day keys that are from prior to the `startDate` or after the `endDate`.  This loop ensures that we have the correct date bounding on the result that gets returned.

greinard: Related to this, I'm noticing that we have some inconsistency across the codebase with regard to whether this kind of loop should be using `currentDate < endDate` or `currentDate <= endDate`.  A quick cursory search shows around a 50/50 distribution of each.  The version used here matches what the AH sleep provider was using, so I'm hesitant to change it as part of this PR, but I am going to log a separate issue to have this inconsistency reviewed and corrected, if necessary.

greinard: https://github.com/CareEvolution/MyDataHelpsUI/issues/420

bxh: In the current code, we loop through every calendar day from `startDate` to `endDate`, even though `result` only contains keys for days that actually have data. This can be unnecessarily large if `startDate` and `endDate` are far apart, but the data points only exist on a few days. If a day key falls out of range, simply skip it.

greinard: I have updated the code to loop the keys instead.  I'm not sure we are going to see much of a performance difference at the scale we are generally using these, but with extreme enough examples, there was a slight performance benefit.

We tend to use date ranges that span a week, a month, or in some less common cases, up to a year, when querying for daily data.  With a 365 day date range and only a few days populated with data, the difference in the two implementations was ~2ms in my testing.  With a 10,000 day range, and the same few days populated with data, the difference was ~40ms.  At shorter date ranges or when the date ranges were more fully populated with data, the performance was approximately the same.  I think this may be because although we are looping fewer times over the keys, we introduce a date parsing operation and two date comparisons for each iteration of the loop.  The former impl had to increment and format a date, but then it just used that string to index into the existing data.  🤷‍♂️ 

In any case, while the performance benefit I observed was slight, it is there, and perhaps it will be more so when running in a browser or on a mobile device than it was in my dev machine unit tests. 😃

bxh: Thanks for making the adjustment! While your current tests show just a small difference, that overhead can compound under heavier loads especially on hand devices. Further, iterating only over days you actually need is conceptually cleaner—why iterate days you know have no data? Whether the overhead is 40ms or 2ms, it’s a best practice to skip unnecessary loops. Plus, you get a code path that scales better by design. Suppose that we need to handle years long retrospective calculation, the new approach won't degrade as quickly.

bxh (approved):
🚢

PR #415, opened 2025-02-16
Fix nav bar height when no title / subtitle

chrisnowak:
## Overview

When there is no title/subtitle in the navigation bar, it was getting squished like this:
image

This is because 1) the nav bar uses box-sizing border-box and 2) the top padding adds the safe-area-inset-top (which is like 90px or something on iOS), and the only thing causing the nav bar to have height was min-height:56px, which was essentially irrelevant because of 1/2.

Easy solution is to incorporate the safe area calculation into the min-height.

I have verified this fix in the iOS safari debugger.

## Security

### Testing
- [x] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/presentational/NavigationBar/NavigationBar.css
+++ b/src/components/presentational/NavigationBar/NavigationBar.css
@@ -1,6 +1,6 @@
 ﻿.mdhui-navigation-bar {
     text-align: left;
-    padding: calc(8px + 24px) 16px 8px;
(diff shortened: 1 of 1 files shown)


mmertsock:
Confirmed (via Safari debugger) that this fixes the layout issues I was seeing in the iOS native app.

bxh (approved):

PR #416, opened 2025-02-16
Fix relative activity bars if no available data

chrisnowak:
## Overview

This fixes a bug with the relative activity day coordinator.  The intent was that if you do not have data for a particular data type, it would just not show up in this view; for instance we might include a bunch of data types from Fitbit, Garmin, Apple Health ,etc. but they would only show if you had data.  

However, the bars were just rendering based on props.datatypes rather than availableDataTypes, so it would awkwardly look like this:
Screenshot 2025-02-16 at 10 05 41 AM

This fixes it to use availableDataTypes 

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


bxh (approved):
Out of scope but the function name `RelativeActivityDateRangeCoordinator` doesn't match the file name.

PR #423, opened 2025-02-21
Connect Devices: Air Quality / Weather

ajmenca:
## Overview

This branch adds Air Quality / Weather to the Connect Devices menu.  Since Weather/AQI data is only collected for participants that have a Postal Code specified, this is often unavailable if this data is not collected during onboarding.  This change allows the Connect Devices menu to leverage a `postalCodeSurveyName` which can then be launched in order to set a Postal Code value in the participant's Demographics.  This survey can therefore be customized to either write data to the Demographics Postal Code field, or specifc Custom Fields for Home/Work (as described on the MDHD Data Collections page).

For this option to appear in the menu, the Project needs to have either Weather or AQI data collection enabled.  The menu item is dynamic for prompting the appropriate connections:
image
image
image

When connected, the menu item will display the associated Postal Code:
image

This branch also adds a new Daily Data Provider for the base `AirQuality` type.  This is where device data points are written to by default.  Our current Daily Data Providers assumed that a Project had set up `HomePostalCode` and `WorkPostalCode` Custom Fields.  If a project didn't set those up, they currently couldn't display AQI data in a chart via our predefined options.

The menu item does attempt to be smart about displaying a connected postal code from custom fields (prioritizing it above the base demographics field).

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
```diff
@@ -65,18 +70,22 @@ function useConnectDevicesMenuState(
...
             previewStateObject.platform = "iOS";
             previewStateObject.hasRecentAppleHealthData = true;
+            previewStateObject.participantInfo!.demographics.postalCode = "00000";
         }
         else if (previewState == "iOS") {
             previewStateObject.platform = "iOS";
         } else if (previewState == "Android") {
             previewStateObject.platform = "Android";
+        } else {
+            previewStateObject.participantInfo!.demographics.postalCode = "";
```
greinard, line 88: I'm not sure this is necessary.  The `sampleParticipantInfo` looks like it will already have an empty string postal code.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
```diff
@@ -89,16 +98,18 @@ function useConnectDevicesMenuState(
...
             MyDataHelps.getDataCollectionSettings(),
             MyDataHelps.getExternalAccounts(),
-            MyDataHelps.getDeviceInfo()
+            MyDataHelps.getDeviceInfo(),
+            MyDataHelps.getParticipantInfo()
         ])
-            .then(([settings, accounts, deviceInfo]) => {
+            .then(([settings, accounts, deviceInfo, participantInfo]) => {
                 let newState: ConnectDevicesMenuState = {
                     loading: true,
```
greinard, line 106: I know this existed before your changes, but it looks like this could be set to `false` here and the few `newState.loading = false;` calls below could be removed.  Since this `newState` never gets set as THE `state` while the flag is `true`.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
```diff
@@ -49,7 +53,8 @@ function useConnectDevicesMenuState(
         deviceExternalAccounts: null,
         platform: null,
         hasRecentAppleHealthData: false,
-        healthConnectStatus: null
+        healthConnectStatus: null,
+        participantInfo: null
     });
 
     if (previewState) {
```
greinard, line 60: Pre-existing condition, but I think this whole preview block should be moved to be the first thing inside the `initialize()` function, and it should set the `state` to be the preview state rather than returning it.  As-is, this technically breaks the React Rules Of Hooks, because the `useInitializeView(...)` hook call below (which calls `useEffect` internally) is being called conditionally.

It is splitting hairs a little bit because we are not likely to swap between a preview state and a non-preview state during a particular use of the component, but that possibility does exist.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.previewdata.ts
```diff
@@ -82,4 +82,34 @@ export const previewHealthConnectStatus: HealthConnectStatus = {
     enabledQueryRules: [],
     running: false,
     lastPrompted: ""
+};
+
+export const sampleParticipantInfo : ParticipantInfo = {
```
greinard, line 87: Not a big deal, but do we need to fully populate this?  You can probably just use a few `as Type` assertions and only populate the fields you need for this component to render.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
```diff
@@ -141,7 +152,7 @@ export default function (props: ConnectDevicesMenuProps) {
 
     console.log("settings", deviceExternalAccounts);
```
greinard, line 153: Pre-existing condition, but I see a few debug console logs got left in this file at some point.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
```diff
@@ -422,4 +445,35 @@ function AppleHealthMenuItem(props: AppleHealthMenuItemProps) {
             </div>
         }
     </div>;
+}
+
+function WeatherMenuItem(props: { settings: DataCollectionSettings | null, participantInfo: ParticipantInfo | null, postalCodeSurveyName: string }){
```
greinard, line 450: I don't feel that strongly about it, but you can probably drop the `| null`'s from the types here.  I think those will need to be set by the time we get to this point, so the calling code could just use `!`.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
```diff
@@ -422,4 +445,35 @@ function AppleHealthMenuItem(props: AppleHealthMenuItemProps) {
...
+    let action = () => {
+        MyDataHelps.startSurvey(props.postalCodeSurveyName);
+    }
+
+    const postalCodeCustomFields = Object.keys(props.participantInfo ? props.participantInfo.customFields : {}).filter( (k) => k.endsWith('PostalCode'));
+    const postalCodes = [...postalCodeCustomFields.map((cf) => props.participantInfo?.customFields[cf]), props.participantInfo?.demographics.postalCode];
```
greinard, line 456: Minor:  You can drop the parens from these single arg arrow functions if you'd like.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
```diff
@@ -422,4 +445,35 @@ function AppleHealthMenuItem(props: AppleHealthMenuItemProps) {
...
+    const postalCode = postalCodes.find((p) => !!p);
+
+    const indicator = postalCode ? 
+        <div className="mdhui-connect-devices-menu-connect">{postalCode}</div> :
```
greinard, line 460: I'm kind of thinking we should `.join(', ')` all of the non-empty postal codes.  Being able to see both in the home/work scenario would be ideal.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
```diff
@@ -422,4 +445,35 @@ function AppleHealthMenuItem(props: AppleHealthMenuItemProps) {
...
+    }
+    if(props.settings?.weatherEnabled) {
+        titleBits.push('Weather');
+    }
+    const title = titleBits.join(" / ");
+
+    return (
+        <div className="mdhui-connect-devices-menu-device">
+            <Action onClick={action} indicator={indicator}>
+                <Title autosizeImage image={<FontAwesomeIcon icon={faSun} color={"yellow"}/>} order={4}>{title}</Title>
```
greinard, line 475: I think we typically use `FontAwesomeSvgIcon` for the tree shaking/bundle size benefits.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
```diff
@@ -422,4 +445,35 @@ function AppleHealthMenuItem(props: AppleHealthMenuItemProps) {
...
+            </Action>
+        </div>
+    )
```
greinard, line 478: Minor:  Missing semicolon.  I noticed there are some other pre-existing missing semi-colons up through the file too.

src/helpers/strings-fr.ts
```diff
@@ -10,6 +10,7 @@ let strings: { [key: string]: string } = {
     "settings": "Paramètres",
     "connect": "Connecter",
     "reconnect": "Reconnecter",
+    "setup": "Configuration",
```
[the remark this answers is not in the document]
lynnfaraday: If we intended it as a verb, then it would be "Set Up" in English, not "Setup".  So I think since we're using it as a noun (implication: "Access the Weather Setup"), what is here matches.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
```diff
@@ -15,8 +15,10 @@ import OmronLogo from '../../../assets/omron-logo.png';
 import HealthConnectLogo from '../../../assets/healthconnect-logo.svg';
 import { add, formatISO } from 'date-fns';
 import { useInitializeView } from '../../../helpers';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSun } from '@fortawesome/free-regular-svg-icons';
 
-export type DeviceAccountType = "Fitbit" | "Garmin" | "Dexcom" | "AppleHealth" | "GoogleFit" | "Omron" | "HealthConnect";
+export type DeviceAccountType = "Fitbit" | "Garmin" | "Dexcom" | "AppleHealth" | "GoogleFit" | "Omron" | "HealthConnect" | "Weather";
```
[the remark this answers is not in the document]
lynnfaraday: Agreed it's more clear to use "Geographic" since it (presumably) includes both Weather and Air Quality.

src/helpers/strings-fr.ts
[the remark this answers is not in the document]
lynnfaraday: (Of course we could always change the English to verb form as well, but it looks like that was probably pre-existing and we'd have to see where else it was used.)

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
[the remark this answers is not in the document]
chrisnowak: Actually was just chatting with Vik on this on Friday; I think we should actually go with "Environmental" for this category (and update DesignerUI as well)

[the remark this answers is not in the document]
lynnfaraday: Environmental is a great name for it.

src/helpers/strings-fr.ts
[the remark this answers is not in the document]
greinard: My impression was that we wanted the button to be an action "Set this up", rather than "The Setup", but if the latter is the case, then the other language entries may need adjustment, as I believe they are using verbs.  🙂

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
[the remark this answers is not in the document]
greinard: I agree.

src/helpers/strings-fr.ts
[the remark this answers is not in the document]
lynnfaraday: In general I agree - buttons should normally use verbs, and we should be consistent across the languages. 

Grammatically, though, "Setup" is the noun form and "Set Up" is a tiny bit awkward on a button prompt due to the dangling preposition. 

Based on the ChatGPT feedback and the languages I recognize, it seems that they're mostly consistent as nouns currently (e.g., the German one is "Settings").

If we want to change it, I would suggest making the English button either "Enable", "Activate", or "Configure", and then make the other languages match.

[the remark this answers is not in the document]
ajmenca: I grabbed this translation (and all the ones mentioned below from the existing `asthma-air-qualities-setup-button-text` translation key.  I assumed that the usage was going to be similar.  I'll be honest, I don't think I would notice the discrepancy between the other menu items in this block being verbs and "Setup" being a noun.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
ajmenca: Annoyingly, it is, because of what I'm doing on line 74.  That change ends up persisting when you move between stories if you don't reset it here.  Changing line 74 to `participantInfo: {...sampleParticipantInfo}` isn't sufficient either because both things end up with a reference to the same `demographics` object.  I could avoid this by doing a deep copy on line 74 with JSON.parse tricks, but that seemed almost excessive.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
ajmenca: I'd prefer not to make a change like that here in a (mostly) unrelated PR.  I generally agree with you because when I came to edit this component I was kind of confused by the order of events, and it uses a different pattern than we've generally established (`useState` gets called in `useConnectDevicesMenuState` and then gets decomposed in the main component).  But I'd rather do that and test it in its own branch/PR as a bug fix/optimization.

src/helpers/strings-fr.ts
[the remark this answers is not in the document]
greinard: Well, there's your problem. 😆  TBH, I'm not entirely sure where I got the Asthma related translations from.  I think it was ChatGPT, but I don't know enough about the other languages to say whether they are correct.

I do think you are right, though, that the intent/usage for that button is the same.  So if these need to be updated, we should probably update the ones being used for Asthma AQI as well. 🙂

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
greinard: Ah, I see what you mean.

Not a big deal, but if you want to drop this, you could probably export a `createSampleParticipantInfo()` function instead of exporting the `sampleParticipantInfo` object directly.  Then each story would have a different one.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
```diff
@@ -237,6 +244,14 @@ export default function (props: ConnectDevicesMenuProps) {
             healthConnectStatus={healthConnectStatus} />;
     }
 
+    function getEnvironmentalMenuItem() {
+        if (!accountTypes.includes("Environmental") || !props.postalCodeSurveyName) {
```
greinard, line 248: This should work fine, but the second condition could also be moved up to the account type filtering so you can just check the account types here like we do for the other menu items.

```
if ((!settings?.airQualityEnabled && !settings?.weatherEnabled) || !props.postalCodeSurveyName) { 
	accountTypes = accountTypes.filter(a => a != "Environmental");
}
```

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
```diff
@@ -422,4 +438,38 @@ function AppleHealthMenuItem(props: AppleHealthMenuItemProps) {
...
+
+function EnvironmentalMenuItem(props: { settings: DataCollectionSettings | null, participantInfo: ParticipantInfo | null, postalCodeSurveyName: string }){
+    let action = () => {
+        MyDataHelps.startSurvey(props.postalCodeSurveyName);
+    }
+
+    const postalCodeCustomFields = Object.keys(props.participantInfo ? props.participantInfo.customFields : {}).filter( (k) => k.endsWith('PostalCode'));
+    const postalCodes = [...postalCodeCustomFields.map((cf) => props.participantInfo?.customFields[cf]), props.participantInfo?.demographics.postalCode];
+    var postalCode = postalCodes.filter((p) => !!p).slice(0,3).join(', ');
+    if(postalCodes.length > 3) {
```
greinard, line 451: If 3 ends up being too long, I think 2 + ", ..." would be fine as well.  I suspect most projects don't go above 2.

Alternatively, maybe you could join them all and update the `mdhui-connect-devices-menu-connect` style class to use text overflow ellipsis.  I haven't tinkered with that though, so it may be more trouble than it is worth.  😃

greinard (approved):
LGTM.  I added a couple more minor suggestions.  Feel free to :ship: when ready.

Let me know if you'd like me to take another pass through before you do.

src/components/container/ConnectDevicesMenu/ConnectDevicesMenu.tsx
ajmenca: Yeah, I figured most wouldn't get beyond 2, but based on the length of the next longest thing which seems to show up in this area which is "Download MyDataHelps", 3 seemed to fit fine.  I figured it was an unlikely edge case in general, but wanted to cover it just in case.  I didn't want to mess with the css class just to have this sort of decision be localized to this specific menu item.

PR #427, opened 2025-02-24
Fix MicroTrend Grid

chrisnowak:
## Overview

Turns out my fancy CSS rules for hiding grid columns if there's just an empty card child causes some major issues in iOS safari.

Unfortunately it looks like there's not much of an elegant solution to the whole "hide an element if it's children are hidden" problem, without implementing a whole lot of "onVisible" props or something.  I would definitely like to avoid that.

So this tackles it a bit of a different way by having a "Card" variant for "GridColumn".  This means that Grid columns themselves can just render as cards, which of course already hide themselves if they are empty via the tried and true ```:empty``` css selector

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@careevolution/mydatahelps-ui",
-  "version": "2.69.0",
+  "version": "2.69.1-FixTrendGrid.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@careevolution/mydatahelps-ui",
-      "version": "2.69.0",
+      "version": "2.69.1-FixTrendGrid.0",
       "license": "MIT",
       "dependencies": {
         "@emotion/react": "11.11.3",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@careevolution/mydatahelps-ui",
-  "version": "2.69.0",
+  "version": "2.69.1-FixTrendGrid.0",
   "description": "MyDataHelps UI Library",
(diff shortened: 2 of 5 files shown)


felideon (approved):
The changes make sense. 

There's something funky in Safari (mostly on iOS but also on Desktop) with that story, where the grid wouldn't render at all, but it's unrelated to the code changes as I can reproduce it on main. Might have to do with the 401 errors, but not sure why Chrome or Firefox have no problem at all.

Screenshot 2025-02-25 at 3 11 49 PM

chrisnowak:
The 401s are expected, as i'm just basically making a couple components fail on purpose to show the grid collapsing.  But i'll double check safari here...

chrisnowak:
@felideon how are you testing?  Seems like it's working for me in storybook on both safari and iOS...what you're describing sounds like the behavior before the fix

I did fix an issue with the stories which was causing that "Each child in the list should have a unique key" thing, but I would be surprised if that caused a safari-specific rendering bug

iOS Safari:
image

felideon:
Just navigating on my phone to storybook served from my laptop:

felideon:
And yeah, I don't see that React error anymore (I'm connecting the inspector to my phone browser) but still no Grid. iOS 18.3.1 and cleared cache, fwiw.

chrisnowak:
(we realized this error was to do with how @felideon was serving the content from http, not https)

PR #428, opened 2025-02-26
Missing Strings and Unit Test

lynnfaraday:
## Overview

Adding unit test to ensure there are no missing strings. (Which found a few missing strings, now fixed.)

## Security

### Testing

Check language switching with storybook.

### Documentation
n/a

## Reviewers

- Subject-matter experts
- Style/editing reviewers


__tests__/helpers/locale/language.test.ts
```diff
@@ -45,9 +46,31 @@ describe('Language Tests', () => {
...
+        it('Should return empty string if the key is not found.', () => {
             const result = language("no_such_string", "xx");
-            expect(result).toBe("no_such_string");
+            expect(result).toBe("");
         });
     });
+
+    describe('supportedLocales', () => {
+        it('Should have all strings in all locales', () => {
+            const deliberatelyBlankStrings =
```
lynnfaraday, line 57: This is a temporary workaround. We really shouldn't have blank strings. The only reason this one exists is because the asthma log is not properly handling string interpolation for different languages. 

`asthma-log-entry-details-component-no-data-p1` and `asthma-log-entry-details-component-no-data-p2` currently try to account for the before/after pieces of a string like this:

English: "No ((symptoms/triggers/impacts/etc.)) recorded"
Spanish: "No recorded ((symptoms/triggers/impacts/etc.))"

This solution isn't flexible enough for all future languages, and I'll do a separate PR to fix it and clean this up.

greinard, line 57: I'm seeing an error below because the key type for this const is inferred as `any` and the key being used to index it below is a `string`.  This should fix that.

`const deliberatelyBlankStrings: { [key: string]: string[] } = `

__tests__/helpers/locale/language.test.ts
```diff
@@ -45,9 +46,31 @@ describe('Language Tests', () => {
...
+            };
+
+            Object.keys(englishStrings).forEach(localeStr => {
+                supportedLocales().forEach(localeName => {
+                    const otherString = language(localeStr, localeName);
+                    if (!otherString) {
+                        const exceptions = deliberatelyBlankStrings[localeStr] || [];
+                        if (!exceptions.includes(localeName)) {
+                            console.log(`Missing ${localeStr} in ${localeName}`);
+                            expect(otherString).toBeTruthy();
```
greinard, line 69: Maybe consider replacing these two lines with: 
```
throw `Missing ${localeStr} in ${localeName}`;
```

src/helpers/strings-pt-pt.ts
```diff
@@ -522,7 +522,13 @@ let strings: { [key: string]: string } = {
...
     "daily": "Diário",
     "weekly": "Semanal",
-    "monthly": "Mensal"
+    "monthly": "Mensal",
+    "bonus": "Bónus",
+    "syncing-data": "Sincronização de dados...",
+    "health-connect-phr-sync-title": "Sincronizar com o Health Connect",
+    "health-connect-phr-sync-prompt": "Escolha os registos de saúde para ler e escrever no Health Connect",
+    "setup": "Configurar",
+    "air-quality": "Qualidade do Ar"
```
greinard, line 531: Though it isn't required for these to work, I find it helpful if the language strings are in the same position in all the files.  Any chance we could insert these missing strings at the same locations where they exist in the English file?

greinard (approved):
LGTM.  Just few minor suggestions.  Feel free to :ship: when ready.

__tests__/helpers/locale/language.test.ts
nathia: That's a nice work around for now, but yes we will have to revisit this and make it work for all languages. Agreed, one step at a time.

nathia (approved):
Everything looks great! That unit test was definitely needed, and as you mentioned, it’s already helping us catch missing strings. Well done!

src/helpers/strings-pt-pt.ts
lynnfaraday: The  inconsistent order bugs me too, but they're not all in the same places now regardless. 
Suggest a follow-up PR to put all the string files in alphabetical order, and then hopefully it'll be easier to maintain going forward?

greinard: > Suggest a follow-up PR to put all the string files in alphabetical order, and then hopefully it'll be easier to maintain going forward?

I like that idea.  Alpha sort may "ungroup" some keys that were purposefully co-located when added, but at least the order would then be consistent across the files, so I'd be fine with that.

PR #429, opened 2025-02-26
Add Health Connect support to daily data types

bxh:
## Overview

This pull request replaces [PR #413](https://github.com/CareEvolution/MyDataHelpsUI/pull/413) and introduces Health Connect as a new data source across multiple components and stories. It adds the appropriate data providers and availability checks to allow viewing Health Connect data (e.g., resting heart rate, sleep minutes) in a number of charts.

1.  `DataCollectionSettings`
    - New helper class for checking if Health Connect and specific Health Connect data types are enabled. This centralizes logic to determine whether a given namespace (Fitbit, Garmin, Apple Health, Health Connect) is queryable.
  
2. Daily Data Providers:
    - Added Health Connect resting heart rate, glucose and sleep providers.
    - Updated combined providers for resting heart rate, glucose and sleep to include Health Connect data when enabled.

3. Availability Checks:
    - Extended existing checks to handle Health Connect data types via V2 API queries (e.g., simpleAvailabilityCheckV2).

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

Tested with project with and without HC rules enabled.

Screenshot 2025-02-25 214659
Screenshot 2025-02-25 214854
Screenshot 2025-03-11 225029

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/helpers/daily-data-types/combined.tsx
```diff
@@ -5,28 +5,27 @@ import React from "react";
...
+import MyDataHelps, { DeviceDataNamespace, DeviceDataV2Namespace } from "@careevolution/mydatahelps-js";
 import { formatNumberForLocale } from "../../helpers/locale";
+import { DataCollectionSettings } from "../daily-data-providers/data-collection-settings";
 
 let combinedTypeDefinitions: DailyDataTypeDefinition[] = [
     {
         dataSource: "Unified",
         type: DailyDataType.RestingHeartRate,
         dataProvider: combinedRestingHeartRate,
         availabilityCheck: async (modifiedAfter?: Date): Promise<boolean> => {
```
chrisnowak, line 18: Outside this PR, but we should add a follow up issue to refactor availabilityCheck to take in DataCollectionSettings; this is gonna be super chatty.

src/helpers/daily-data-providers/data-collection-settings.ts
```diff
@@ -0,0 +1,63 @@
...
+        this.settings = settings;
+        this.deviceDataTypes = deviceDataTypes;
+    }
+
+    static async create(): Promise<DataCollectionSettings> {
+        const settings = await MyDataHelps.getDataCollectionSettings();
+        let deviceDataTypes: SupportedDeviceDataV2DataType[] = [];
+        
+        // HACK HACK HACK: Need a better way to tell if V2 is needed.
+        if (settings.healthConnectEnabled) {
```
chrisnowak, line 25: I think we just get v2 every time

src/helpers/daily-data-types/availability-check.ts
```diff
@@ -1,15 +1,88 @@
...
+            const result = await MyDataHelps.queryDeviceDataV2(parameters);
+            return result.deviceDataPoints.length > 0;
+        } catch {
+            return false;
+        }
+    };
+}
+
+export async function checkSourceAvailability(
+    settings: DataCollectionSettings,
```
chrisnowak, line 53: Yeah this method is not bad but I would have it take in the "CombinedDataCollectionSettings" I suggested

src/helpers/daily-data-providers/data-collection-settings.ts
```diff
@@ -0,0 +1,63 @@
+import MyDataHelps, {
+    DeviceDataNamespace,
+    DeviceDataV2Namespace,
+    DataCollectionSettings as MDHDataCollectionSettings,
+    SupportedDeviceDataV2DataType,
+} from "@careevolution/mydatahelps-js";
+
+export class DataCollectionSettings {
```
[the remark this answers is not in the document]
chrisnowak: But yeah I'd also name the combined object "CombiendDataCollectionSettings" to differentiate it from the MDH.js SDK thing

src/helpers/glucose-and-meals/glucose.ts
```diff
@@ -41,20 +43,39 @@ export async function googleFitBloodGlucoseDataProvider(startDate: Date, endDate
     });
 }
 
+export async function healthConnectBloodGlucoseDataProvider(startDate: Date, endDate: Date): Promise<Reading[]> {
+    const query: DeviceDataV2Query = {
+        namespace: "HealthConnect",
+        type: "blood-glucose",
+        observedAfter: add(startDate, { days: -1 }).toISOString(),
```
chrisnowak, line 50: Why are you adding a day to the start/end date?  Doesn't seem like that is what we are doing for Apple Health or google fit.

I think that is only necessarily in some weird cases for timezone issues with the daily data providers, but this isn't daily data, it's intraday

src/helpers/daily-data-providers/health-connect-sleep.ts
```diff
@@ -0,0 +1,27 @@
+﻿import { querySleepMinutesV2 } from "./common-sleep-v2";
+
+export function totalSleepMinutes(startDate: Date, endDate: Date) {
+    return querySleepMinutesV2("HealthConnect", "sleep", startDate, endDate, [
```
chrisnowak, line 4: I think per @greinard we need to update the querySleepMinutesV2 to handle overlaps still...so it needs to be more like the Apple Health V1 provider.

A bit conflicted on whether we can/should ship it like this.  It would only impact people with multiple sources of sleep data - which I think would be mostly uncommon - but certanily not unheard of (Oura ring and Fitbit, or use of Apple's "InBed" timer and a Garmin, etc.)

I think we can address it in a separate PR but likely should do so soon

chrisnowak (changes requested):
I think we're logically mostly right here.  Architecture wise the big thing is that I would drop the DataCollectionSettings full abstraction in favor of a simpler approach and perhaps a few helper methods - as I think we need to consider things like querying V2 Fitbit and Apple Health data as well.

src/helpers/glucose-and-meals/glucose.ts
bxh: Removed.

src/helpers/daily-data-providers/health-connect-sleep.ts
greinard: I can work on a PR for that today.

greinard: https://github.com/CareEvolution/MyDataHelpsUI/pull/441

src/helpers/daily-data-providers/combined-therapy-minutes.ts
```diff
@@ -1,16 +1,17 @@
...
 
-    const settings = await MyDataHelps.getDataCollectionSettings();
-    if (settings.queryableDeviceDataTypes.find(type => type.namespace == 'AppleHealth' && type.type == 'MindfulSession')) {
+    const combinedSettings = await getCombinedDataCollectionSettings();
+    const { settings } = combinedSettings;
+    if (settings.appleHealthEnabled && settings.queryableDeviceDataTypes.find(type => type.namespace == 'AppleHealth' && type.type == 'MindfulSession')) {
         providers.push(appleHealthTherapyMinutesDataProvider(startDate, endDate));
     }
-    if (settings.queryableDeviceDataTypes.find(type => type.namespace == 'GoogleFit' && type.type == 'SilverCloudSession')) {
+    if (settings.googleFitEnabled && settings.queryableDeviceDataTypes.find(type => type.namespace == 'GoogleFit' && type.type == 'SilverCloudSession')) {
```
bxh, line 14: Need to check `googleFitEnabled` here? @chrisnowak

src/helpers/daily-data-types/availability-check.ts
```diff
@@ -1,15 +1,146 @@
...
+        combinedSettingsPromise = getCombinedDataCollectionSettings();
+    }
+    return combinedSettingsPromise;
+}
+
+export async function checkSourceAvailability(
+    sources: DataSource[],
+    modifiedAfter?: Date,
+): Promise<boolean> {
+    const combinedSettings = await getCachedCombinedSettings();
```
bxh, line 83: No longer needs to make a DateSourceCollectionSetting request for each namespace.

src/helpers/daily-data-providers/data-collection-settings.ts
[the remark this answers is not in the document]
bxh: I refactored the code to use `getCombinedDataCollectionSettings`, as an interim approach until we figure out the right pattern. @chrisnowak 

Let me know what you think.

bxh:
I'll review this today.

src/helpers/daily-data-providers/combined-steps.ts
```diff
@@ -1,45 +1,50 @@
...
+    garminStepsDataProvider,
+    googleFitStepsDataProvider,
+} from ".";
 import getDayKey from "../get-day-key";
-import MyDataHelps from "@careevolution/mydatahelps-js";
 import { DailyDataQueryResult } from "../query-daily-data";
+import { getCombinedDataCollectionSettings } from "./combined-data-collection-settings";
+
+export default async function (startDate: Date, endDate: Date, includeGoogleFit?: boolean) {
+    const combinedSettings = await getCombinedDataCollectionSettings();
```
chrisnowak, line 13: Seems like we can just call MyDataHelps.getDataCollectionSettings to get settings here, otherwise this just adds an extra API call for nothing

src/helpers/daily-data-providers/combined-therapy-minutes.ts
```diff
@@ -1,16 +1,17 @@
...
 import { DailyDataQueryResult } from '../query-daily-data';
 import { combineResults } from './common-mindful-and-therapy';
+import { getCombinedDataCollectionSettings } from './combined-data-collection-settings';
 
 export default async function (startDate: Date, endDate: Date): Promise<DailyDataQueryResult> {
     const providers: Promise<DailyDataQueryResult>[] = [];
 
-    const settings = await MyDataHelps.getDataCollectionSettings();
-    if (settings.queryableDeviceDataTypes.find(type => type.namespace == 'AppleHealth' && type.type == 'MindfulSession')) {
+    const combinedSettings = await getCombinedDataCollectionSettings();
```
chrisnowak, line 9: This one too...it's just adding another API call right now until this data lives in V2.  May as well just leave this as-is

src/helpers/daily-data-providers/combined-therapy-minutes.ts
chrisnowak: If googleFitEnabled is false, there'd be no queryableDeviceDataTypes for google fit

chrisnowak: So it's redundant but fine to leave

src/helpers/daily-data-types/availability-check.ts
chrisnowak: Data collection settings can and does absolutely change based on user interaction (e.g. a user opting in to sharing some data), so we wouldn't want to permanently cache it.  

We can likely bust the cache on applicationDidBecomeVisible, but we'd need to test that extensively...likely would have some weird order of operations issues if we just did that

src/components/container/GlucoseChart/GlucoseChart.stories.tsx
```diff
@@ -44,28 +67,18 @@ export const Default = {
...
-        averageGlucoseLineColor: {
-            name: 'average glucose line color',
-            control: 'color'
-        }
-    },
+    argTypes: sharedArgTypes,
     render: render
 };
 
+export const Live = {
```
greinard, line 74: What does having this additional export gain us?  The default export above already supports the "live" state.

src/components/container/DeviceDataMonthChart/DeviceDataMonthChart.stories.tsx
```diff
@@ -1,63 +1,119 @@
...
+    },
 } as ComponentMeta<typeof DeviceDataMonthChart>;
 
-const Template: ComponentStory<typeof DeviceDataMonthChart> = (args: DeviceDataMonthChartProps) =>
-	<Layout colorScheme="auto">
-		<Card>
-			<DeviceDataMonthChart {...args} />
-		</Card>
-	</Layout>;
+const Template: ComponentStory<typeof DeviceDataMonthChart> = (
```
greinard, line 18: Since this story is being updated, it should also be converted to the newer non-deprecated format to clean up the rollup errors.  I believe most/all of the presentational component stories have been updated if you'd like to see an example.

src/components/view/DeviceActivityView/DeviceActivityView.stories.tsx
```diff
@@ -16,4 +16,9 @@ export const Push = {
 export const Modal = {
     args: { presentation: "Modal", previewState: "default" },
     render: render
+};
+
+export const Live = {
+    args: { previewState: "live" },
```
greinard, line 22: I don't think this is correct here.  This view only supports a preview state of `"default" | undefined`, so if you want a live state, I think you can just omit the preview state.

src/helpers/daily-data-providers/combined-data-collection-settings.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export async function getCombinedDataCollectionSettings(): Promise<CombinedDataCollectionSettings> {
+    const [settings, deviceDataV2Types] = await Promise.all([
+        MyDataHelps.getDataCollectionSettings(),
+        MyDataHelps.getDeviceDataV2AllDataTypes(),
+    ]);
+
+    return {
+        settings,
```
greinard, line 18: Rather than having a settings object with a settings property.  Maybe we use `...settings` here and adjust the interface definition above to be:
```
export interface CombinedDataCollectionSettings extends DataCollectionSettings {
    deviceDataV2Types: SupportedDeviceDataV2DataType[];
}
```

src/helpers/daily-data-providers/combined-resting-heart-rate.ts
```diff
@@ -1,42 +1,73 @@
...
+    }
+    if (settings.garminEnabled) {
+        providers.push(garminRestingHeartRateDataProvider(startDate, endDate));
+    }
+    if (
+        settings.appleHealthEnabled &&
+        settings.queryableDeviceDataTypes.some(
+            (ddt) =>
+                ddt.namespace === "AppleHealth" &&
+                ddt.type === "RestingHeartRate",
```
greinard, line 31: Extra comma.

src/helpers/daily-data-providers/combined-resting-heart-rate.ts
```diff
@@ -1,42 +1,73 @@
...
+        providers.push(
+            appleHealthRestingHeartRateDataProvider(startDate, endDate),
+        );
+    }
+    if (
+        settings.healthConnectEnabled &&
+        deviceDataV2Types.some(
+            (ddt) =>
+                ddt.namespace === "HealthConnect" &&
+                ddt.type === "resting-heart-rate",
```
greinard, line 43: Extra comma.

src/helpers/daily-data-providers/combined-resting-heart-rate.ts
```diff
@@ -1,42 +1,73 @@
...
+        settings.appleHealthEnabled &&
+        settings.queryableDeviceDataTypes.some(
+            (ddt) =>
+                ddt.namespace === "AppleHealth" &&
+                ddt.type === "RestingHeartRate",
+        )
+    ) {
+        providers.push(
+            appleHealthRestingHeartRateDataProvider(startDate, endDate),
+        );
```
greinard, line 36: Some of the wrapping here seems unnecessary (vs Fitbit/Garmin above).  The formatting can probably be set to allow something a little wider so these remain on a single line.

src/helpers/daily-data-providers/combined-resting-heart-rate.ts
```diff
@@ -1,42 +1,73 @@
...
-	});
+    if (!providers.length) {
+        return {};
+    }
+
+    const values = await Promise.all(providers);
+    const data: Record<string, number> = {};
+    while (startDate < endDate) {
+        let dayKey = getDayKey(startDate);
+        let heartRates: number[] = [];
```
greinard, line 59: I think these can both be `const`.

src/helpers/daily-data-providers/combined-resting-heart-rate.ts
```diff
@@ -1,42 +1,73 @@
...
+            if (value[dayKey] && value[dayKey] > 0) {
+                heartRates.push(value[dayKey]);
+            }
+        });
+        if (heartRates.length > 0) {
+            data[dayKey] = Math.round(
+                heartRates.reduce((a, b) => a + b) / heartRates.length,
+            );
+        }
+        startDate = add(startDate, { days: 1 });
```
greinard, line 70: I know this was pre-existing, but I'd suggest using an intermediate variable to store the current date value for the loop.  I think it is generally considered bad practice to modify the function parameter reference.  By changing the referenced value here, it no longer actually represents the start date, so it could lead to confusion or possibly coding errors for future maintainers (e.g. if someone doesn't notice that `startDate` was modified and tries to use it after the for loop, expecting the original value).

Not a big deal, but as long as we are adjusting this, might as well clean that up, too.

src/helpers/daily-data-providers/combined-sleep-v2.ts
```diff
@@ -1,39 +1,80 @@
...
         providers.push(garminTotalSleepMinutesDataProvider(startDate, endDate));
     }
-    if (settings.queryableDeviceDataTypes.find(s => s.namespace == "AppleHealth" && s.type == "SleepAnalysisInterval")) {
+
+    if (
+        settings.appleHealthEnabled &&
+        settings.queryableDeviceDataTypes.some(
+            (ddt) =>
+                ddt.namespace === "AppleHealth" &&
+                ddt.type === "SleepAnalysisInterval",
```
greinard, line 31: Extra comma.

src/helpers/daily-data-providers/combined-sleep-v2.ts
```diff
@@ -1,39 +1,80 @@
...
     }
 
-    const results = providers.length ? await Promise.all(providers) : [];
+    if (
+        settings.healthConnectEnabled &&
+        deviceDataV2Types.some(
+            (ddt) =>
+                ddt.enabled &&
+                ddt.namespace === "HealthConnect" &&
+                ddt.type === "sleep",
```
greinard, line 43: Extra comma.

src/helpers/daily-data-providers/combined-sleep-v2.ts
```diff
@@ -1,39 +1,80 @@
...
+        )
+    ) {
+        providers.push(
+            healthConnectTotalSleepMinutesDataProvider(startDate, endDate),
```
greinard, line 47: Extra comma.

src/helpers/daily-data-providers/combined-sleep-v2.ts
```diff
@@ -1,39 +1,80 @@
...
+    if (!providers.length) {
+        return {};
+    }
+
+    const results = await Promise.all(providers);
+    const data: Record<string, number> = {};
 
-    let data: { [key: string]: number; } = {};
-    let currentDate = startDate;
+    let currentDate = new Date(startDate);
```
greinard, line 58: Curious why we need a new date here?

src/helpers/daily-data-providers/combined-sleep-v2.ts
```diff
@@ -1,39 +1,80 @@
...
     while (currentDate < endDate) {
         let dayKey = getDayKey(currentDate);
```
greinard, line 60: Pre-existing, but I think `dayKey` can be a `const`.

src/helpers/daily-data-providers/combined-sleep.ts
```diff
@@ -1,42 +1,74 @@
...
+    }
+    if (settings.garminEnabled) {
+        providers.push(garminTotalSleepMinutesDataProvider(startDate, endDate));
+    }
+    if (
+        settings.appleHealthEnabled &&
+        settings.queryableDeviceDataTypes.some(
+            (ddt) =>
+                ddt.namespace === "AppleHealth" &&
+                ddt.type === "SleepAnalysisInterval",
```
greinard, line 28: Extra comma.

src/helpers/daily-data-providers/combined-sleep.ts
```diff
@@ -1,42 +1,74 @@
...
+        )
+    ) {
+        providers.push(appleHealthSleepDataProvider(startDate, endDate));
+    }
+    if (
+        settings.healthConnectEnabled &&
+        deviceDataV2Types.some(
+            (ddt) =>
+                ddt.namespace === "HealthConnect" &&
+                ddt.type === "sleep",
```
greinard, line 38: Extra comma.

src/helpers/daily-data-providers/combined-sleep.ts
```diff
@@ -1,42 +1,74 @@
...
+        )
+    ) {
+        providers.push(
+            healthConnectTotalSleepMinutesDataProvider(startDate, endDate),
```
greinard, line 42: Extra comma.

src/helpers/daily-data-providers/combined-sleep.ts
```diff
@@ -1,42 +1,74 @@
...
             }
-            return data;
         });
-    });
+
+        if (maxSleep !== null) {
+            data[dayKey] = maxSleep;
+        }
+
+        startDate = add(startDate, { days: 1 });
```
greinard, line 70: Same comment here RE using an intermediate variable instead of modifying `startDate`.

src/helpers/daily-data-types/availability-check.ts
```diff
@@ -1,15 +1,146 @@
...
+    DeviceDataV2Query,
+} from "@careevolution/mydatahelps-js";
+import {
+    CombinedDataCollectionSettings,
+    getCombinedDataCollectionSettings,
+} from "../daily-data-providers/combined-data-collection-settings";
+
+export function simpleAvailabilityCheck(
+    namespace: DeviceDataNamespace,
+    type: string | string[],
```
greinard, line 14: A number of extra commas in this file, too.

src/helpers/daily-data-types/availability-check.ts
```diff
@@ -1,15 +1,146 @@
...
+            )(modifiedAfter);
+        } else {
+            return simpleAvailabilityCheck(
+                namespace as DeviceDataNamespace,
+                type,
+            )(modifiedAfter);
+        }
+    });
+
+    const finalResults = await Promise.all(availabilityChecks);
```
greinard, line 144: The combined data type availability checks used to occur in sequential order with a short circuit on the first successful check.  We put the highest likelihood availability checks first to avoid unnecessary availability check API queries for the less common sources of data.

It looks like this impl will always check all enabled sources for availability.

I think this should use a loop of sequential checks with a short circuit to look for the first successful check instead of using `Promise.all`.  I believe `Promise.all` runs the promises sequentially under the hood anyways, so the performance of a looping sequential check with a short circuit should be the same or better.

src/helpers/daily-data-types/combined.tsx
```diff
@@ -15,138 +29,125 @@ let combinedTypeDefinitions: DailyDataTypeDefinition[] = [
...
+                    ["HealthConnect", "resting-heart-rate"],
+                ),
+                modifiedAfter,
+            );
         },
         labelKey: "resting-heart-rate",
         icon: <FontAwesomeSvgIcon icon={faHeartbeat} />,
         formatter: heartRateFormatter,
-        previewDataRange: [40, 100]
+        previewDataRange: [40, 100],
```
greinard, line 45: Seeing lots of these extra commas.  Is there a reason for that?

src/helpers/glucose-and-meals/glucose.ts
```diff
@@ -41,20 +43,40 @@ export async function googleFitBloodGlucoseDataProvider(startDate: Date, endDate
...
+        namespace: "HealthConnect",
+        type: "blood-glucose",
+        observedAfter: startOfDay(startDate).toISOString(),
+        observedBefore: endOfDay(endDate).toISOString()
+    };
+
+    const dataPoints = await queryAllDeviceDataV2(query);
+    return dataPoints.map(dataPoint => {
+        return {
+            timestamp: parseISO(dataPoint.observationDate!),
```
greinard, line 57: You may need to do something [similar to this](https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/helpers/time-range.ts#L105-L115) to get an accurate timestamp here since the dates and offsets are in separate properties.

More broadly, this may need to be done within all device data V2 data providers that are not using aggregates.  Otherwise, the data could be being applied to the wrong days.

chrisnowak: @greinard hmm...IIRC, *all* data coming in from V2 is local time, that was one of the major simplifications we made.  For purposes of determining which date something belongs to, you should be able to just use observationdate and not care about offset for all data types.  

(So I think we should be able to treat all V2 data basically like we do Fitbit V1 for instance)

src/helpers/daily-data-providers/combined-steps.ts
bxh: I think it is better to maintain a single code path. But i added a `useV2` parameter to make the second API call optional. See https://github.com/CareEvolution/MyDataHelpsUI/pull/429/commits/2a404b228a7e65c18a547488dd510dd619f93367.

src/helpers/daily-data-types/availability-check.ts
bxh: There is a better approach that doesn't involve a cache at all. I found a way to share data collection setting across availability checks. Let me know what you think: https://github.com/CareEvolution/MyDataHelpsUI/pull/429/commits/2a404b228a7e65c18a547488dd510dd619f93367.

ES2021 upgrade is necessary to use `Promise.any`.

src/helpers/daily-data-types/availability-check.ts
bxh: Good catch. See https://github.com/CareEvolution/MyDataHelpsUI/pull/429/commits/2eb92d28c2b0245f0f677e17168014dcb232b37e.

`Promise.any` can short-circuit input promises, but `Promise.all` executes them concurrently without guaranteeing their resolution order.

src/components/container/ViewDeviceActivity/ViewDeviceActivity.tsx
```diff
@@ -36,15 +37,28 @@ export default function (props: ViewDeviceActivityProps) {
...
         });
 
-        //instead of using Promise.all, we can show detail views when the first availability check returns true
-        //to avoid waiting for all of the checks
-        props.dataTypes.forEach(dataType => {
-            dataType.availabilityCheck().then(available => {
-                if (available) {
-                    setDeviceDataAvailable(true);
-                }
+        getCombinedDataCollectionSettings(true).then(combinedSettings => {
```
greinard, line 40: I think this can be simplified to:
```
getCombinedDataCollectionSettings(true).then(async combinedSettings => {
	for (const dataType of props.dataTypes) {
		if (await dataType.availabilityCheck(combinedSettings)) {
			setDeviceDataAvailable(true);
			return;
		}
	}
	setDeviceDataAvailable(false);
});
```

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     // Default
-    "target": "es2015",
+    "target": "es2021",
```
greinard, line 4: I'm not sure we can make this bump yet.  I could be mistaken, though.  From some quick Googling, it looks like `es2015` is supported by Safari back through iOS 10, but `es2021` requires iOS 15.4+.

I think this change would probably require some additional testing to be sure it works on all of our platforms and their supported versions.

src/helpers/daily-data-types/availability-check.ts
greinard: I don't think we can use `Promise.any` yet.  See my other comment RE the target version bump.

Also, the `Promise.any` impl just seems a lot less readable with the error throwing and such.  Why not just loop the promises?

src/helpers/glucose-and-meals/glucose.ts
greinard: Hmm.  That does makes sense, I think, but it could create inconsistencies with V1, right?  Please forgive the rambling here, but just thinking this through.

I'm fairly sure it would create inconsistencies for intraday data (like these glucose readings).  If we don't include the offset with the date when parsing, it would default to using the browser's offset.  If that doesn't match the data's offset, I think it would shift the data forward or backward some number of hours on the graph.  I'm trying to think through whether that could be an issue at the day boundaries for daily data as well.

As an concrete example, suppose a person has a glucose reading at 2pm on the east coast, and someone is viewing that person's glucose chart from the west coast later that day.  Maybe the person flew from NY to CA or maybe the viewer is a coordinator based in CA.  Should the glucose chart show the reading at 10am or at 2pm?  I can see arguments for both, but at present, with V1 data, I believe it would show at 10am when viewed from a browser on the west coast because the V1 data would have the offset included and the chart would be rendered relative to west coast time.  If the reading was recorded from a V2 source, and we ignore the offset, I think it would display at 2pm.  If we have readings from both a V1 and V2 source, things could be messy?

If we think the data should always display on charts in the participant's local time at the time the reading was recorded, regardless of where it gets viewed from, then maybe we need to start dropping the offsets from the V1 dates, too?  If we want to continue to chart data relative to the viewing (browser) timezone (as V1 data does today) then I think we'd need to include the V2 date offsets when parsing those dates.

Regarding the daily data boundaries.  I think a similar question exists.  Should a 2am step on the east coast on the 2nd day of the month be counted toward the 1st or 2nd day of the month when the chart is viewed from the west coast.  I think today, given how we are determining day keys, it would count toward the 2nd day of the month for both V1 and maybe V2 data, which is probably how it should be, but I'm also pretty sure that that same data point would render on an intraday chart at 10pm on the 1st of the month when viewed from the west coast, which is inconsistent.

tsconfig.json
bxh: iOS 15 was released four years ago. Three major versions older than the latest.

src/components/container/ViewDeviceActivity/ViewDeviceActivity.tsx
bxh: No, they are different. Your code calls `dataType.availabilityCheck` sequentially, whereas mine calls them concurrently.

src/helpers/daily-data-types/availability-check.ts
bxh: `Promise.any` resolves as soon as anyone of the input promises resolves, whereas looping depends on their order.

src/helpers/daily-data-types/availability-check.ts
bxh: Well, it is just a lint rule. Adding a comma even to the last element is the default.

src/helpers/glucose-and-meals/glucose.ts
chrisnowak: @greinard so @vladtar and @dherr and I went through this when we designed V2.  We determined that based on the various devices available and how they record data, it would be better to pretty much always just use local time at the time of recording.  The majority of integrations we have basically do it that way anyways: Fitbit / Garmin / Health Connect all do it that way; Fitbit/Garmin don't even have an offset I think, whereas Health Connect supplies an offset *in a separate field* (like our V2 data model does).  Apple Health is the only one that is a bit awkward - but our offset there is basically "Fake"; it is the offset at the time we upload data and *not the offset for time it's captured*, as Apple Health does not actually record an offset with it's data points.  So there's not much controlling for that, unfortunately.  (We do some fancy upserting to make sure we don't get duplicate records with Apple Health though)

So yeah, I think we move toward "where the user is is irrelevant"; as that is the case for Fitbit, Garmin, Health Connect - and even actually the Apple Watch itself does that as it's activity summaries are calculated midnight-midnight even if your "midnight" changes.

Keeping in mind also that coordinators may be viewing these in a geographically different area as well - and they would want to see the same thing as the participant.

More detailed:

> I think it would shift the data forward or backward some number of hours on the graph

Yes, that would just be a bug and we would want to guard against that...but again we can treat it like we do Fitbit data where there is no offset (ever)

> As an concrete example, suppose a person has a glucose reading at 2pm on the east coast, and someone is viewing that person's glucose chart from the west coast later that day. Maybe the person flew from NY to CA or maybe the viewer is a coordinator based in CA. Should the glucose chart show the reading at 10am or at 2pm?

We'd show it at 2pm

> Hmm. That does makes sense, I think, but it could create inconsistencies with V1, right?

I think we're OK if we have the minor inconsistency of local time when recorded vs. local time now.  And again I believe this only affects Apple Health data as I think that's the only type of data where we currently take that offset into account.  (Maybe also Google Fit...not sure if that brings an offset with it or if that is offset at time of upload...).  It may be something we want to adjust; but 

> Regarding the daily data boundaries. I think a similar question exists. Should a 2am step on the east coast on the 2nd day of the month be counted toward the 1st or 2nd day of the month when the chart is viewed from the west coast. 

I don't think this is a different question...we treat it like we do Fitbit intraday data as if the data does not have an offset at all.  So whatever date it falls in is the date it belongs to - in this case the 2nd day of the month.

greinard: Okay.   Makes sense.

Just to clarify.  That means we need to update any intraday usage of V1 data to ignore offsets then, too, correct?

And we probably need to update the sleep de-dup/overlap handling code to ignore offsets as well?

tsconfig.json
greinard: I think the age is less important than whether our iOS app still supports it.  However, it does look like there was some talk last month about dropping support for 15 and only supporting back to 16.  https://careevolution.slack.com/archives/C5Z45KQN5/p1739378890202979

Assuming that is the case, then we are probably good on this.  Still worth a little additional testing though.

src/helpers/glucose-and-meals/glucose.ts
chrisnowak: Yeah it seems like the right thing to do.  Not sure we need to drop everything to do it but something we should probably put on the backlog.

src/helpers/daily-data-types/availability-check.ts
greinard: I guess this is just one of the lint rules I don't happen to agree with.  Adding lint to reduce lint.  🙂

src/helpers/daily-data-types/availability-check.ts
greinard: I think we may actually want some ordering here though, to attempt to perform as few availability check queries to the server as possible.  The reason for the availability check wasn't to speed things up for the participant user.  It was to reduce the load on the server.  `Promise.any` will likely fire all the availability check fetches before any of them return a response, so it essentially makes `n` API queries every time instead of `1...n`.

From the participant perspective, the fastest way to see data would be for us to just query for the actual data from all sources to get it back as fast as we could.  The no-data scenario might take a tad longer to display in that scenario, but the with-data scenario would always be faster than doing availability checks followed by those same queries.

The availability checks are to throttle the load on the API.  If we always perform all of the availability checks every time, instead of checking them in prioritized sequential order, we lose some/much of the intended benefit of using the availability checkers in the first place.

greinard: As a related side note.  It looks like `DailyDataChart` and `DeviceDataMonthChart` would still query for all data even if there is no availability.  @chrisnowak - Do you know if that is/was intentional or is a bug?  The `RecentDailyDataBarChart` and the `RelativeActivityDayCoordinator` do not query for data unless there was availability, but the other two I mentioned look like they are?

greinard: Oh wait, the `RelativeActivityDayCoordinator` is also querying for all data types, even if they are not available?

chrisnowak: The point of availability checks is not really to reduce load on the server; though that is a nice side benefit I guess.  I am not sure there's considerable load on the server if you are to query and get zero results.

Actually the point is that we often want to display a thing if there is ANY data available vs. just data in the current interval.  For DailyDataChart for instance - if you have a sleep chart with "hideIfNoData = true", and you have no data for the *current week* but you have data for the *previous week*, we still want to display the chart so your charts don't all shift around from week to week.

chrisnowak: DailyDataChart could potentially bit a bit smarter about doing the availability check first though, as obviously if there's zero data available we can skip querying the month

greinard: > Actually the point is that we often want to display a thing if there is ANY data available vs. just data in the current interval. For DailyDataChart for instance - if you have a sleep chart with "hideIfNoData = true", and you have no data for the _current week_ but you have data for the _previous week_, we still want to display the chart so your charts don't all shift around from week to week.

Ah, got it.  Thank you for the clarification.  That was a misunderstanding on my part.

@bxh - Please ignore my suggestion here.  `Promise.any` should work fine for that then.  🙂

src/helpers/daily-data-types/availability-check.ts
```diff
@@ -1,15 +1,152 @@
...
+        }
+    });
+
+    try {
+        await Promise.any(availabilityChecks.map(promise => 
+            promise.then(result => {
+                if (result) {
+                    return true;
+                }
+                throw new Error("Result was falsy");
```
greinard, line 145: Could we adjust this to be something like `promise.then(result => result ? Promise.resolve() : Promise.reject())` instead of throwing errors?

I'm always hesitant to throw errors in expected and non-exceptional circumstances.

src/helpers/daily-data-types/availability-check.ts
bxh: Fixed.

src/helpers/daily-data-types/availability-check.ts
bxh: Done. Good suggestion.

src/helpers/daily-data-providers/combined-sleep.ts
bxh: Done.

src/helpers/daily-data-providers/combined-resting-heart-rate.ts
bxh: I tried using a class to manage all the `isXXXEnabled` queries, but since we're still figuring out the pattern, I switched to a helper function `getCombinedDataCollectionSetting` as a temporary solution. Each data provider now handles its own `isXXXEnabled` queries. You can find the discussion above between @chrisnowak and me.

src/components/container/DeviceDataMonthChart/DeviceDataMonthChart.stories.tsx
bxh: What is the newer format like?

greinard: Here is an example:  https://github.com/CareEvolution/MyDataHelpsUI/blob/main/src/components/presentational/Button/Button.stories.tsx

src/helpers/daily-data-providers/combined-resting-heart-rate.ts
greinard: Sorry, I was just referring generally to the line breaks/formatting.

For example:
```
providers.push(garminRestingHeartRateDataProvider(startDate, endDate));
```
vs
```
providers.push(
	appleHealthRestingHeartRateDataProvider(startDate, endDate),
);
```

Likewise, maybe, for some of these:
```
settings.queryableDeviceDataTypes.some(ddt => ddt.namespace === "AppleHealth" && ddt.type === "RestingHeartRate")
```

I think they tend to be okay on a single line as long as they aren't overly long.

In any case, I don't feel that strongly about it, and as long as I can hit auto-format without creating diffs, I'm fine with it.  🙂

src/helpers/daily-data-providers/combined-sleep-v2.ts
greinard: Doesn't hurt to have this, but `date-fns` never modifies the dates.  It always returns new ones.  So `currentDate` just gets re-assigned below without any modification to the original.

src/helpers/daily-data-providers/combined-resting-heart-rate.ts
bxh: Got it. I'm going to create a separate PR exclusively for formatting stuff.

src/helpers/relative-activity.tsx
```diff
@@ -27,30 +27,31 @@ export function queryRelativeActivity(startDate: Date, endDate: Date, dataTypes:
...
-                let dayKey = getDayKey(currentDate);
-                let value = dataTypeData?.[dayKey] ?? 0;
-                let threshold = (dataType.threshold === "30DayAverage" || dataType.threshold === undefined) ? calculatePrevious30DayAverage(dataTypeData, currentDate) : dataType.threshold;
-                if (threshold !== undefined) {
-                    let fillPercent = value / (threshold * 2);
-                    if (fillPercent > 1) {
-                        fillPercent = 1;
-                    }
+            if (results[index].status === "fulfilled") {
+                let dataTypeData = results[index].value;
```
reneefromhold, line 35: const

src/helpers/relative-activity.tsx
reneefromhold, line 1: Unit tests?

__tests__/helpers/relative-activity.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+    expect(queryDailyData).toHaveBeenCalledWith(
+      'steps',
+      add(startDate, { days: -1 }),
+      add(endDate, { days: 1 }),
+      false
+    );
+    expect(result).toEqual({
+      steps: {
+        [dayKey]: {
+          relativePercent: 0.25, // 5000 / (10000 * 2) = 0.25
```
reneefromhold, line 47: *clean out comments

bxh: This was on purpose. Better to explain why 0.25 is used.

__tests__/helpers/daily-data-providers/combined-resting-heart-rate.test.ts
```diff
@@ -92,37 +208,65 @@ export function buildOuraDataPoint(value: string, startDate: string, propertyVal
...
-        startDate: startDate,
-        type: ddpType,
-        value: value,
-        properties: {},
-    };
-    ddp.properties![propertyValueName] = value;
-    return ddp;
-}
\ No newline at end of file
+function buildOuraDataPoint(
```
reneefromhold, line 234: There is some similarity between this and it's combined-steps.test.ts counterpart. Can these be combined into one, more general helper for use across this and other tests

reneefromhold (commented):
@bxh The tests look good, thank you for adding them. I saw some helpers in the combine-resting-hr tests that looked like they could be generalized, and useful in other tests. 

I will defer to the other SME/reviewers for final approval.  lgtm otherwise

src/components/container/DeviceDataMonthChart/DeviceDataMonthChart.stories.tsx
greinard: You can ignore this comment.  The conversion to the new format is being done here: https://github.com/CareEvolution/MyDataHelpsUI/pull/448

src/helpers/daily-data-types/availability-check.ts
```diff
@@ -1,37 +1,165 @@
...
+            }
+
+            if (!enabled) return false;
+
+            let parameters: DeviceDataPointQuery | DeviceDataV2Query;
+
+            if (isV2) {
+                parameters = {
+                    namespace: namespace as DeviceDataV2Namespace,
+                    type: type.join(","),
```
greinard, line 122: The V2 API does not accept a comma separated type string.  It expects a single type per query.

bxh: Great catch. Fixed.

__tests__/helpers/daily-data-types/availability-check.test.ts
```diff
@@ -65,66 +113,129 @@ describe("simpleAvailabilityCheck", () => {
...
+        const checkAvailability = simpleAvailabilityCheck(
+            namespace,
+            multipleTypes,
+            true
+        );
+        const result = await checkAvailability(createTestContext(settings));
+
+        expect(result).toBe(true);
+        expect(queryDeviceDataV2).toHaveBeenCalledTimes(3);
+        expect(queryCompleted.Sleep).toBe(true);
```
greinard, line 206: Should there also be assertions here for `queryCompleted.Activity ===  true` and ` queryCompleted.Readiness === true` to prove that those types were queried as well?

__tests__/helpers/daily-data-types/availability-check.test.ts
```diff
@@ -65,66 +113,129 @@ describe("simpleAvailabilityCheck", () => {
...
+
+        await new Promise(resolve => setTimeout(resolve, 20));
```
greinard, line 208: What is this used for?

__tests__/helpers/daily-data-types/availability-check.test.ts
```diff
@@ -133,15 +244,123 @@ describe("simpleAvailabilityCheckV2", () => {
...
+        const result = await checkAvailability(createTestContext(settings));
+
+        const elapsedTime = Date.now() - startTime;
+
+        expect(result).toBe(true);
+
+        // Should resolve faster than the slowest source (Fitbit: 500ms)
+        // but slower than or around the fastest successful source (Oura: 100ms)
+        // Adding some buffer for test overhead
+        expect(elapsedTime).toBeLessThan(600);
```
greinard, line 340: Based on the comment, shouldn't this be 500 or less?

src/helpers/daily-data-types/availability-check.ts
```diff
@@ -1,37 +1,164 @@
...
+                    const parameters: DeviceDataV2Query = {
+                        namespace: namespace as DeviceDataV2Namespace,
+                        type: t,
+                        limit: 1,
+                        modifiedAfter: modifiedAfter?.toISOString()
+                    };
+
+                    const result =
+                        await MyDataHelps.queryDeviceDataV2(parameters);
+                    return await (result.deviceDataPoints.length > 0
```
greinard, line 128: No need for the extra `await` here.  The `async` would just re-wrap the `true` value, and it'll process a returned `Promise.reject()` the same way.

`return result.deviceDataPoints.length > 0 || Promise.reject();`

src/helpers/daily-data-types/availability-check.ts
```diff
@@ -1,37 +1,164 @@
...
+                        ? Promise.resolve(true)
+                        : Promise.reject());
+                });
+
+                try {
+                    const result = await Promise.any(promises);
+                    return result;
```
greinard, line 135: Minor:  This can be inlined.

src/helpers/daily-data-providers/combined-data-collection-settings.ts
chrisnowak: I do like that a bit better as well now that I think about it (and maybe that is what we should just do server-side as well)

src/helpers/daily-data-providers/combined-sleep-v2.ts
```diff
@@ -1,39 +1,81 @@
...
 
-export default async function (startDate: Date, endDate: Date): Promise<{ [key: string]: number }> {
-    let providers: Promise<{ [key: string]: number }>[] = [];
+export default async function (
+    startDate: Date,
+    endDate: Date
+): Promise<Record<string, number>> {
+    const providers: Promise<Record<string, number>>[] = [];
+
+    const useV2 = true;
```
chrisnowak, line 15: Seems like you don't really need a var for this?

chrisnowak (approved):
@bxh a couple final comments and it looks like @greinard had a few as well but this looks good to me.  Ended up being a lot bigger scope than I think either of us imagined originally!

src/helpers/daily-data-types/availability-check.ts
```diff
@@ -1,37 +1,164 @@
...
+                        );
+                    break;
+                case "Fitbit":
+                    enabled = settings.fitbitEnabled;
+                    break;
+                case "Garmin":
+                    enabled = settings.garminEnabled;
+                    break;
+                case "Oura":
+                    enabled = settings.ouraEnabled;
```
greinard, line 97: It looks like this enabled flag will need to be more like the health connect one below.  It will need to check the V2 data types as well.

See more info here: https://github.com/CareEvolution/MyDataHelpsUI/pull/461#pullrequestreview-2751003526

src/helpers/daily-data-providers/combined-sleep.ts
```diff
@@ -1,45 +1,78 @@
...
+        )
+    ) {
+        providers.push(appleHealthSleepDataProvider(startDate, endDate));
+    }
+    if (settings.ouraEnabled) {
+        providers.push(ouraSleepMinutesDataProvider(startDate, endDate));
+    }
+    if (
+        settings.healthConnectEnabled &&
+        deviceDataV2Types.some(
```
greinard, line 40: For the time being, all of the V2 data type checks will need to include the `enabled` flag on the data type.

I have submitted this PR: https://github.com/CareEvolution/MyDataHelps.js/pull/157

Once that is available, we should be able to call `MyDataHelps.getDeviceDataV2AllDataTypes(true)` within `getCombinedDataCollectionSettings` to get back only the enabled data types.  Then we can drop all the extra enabled flag checks if we want to.

__tests__/helpers/daily-data-types/availability-check.test.ts
bxh: Added.

__tests__/helpers/daily-data-types/availability-check.test.ts
bxh: Great catch. This was just a bug. I used `Promise.all` instead of `Promise.any` for availability check by mistake.

src/helpers/daily-data-types/availability-check.ts
bxh: Fixed.

src/helpers/daily-data-providers/combined-sleep-v2.ts
bxh: Fixed.

src/helpers/daily-data-providers/combined-sleep.ts
bxh: Thanks @greinard. I've incorporated your improvement to my PR.

src/helpers/daily-data-types/availability-check.ts
bxh: Updated.

PR #432, opened 2025-02-26
Adjust locale selection.

greinard:
## Overview

This branch adjusts how locales are selected when formatting numbers, dates, etc.

Specifically, it adjusts the logic such that locale specificity can only be improved by using the browser's language setting when that language's code matches the MDH language code.

Without this change, when the MDH language is set to something like `de` (German), it gets ignored during locale selection if the browser is not also set to some German variant.

## Security

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

- N/A

### DIFF
--- a/__tests__/helpers/locale/locale.test.ts
+++ b/__tests__/helpers/locale/locale.test.ts
@@ -31,6 +31,12 @@ describe('Locale Tests', () => {
             const result = getIntlLocale();
             expect(result).toBe("es-co");
         });
+        it('Should not use browser locale if it is set to a different language.', () => {
+            mockMDHLanguage = "en";
+            jest.spyOn(navigator, 'language', 'get').mockReturnValue('es-CO');
+            const result = getIntlLocale();
+            expect(result).toBe("en");
+        });
         it('Should return base language if no browser language.', () => {
             mockMDHLanguage = "en";
             jest.spyOn(navigator, 'language', 'get').mockReturnValue('');
@@ -68,11 +74,11 @@ describe('Locale Tests', () => {
             const result = getDateLocale();
             expect(result).toBe(enUS);
         });     
-        it('Should use browser locale even if it set to a different (supported) language.', () => {
+        it('Should not use browser locale if it is set to a different language.', () => {
             mockMDHLanguage = "en";
             jest.spyOn(navigator, 'language', 'get').mockReturnValue('es-es');
             const result = getDateLocale();
-            expect(result).toBe(es);
+            expect(result).toBe(enUS);
         });
         it('Should use base language if browser is set to an unsupported language.', () => {
             mockMDHLanguage = "en";
--- a/src/helpers/locale.ts
+++ b/src/helpers/locale.ts
@@ -9,23 +9,21 @@ import { toDate } from "./date-helpers";
 //
 // Priority for locale selection:
 // 1. If MDH locale has a country code (like "en-AU") use that.
-// 2. If MDH locale does NOT have a locale code (like "en"), use the browser locale. 
-//    The browser often has more locale specificity than MDH.
+// 2. If MDH locale does NOT have a locale code (like "en"), use the browser locale
+//    as long as the language code matches. The browser often has more locale
+//    specificity than MDH.
 // 3. Fall back to the original language string.
 export function getIntlLocale() : string {
     // Intl libraries don't support underscores, so it needs "en-US" not "en_US".
     const lang =`${MyDataHelps.getCurrentLanguage()}`.replace("_", "-");
-
     if (lang.length < 2) return "en-us";
 
-    let countryCode = getCountryCodeFromIso(lang);
-    let intlLocale;
-    
-    if (countryCode) {
-        intlLocale = lang;
-    }
-    else {
-        intlLocale = navigator?.language || lang;
+    const languageCode = getLanguageCodeFromIso(lang);
+    const countryCode = getCountryCodeFromIso(lang);
+
+    let intlLocale = lang;
+    if (!countryCode && navigator?.language && getLanguageCodeFromIso(navigator.language) === languageCode) {
+        intlLocale = navigator.language;
     }
     return intlLocale.toLowerCase();
 }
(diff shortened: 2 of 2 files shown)


lynnfaraday:
@chrisnowak This is different than what we discussed when we implemented the date locale selection initially. I thought the entire purpose of using the browser's locale selection for dates was because it would likely be more specific than the MDH selection? 

For example, a European user might have their browser set to German with European date and time formats. It would be jarring to them to see dates in the US format (MM/DD/YY rather than the expected DD/MM/YY) just because a specific project doesn't support German.

lynnfaraday:
Follow-up - I realize it's not ideal to have the month names or "Last Sunday..." translated differently from the rest of the screen. When that came up previously, the sentiment was that it would be better to rework how the language selection was implemented. 

E.g., if the user's preferred (device) language is German, we could have the general skeleton of MDH (including all built-in views) show in German rather than limit it based on what each project supports. That's more consistent with how other apps displaying user-generated content work. The app itself (and all pre-built views) will be in your preferred language, but project-generated content may not be.

greinard:
> For example, a European user might have their browser set to German with European date and time formats. It would be jarring to them to see dates in the US format (MM/DD/YY rather than the expected DD/MM/YY) just because a specific project doesn't support German.

I may be missing a piece to the puzzle here, so my apologies if that is the case, but I believe the `MyDataHelps.getCurrentLanguage()` function returns the browser's language if a `lang` parameter is not present and the MDH language has not been set manually using `MyDataHelps.setCurrentLanguage(...)`.  https://github.com/CareEvolution/MyDataHelps.js/blob/master/src/MyDataHelps.ts#L1210

The `MyDataHelps.setCurrentLanguage(...)` function only updates the MDH language if it gets called with a supported language. https://github.com/CareEvolution/MyDataHelps.js/blob/master/src/MyDataHelps.ts#L1190

So I think in the scenario you mentioned, where German is not supported, if something attempts to call `MyDataHelps.setCurrentLanguage(...)` with German, nothing would happen, and `MyDataHelps.getCurrentLanguage()` would still return the browser's language.

On the other hand, if German is supported, and `MyDataHelps.setCurrentLanguage('de')` gets called or `lang=de` is included on the URL in a browser that is set to `en-US`, I think the prior logic would cause all dates and numbers to be formatted in `en-us` rather than `de`.

lynnfaraday:
> I may be missing a piece to the puzzle here, so my apologies if that is the case, but I believe the `MyDataHelps.getCurrentLanguage()` function returns the browser's language if a `lang` parameter is not present and the MDH language has not been set manually using `MyDataHelps.setCurrentLanguage(...)`. https://github.com/CareEvolution/MyDataHelps.js/blob/master/src/MyDataHelps.ts#L1210

Likewise I may be missing something, but I don't believe that's the behavior—at least not across all browsers/devices. For example - this is what I see when my browser is set to Spanish while viewing a project that doesn't have Spanish enabled. Everything uses English. 

image

The original design respected the browser's locale for dates and numbers, which ensures that those things are displayed as expected even when the project doesn't support the user's chosen language.

This is particularly important with numbers, since comma vs. decimal separators can make a huge difference in interpretation. Dates can also be confusing if you're used to DD/MM format and MDH switches to MM/DD format.

greinard:
I'm not sure it is any less jarring to have English text and German days/dates/numbers.  If someone understands English enough to read it, they probably recognize the number/date difference, just like English speakers do when looking at German dates/numbers.  And we are only doing that if the MDH language is not country specific.  If it is, we are accepting the MDH language setting and using it for everything.

In this example, I have my browser set to German, which is not a supported language for this project, and chose Spanish (which is supported) from the drop down menu.  The app translations are Spanish, but the Day of the week and the date are German.  If I choose to specify Spanish as the language, I think it should all be Spanish, even if my specification of Spanish is `es` rather than `es-ES`.

image

lynnfaraday:
@greinard I don't disagree with you that there are situations that are jarring with the current implementation, I'm just saying there are other situations equally jarring if we do it the other way. This was discussed when implementing the date/numbers localization PR. You can see this decision reflected in the unit tests, QA test cases, PR description, etc. It wasn't an oversight, it was a conscious choice to make it work this way.

Also, this really shouldn't come up in the mobile apps because there is no language selection dropdown - it only uses the device setting. And on the web app, the language selection dropdown behaves a bit weird currently anyway. I think that's a better place to focus our usability efforts.

As for whether we should change it - I'll defer to Chris, since he was the one who weighed in on the original design.

greinard:
I'm not suggesting it was an oversight, or that it wasn't implemented as spec'd.  But that doesn't mean that we can't adjust or improve upon that spec.

I think there are two questions here.

One is, do we want to fall back to unsupported languages for days/dates/numbers when the browser is configured with a unsupported language.  I tend to think the answer should be no, because I think it would be odd if I visited or used a German, French, or Spanish website or app, where all the other content was in those languages, yet it displayed days/dates/times in English.  Others may disagree, and that's okay.  I don't think that is what this PR is really about, though.

I think the second question, and the one that this PR addresses, is should we fall back to the browser language when the MDH language is set to `es`, but NOT when it is set to `es-ES`.  Again, I think the answer should be no.  The presence of a country code on the MDH language setting should not be what determines whether we actually swap to a completely different language.  It is one thing to improve upon the MDH language specificity by getting the country code from the browser language when the base language matches, but quite another to change the language completely just because the MDH language did not have a country specific code on it.

The `MyDataHelps.getCurrentLanguage()` implementation already supports falling back to the browser language when an MDH language hasn't been set.  In the scenarios where we want that fallback behavior, we can just not set an MDH language at all and it should already behave that way (i.e. English text with German/French/Spanish dates).

lynnfaraday:
See convo in #pe slack for context on the original design decision, and discussion of possible alternatives.

lynnfaraday (approved):

chrisnowak:
@greinard So one scenario here that I want to check on:
- User's browser set to en_us
- User switches language to "es" in our language switcher
- What is the resulting locale?

I think just "es" locale may format dates like europe - DD/MM/YYYY .  Whereas because this is a person in the US we actually want to make sure the locale us es_us and dates are formatted MM/DD/YYYY

greinard:
Per our offline discussion.  No special handling of Spanish dates should be necessary at this time to have the same formatting behavior the browser uses for `es-US` users.

PR #437, opened 2025-03-07
GlucoseChart: Updated handling of data gaps and partial data when meals are present/selected.

greinard:
## Overview

This branch fixes a couple issues with the `GlucoseChart` related to partial data when meals are present.

Scenario:
- You have some glucose data for the morning hours only, but logged a meal in the afternoon. 

The issues:
- The chart was connecting the graphed line from the final glucose reading over to the meal.  This was because the chart was using a single series to represent both the glucose readings and the meals.  As such, it treated the meal as if it were another glucose reading, which is incorrect.
- If you selected the afternoon meal on the meal log to zoom in on the glucose chart, it would crash because no glucose readings were present during the zoomed time period in order to compute the meal's vertical position.

Testing:
- I have updated the `GlucoseChart` story so you can test these scenarios there.
- Select the "with partial data" state.
- Verify that the line does not extend over to meals 3 and 4.
- Click meal 3 or meal 4 to zoom in.  Verify that it does not crash.
- You can also select the "with gap in data" state, select meal 3, and verify that it zooms without crashing with a partial gap in the data.
## Security

- No security risk.  Just adjusting the chart rendering to gracefully handle data gaps and partial data when meals are present.

### Testing
- [ ] MyDataHelps iOS app
- [ ] MyDataHelps Android app

### Documentation

### DIFF
--- a/src/components/container/GlucoseChart/GlucoseChart.previewData.ts
+++ b/src/components/container/GlucoseChart/GlucoseChart.previewData.ts
@@ -1,6 +1,6 @@
 import { generateGlucose, generateSteps, Reading } from '../../../helpers';
 
-export type GlucoseChartPreviewState = 'no data' | 'with data';
+export type GlucoseChartPreviewState = 'no data' | 'with gap in data' | 'with partial data' | 'with data';
 
 export interface GlucoseChartPreviewData {
     glucose: Reading[];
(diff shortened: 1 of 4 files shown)


chrisnowak (approved):
Looks awesome, thanks @greinard 

One minor other thing I noticed that might be easy to fix.

Since there can be a LOT of glucose data, sometimes I will say that the graph can be a little slow to re-render when you come back from editing a meal; at least maybe when I was on a slower connection.  It feels like it adds a little bit of slowness/friction after meal logging which we want to be as easy as possible

Wondering if we can skip setting the graph into the "loading" state on applicationDidBecomeVisible so that it just re-renders when it has new data but the old data still shows until it re-renders; like a lot of other components

greinard:
> Wondering if we can skip setting the graph into the "loading" state on applicationDidBecomeVisible so that it just re-renders when it has new data but the old data still shows until it re-renders; like a lot of other components

Yep, that should be fine.  I've pushed that change, too.  You can see it in the storybook by just swapping states.  It'll load in the new state without showing the loading indicator.  Only the initial load will use the loading indicator.

Issue #420, opened 2025-02-18
Review end date inclusivity when date bounding daily data results.


greinard:
It seems we may have some inconsistency across the codebase related to date bounding of daily data results.

These tend to take the form of something like the following:
```
const result: DailyDataQueryResult = {};

let currentDate = startDate;
while (currentDate < endDate) {
	const dayKey = getDayKey(currentDate);
	// Populate result[dayKey]
	...
	currentDate = add(currentDate, { days: 1 });
}

return result;
```

However, in some cases we are also using `currentDate <= endDate` as the condition on the loop.

A simple non-exhaustive search to get an idea of what I'm referring to: https://github.com/search?q=repo%3ACareEvolution%2FMyDataHelpsUI%20%22while%20(currentDate%22&type=code

Note that the variable names may not match the above exactly.  Sometimes it may be `while (startDate < endDate)`, etc.  It may take some manual poking around to identify all the places this logic exists.

I think it would be worth examining these existing date bounding conditions to see if we need to update any of them and perhaps settle on one or the other as the standard.  Do we want the end dates to be inclusive or exclusive?  And should that decision be universally applied or be data type specific?

PR #456, opened 2025-03-31
Refactor, update, and unit test daily data providers.

greinard:
## Overview

My apologies in advance for the size of this PR.  Unfortunately, we have a lot of daily data providers now, so any sort of cross cutting refactor results in a lot of files being touched.  The addition of unit tests for all of them nearly doubles the number of files in the diff.  Hopefully this refactoring and these updates will result in far less cross cutting updates if/when any future adjustments are needed to the common daily data related functions.

There are a few reasons for the refactoring an updates on this branch:
- To eliminate repetitive (sometimes with slight modification) boilerplate code for querying data points and building and combining daily data results.
- To consistently apply start and end date filtering to both the data point queries and the results being returned by daily data providers (See: https://github.com/CareEvolution/MyDataHelpsUI/issues/420).  Previously, we had some providers using inclusive start dates with exclusive end dates, others using inclusive start dates with inclusive end dates, and others using neither and just relying on the API query date filtering.  I chose to use inclusive start and end dates for everything.  This means if you query for -6 days ago through today, you'll also get today's data.
- To consistently handle (ignore) date/time offsets for both V1 and V2 data points.  Prior to this update, V1 data points could shift (including across day boundaries) on charts depending on when they occurred and your current viewing timezone.  (See: https://github.com/CareEvolution/MyDataHelpsUI/pull/429#discussion_r1999677973)
- To add unit test coverage to all of the existing daily data providers.

Other important notes:
- I have not yet added unit tests for the combined data providers.  I wanted to hold off on that until https://github.com/CareEvolution/MyDataHelpsUI/pull/429 gets merged because there are a decent amount of overlapping changes there and I figured the actual code diff will be tricky enough to merge.  I didn't want to have to do the same for the unit tests, too.  Once that PR gets merged, I'll pull those changes out to this branch and then add/update the combined provider UTs if/as needed.
- A few daily data providers only expect a single data point per day when they query.  They were looping all data and essentially taking the "last" value they encountered from each day.  If, by chance, there was more than one data point for a given day, they would arbitrarily use the last one from the API response.  I have updated these to use a more deterministic approach.  If there happens to be more than one data point for a given day, they will now use the value from the data point that has the more recent modified date on it to break the tie.
- This PR does not contain all the updates necessary to consistently handle date offsets for all components.  The scope of the PR was getting huge, so I punted on those other pieces for the time being.  This PR tries to focus on the largest bit, the daily data providers.  I'll have follow up PR's for some other things like Meals, Asthma, and maybe a couple more.

Other less important notes:
- `getDayKey` was updated to accept either a `string` or a `Date`.  It is no longer necessary to parse date strings in order to get a day key.  We have a number of places that are performing day key generation on their own.  I will update those to use the common function as part of some follow up PRs.

## Security

DESCRIBE BRIEFLY WHAT SECURITY RISKS YOU CONSIDERED, WHY THEY DON'T APPLY, OR HOW THEY'VE BEEN MITIGATED.

- No new security risk.  Just refactoring, minor updates, and unit testing of existing data querying logic.

## Testing

DESCRIBE YOUR TEST PLAN

- Much of this could be regression tested locally using storybook, but for non-devs, it would require a ViewBuilder PR.  I will work on getting on of those set up with a snapshot version of MDH-UI.

### Documentation


chrisnowak:
Nice, this pass was definitely needed as we were drifting into inconsistency.  I won't be able to review it depth for a couple weeks (out next week) but my initial first pass looked good

bxh:
I'll review this next Monday.

greinard:
All changes merged from `main`, combined UTs added/updated.

src/helpers/daily-data-providers/query-all-device-data.ts
```diff
@@ -1,22 +1,26 @@
...
+                allDataPoints.push(...page.deviceDataPoints);
+            }
+        } catch {
+            // ignore.
+        }
+
+        return allDataPoints;
+    }
+
+    return getDeviceData();
```
bxh, line 25: Use do...while.

```suggestion
export default async function queryAllDeviceData(query: DeviceDataPointQuery): Promise<DeviceDataPoint[]> {
    const queryParameters: DeviceDataPointQuery = { ...query };
    const allDataPoints: DeviceDataPoint[] = [];
    
    try {
      let page;
      do {
        page = await MyDataHelps.queryDeviceData(queryParameters);
        allDataPoints.push(...page.deviceDataPoints);
        queryParameters.pageID = page.nextPageID;
      } while (page.nextPageID);
    } catch {
    }

    return allDataPoints;
```

src/helpers/daily-data-providers/query-garmin-total-daily-values.ts
```diff
@@ -1,46 +1,13 @@
...
-					data[dayKey] = val;
-				}
-			});
-		});
-		return data;
-	});
+import { DailyDataQueryResult } from "../query-daily-data";
+import { buildTotalValueResult, getStartDate, queryForDailyData } from "./daily-data";
+
+export default async function (types: string[], startDate: Date, endDate: Date, divideBy?: number): Promise<DailyDataQueryResult> {
```
bxh, line 4: `divideBy:number = 1`

src/helpers/daily-data-providers/query-oura-sleep.ts
```diff
@@ -1,30 +1,23 @@
...
-			}	
-			if (!d.properties[type]) { return; }
-			if (parseInt(d.properties[type]) <= 0) { return; }
-			var dayKey = d.properties['day'];
-			const existingValue = data[dayKey] || 0;
-			data[dayKey] = Math.round(Number.parseFloat(d.properties[type])) + existingValue;
-		});
-		return data;
-	});
+export default async function (startDate: Date, endDate: Date, dataType: string, sleepTypesToInclude: OuraSleepType[], divideBy?: number): Promise<DailyDataQueryResult> {
```
bxh, line 6: `divideBy:number = 1`

src/helpers/daily-data-providers/query-oura-sleep.ts
```diff
@@ -1,30 +1,23 @@
...
+    const dailyData = await queryForDailyDataV2("Oura", "sleep", startDate, endDate, dataPoint => dataPoint.properties?.["day"]);
+
+    const filteredDailyData = Object.keys(dailyData).reduce((filteredDailyData, dayKey) => {
```
bxh, line 9: Use a different variable name inside to avoid confusion.
```suggestion
    const filteredDailyData = Object.keys(dailyData).reduce((acc, dayKey) => {
```

src/helpers/date-helpers.ts
```diff
@@ -204,3 +204,7 @@ export function getShortTimeOfDayString(dateOrDateString: Date | string) {
 
 		return formatDateForLocale(date, formatOptions.hour12 ? "h a" : "p");
 }
+
+export function parseISOWithoutOffset(dateStr: string): Date {
+	return parseISO(dateStr.substring(0, 19));
```
bxh, line 209: Add a comment to explain 19.

src/helpers/get-day-key.ts
```diff
@@ -1,5 +1,8 @@
 import { formatISO } from "date-fns";
 
-export default function getDayKey(date: Date) {
-	return formatISO(date).substr(0, 10);
+export default function getDayKey(date: string | Date) {
+    if (typeof date === "string") {
+        return date.substring(0, 10);
```
bxh, line 5: Add a comment to explain 10.

src/helpers/insight-matrix/survey-data-type.ts
```diff
@@ -28,14 +27,14 @@ export function getSurveyDataProvider(dataType: SurveyDataType): DailyDataProvid
             query.resultIdentifier = dataType.resultIdentifier;
         }
 
-        const answers = await queryAllSurveyAnswers(query);
+        const answers = await queryAllSurveyAnswers(query).then(answers => answers.sort((a, b) => compareDesc(a.date, b.date)));
```
bxh, line 30:
```suggestion
        const answers = await queryAllSurveyAnswers(query);
        answers.sort((a, b) => compareDesc(a.date, b.date));
```

src/helpers/insight-matrix/survey-data-type.ts
```diff
@@ -28,14 +27,14 @@ export function getSurveyDataProvider(dataType: SurveyDataType): DailyDataProvid
...
 
         const result: DailyDataQueryResult = {};
 
         let currentDate = startDate;
         while (currentDate <= endDate) {
             const currentDayKey = getDayKey(currentDate);
-            const answerForDate = answers.find(answer => isSameDay(parseISO(answer.date), currentDate))?.answers[0];
+            const answerForDate = answers.find(answer => getDayKey(answer.date) === currentDayKey)?.answers[0];
```
bxh, line 37:
```suggestion
            const answerForDate = answers.find(answer => getDayKey(answer.date) === currentDayKey)?.answers?.[0];
```

src/helpers/query-all-device-data-v2-aggregates.tsx
```diff
@@ -1,26 +1,26 @@
...
 
     async function getDeviceDataV2Aggregates(): Promise<DeviceDataV2Aggregate[]> {
-        let page = await getDeviceDataV2AggregatesPage();
-        let allAggregates = page.intervals;
+        const allAggregates: DeviceDataV2Aggregate[] = [];
 
-        while (page.nextPageID) {
-            page = await getDeviceDataV2AggregatesPage(page.nextPageID);
-            allAggregates = allAggregates.concat(page.intervals);
+        try {
```
bxh, line 9: Use do...while as above.

src/helpers/query-all-device-data-v2.tsx
```diff
@@ -1,26 +1,26 @@
...
 
     async function getDeviceDataV2(): Promise<DeviceDataV2Point[]> {
-        let page = await getDeviceDataV2Page();
-        let allDataPoints = page.deviceDataPoints;
+        const allDataPoints: DeviceDataV2Point[] = [];
 
-        while (page.nextPageID) {
-            page = await getDeviceDataV2Page(page.nextPageID);
-            allDataPoints = allDataPoints.concat(page.deviceDataPoints);
+        try {
```
bxh, line 9: Use do...while.

bxh (changes requested):
Great work on consolidating the numerous code paths--this is a solid step forward in terms of maintainability and clarity. 👍 

That said, I noticed a few minor issues that should be addressed before merging. most importantly, I think it is helpful to define a comprehensive list of daily data types.

src/helpers/daily-data-providers/apple-health-heart-rate-range.ts
```diff
@@ -1,43 +1,20 @@
...
+        const maxHeartRate = maxHeartRateDataPoints.length > 0 ? Math.max(...maxHeartRateDataPoints.map(getFloatValue)) : 0;
+
+        const minHeartRateDataPoints = dailyData[dayKey].filter(dataPoint => dataPoint.type === "HourlyMinimumHeartRate");
+        const minHeartRate = minHeartRateDataPoints.length > 0 ? Math.min(...minHeartRateDataPoints.map(getFloatValue)) : 0;
+
+        if (maxHeartRate && minHeartRate) {
+            result[dayKey] = maxHeartRate - minHeartRate;
+        }
+    });
+    return result;
```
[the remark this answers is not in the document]
greinard: Done.

src/helpers/daily-data-providers/apple-health-steps-while-wearing-device.ts
```diff
@@ -38,19 +38,15 @@ export default async function (startDate: Date, endDate: Date): Promise<DailyDat
...
-        type: 'HourlySteps',
-        observedAfter: add(startDate, { days: -1 }).toISOString(),
-        observedBefore: add(endDate, { days: 1 }).toISOString()
-    }).then(dataPoints => dataPoints.filter(dataPoint => dataPoint.startDate && parseInt(dataPoint.value) > 0));
+    const dailyData = await queryForDailyData('AppleHealth', 'HourlySteps', startDate, endDate, getStartDate);
 
-    const result: DailyDataQueryResult = {};
-    hourlyStepsDataPoints.forEach(dataPoint => {
-        const dayKey = getDayKey(parseISO(dataPoint.startDate!));
+    const filteredDailyData = Object.keys(dailyData).reduce((filteredDailyData, dayKey) => {
```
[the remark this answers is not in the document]
greinard: Given that only the keys are necessary for this filtering, I do think the entries/fromEntries version could be slightly less performant in extreme cases, due to using more memory for the entries array and because it performs the result building in two passes instead of during single one, but I do find it easier to read.  I think for our use case, the performance difference, if any, will be negligible, so I've made this change.

src/helpers/daily-data-providers/combined-sleep-v2.ts
```diff
@@ -1,82 +1,35 @@
...
-            });
-
-        if (sleepMinutes) {
-            data[dayKey] = sleepMinutes;
-        }
-
-        currentDate = add(currentDate, { days: 1 });
-    }
+    if (providers.length === 0) return {};
+    if (providers.length === 1) return providers[0];
```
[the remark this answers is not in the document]
greinard: It isn't necessary, but it does reduce the call stack and avoids some additional looping by not calling the combine results function, which is not necessary when you have a single provider result.

src/helpers/daily-data-providers/daily-data.ts
```diff
@@ -0,0 +1,186 @@
...
+export type DailyDataValueFunction = (dataPoint: DeviceDataPoint | DeviceDataV2Point) => number;
+export const getFloatValue: DailyDataValueFunction = dataPoint => parseFloat(dataPoint.value);
+export const getIntValue: DailyDataValueFunction = dataPoint => parseInt(dataPoint.value);
+
+export type DailyData = Record<string, DeviceDataPoint[]>;
+export type DailyDataV2 = Record<string, DeviceDataV2Point[]>;
+
+export async function queryForDailyData(
+    namespace: DeviceDataNamespace,
+    type: string | string[],
```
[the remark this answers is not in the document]
greinard: Avoiding the typos would be great, but it would only ensure that the type being passed to the function was valid for *some* data provider.  It wouldn't necessarily indicate that it was valid for the namespace or the data provider being added/modified.  In other words, someone could typo "steps" instead of "sleep", and it would still compile, but could be way wrong for the situation.

I think some unit tests on the calling code to ensure they are passing the correct type values is probably more likely to catch those mistakes than having all the values enumerated in a list.

I'm not totally opposed to the idea if someone wants to give that a try on a branch to see how it looks, but I'd prefer not to include that as part of this PR.

src/helpers/daily-data-providers/daily-data.ts
```diff
@@ -0,0 +1,186 @@
...
+    return combineResults(startDate, endDate, resultsToCombine, values => Math.round(values.reduce((a, b) => a + b) / values.length));
+}
+
+function combineResults(startDate: Date, endDate: Date, resultsToCombine: DailyDataQueryResult[], computeValue: (values: number[]) => number): DailyDataQueryResult {
+    const combinedResult: DailyDataQueryResult = {};
+
+    let currentDate = startDate;
+    while (currentDate <= endDate) {
+        const dayKey = getDayKey(currentDate);
+        const values = resultsToCombine.filter(result => result[dayKey]).map(result => result[dayKey]);
```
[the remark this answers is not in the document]
greinard: I think the one liner with filter/map is easier to follow for this one than using reduce.

src/helpers/daily-data-providers/fitbit-elevated-heart-rate.ts
```diff
@@ -1,41 +1,34 @@
...
+
+    const filteredDailyData = Object.keys(dailyData).reduce((filteredDailyData, dayKey) => {
+        const filteredDataPoints = dailyData[dayKey].filter(dataPoint => {
+            return dataPoint.value !== "Out of Range" && (!zone || dataPoint.value === zone);
+        });
+        if (filteredDataPoints.length > 0) {
+            filteredDailyData[dayKey] = filteredDataPoints;
+        }
+        return filteredDailyData;
+    }, {} as DailyData);
```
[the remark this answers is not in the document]
greinard: Done.

src/helpers/daily-data-providers/garmin-sleep.ts
```diff
@@ -1,57 +1,34 @@
...
-        });
-        return data;
+import { DailyDataQueryResult } from "../query-daily-data";
+import { buildMaxValueResult, getSleepDate, queryForDailyData } from "./daily-data";
+
+async function querySleep(property: string, startDate: Date, endDate: Date, divideBy?: number): Promise<DailyDataQueryResult> {
+    const dailyData = await queryForDailyData("Garmin", "Sleep", startDate, endDate, getSleepDate);
+    return buildMaxValueResult(dailyData, dataPoint => {
+        const value = parseFloat(dataPoint.properties?.[property] ?? "0");
+        return divideBy ? (value / divideBy) : value;
```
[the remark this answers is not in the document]
greinard: Done.

src/helpers/query-all-device-data-v2-aggregates.tsx
greinard: Done.

src/helpers/query-all-device-data-v2.tsx
greinard: Done.

src/helpers/daily-data-providers/query-oura-sleep.ts
greinard: Done.

src/helpers/daily-data-providers/query-garmin-total-daily-values.ts
greinard: Done.

src/helpers/daily-data-providers/garmin-sleep.ts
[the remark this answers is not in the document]
greinard: I went back and reverted these `divideBy` default adjustments.  I forgot that the reason I originally implemented it this way was to protect against dividing by 0.

With the default of 1, it does read a little easier, but it also allows users to pass in 0 and cause an exception.  With the conditional implementation, 0 gets treated like 1, and the value just gets returned without modification.

src/helpers/daily-data-providers/query-garmin-total-daily-values.ts
greinard: Reverted.  See above comment.

src/helpers/daily-data-providers/query-oura-sleep.ts
greinard: Reverted. See above comment.

src/helpers/daily-data-providers/query-oura-sleep.ts
greinard: I tend to like using the same name when reducing to build something.  It ties together the fact that the `filteredDataPoints` being constructed within the reduce is the same object that gets returned and assigned to the `filteredDataPoints` const.

src/helpers/date-helpers.ts
greinard: Done.

src/helpers/insight-matrix/survey-data-type.ts
greinard: I'd like to keep those together.  I don't see a reason to separate the query and the sorting since everything after that should be operating on the sorted list anyways.

I could add a line break before the `.then` if you think that reads better?

src/helpers/insight-matrix/survey-data-type.ts
greinard: The definition for the `answers` field is `answers: string[];`, so if an answer is found for the current day key, `answers` should always be defined.

Have you encountered situations where that wasn't the case?

src/helpers/daily-data-providers/daily-data.ts
```diff
@@ -0,0 +1,186 @@
+import { DeviceDataNamespace, DeviceDataPoint, DeviceDataV2Namespace, DeviceDataV2Point } from "@careevolution/mydatahelps-js";
```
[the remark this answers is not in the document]
greinard: Good idea.   Done.

greinard:
Thank you for the in-depth review @bxh.  I appreciate it.

I think I have addressed and/or responded to all of the feedback thus far.  Please let me know if you would like to see any further revisions.

src/helpers/daily-data-providers/daily-data.ts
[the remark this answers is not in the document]
bxh: Fair enough.

src/helpers/insight-matrix/survey-data-type.ts
bxh: My bad.

src/helpers/daily-data-providers/combined-sleep-v2.ts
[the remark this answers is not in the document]
bxh: `Promise.all` iterates the input array once, With a single element that 's one very cheap property access. The call-stack depth is identical. My point was the second branching doesn't bring any real payoff.

bxh (approved):
LGTM, @greinard. Thanks for taking on this heavy lift and resolving all the outstanding issues.

src/helpers/daily-data-providers/combined-sleep-v2.ts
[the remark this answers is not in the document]
greinard: I'm referring to the call to `combineResultsUsingMaxValue(...)`, which does some looping and constructs a new combined result object.  We can skip that part when only one result will exist, so this short circuit just returns the existing promise.

https://github.com/CareEvolution/MyDataHelpsUI/pull/456/files#diff-8c37ea80434215af36ae2b24f63f5b766963ae77b11eef39f425c62ee1f40338R34

greinard:
Thanks @bxh.

PR #463, opened 2025-04-12
Micro Trend Redesign and additional options

chrisnowak:
## Overview

Supercedes https://github.com/CareEvolution/MyDataHelpsUI/pull/458

This does a bit of a redesign of the "MicroTrend" component and adds a lot of additional options for it.  The current micro trend had a few issues; in particular the bar chart was too dense and the bars were too close together to meaningfully see what was happening

Screenshot 2025-04-12 at 10 45 28 AM

- Font size has been adjusted
- Spark bar chart now has a "rounded" variant which is used for this to give the bars a rounded feel
- Spark bar chart now has a "gap" property which can provide a gap between
- Spark bar chart bars now can have an "opacity" attached to them; this is used to highlight the current day on the spark bar chart for the micro trend
- Chart position can now be top, bottom, or responsive
  - Bottom is the default
  - Responsive means that the the chart will be positioned on the right by default, but will change to bottom if the width of the component goes below 150px
- You can attach a click handler to MicroTrend now
- You can specify an indicator for MicroTrend now (e.g. an info button or chevron)
- Also fixes https://github.com/CareEvolution/MyDataHelpsViewBuilder/issues/248

Couple more detailed notes:
- Some of the behavior depends on the width of the micro trend component; e.g. responsive mode and the number of bars that display on the trend (which also impacts the amount of data queried).  This introduced a couple weird order of operations issues because we weren't rendering the component until we had data, but we couldn't calculate how much data to query until we knew the width of the component, and we can't calculate the width until the component is rendered.
  - The solution here was to render an empty div we could attach a ref to, then calculate the bars/chart position using that empty div
- I considered having a default indicator when there was a click handler specified.  However, since this is meant to be used in a compressed space; I didn't want to always force an indicator as it could make it look pretty crowded if the grid was pretty tight.  

## Security

Standard react code, no unsafe html or anything

## Testing

- Storybook should exercise this pretty well, but I will create a corresponding view builder PR as this component has been overdue for inclusion in view builder.

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/helpers/relative-activity.tsx
```diff
@@ -30,26 +30,25 @@ export function queryRelativeActivity(startDate: Date, endDate: Date, dataTypes:
 
     return Promise.allSettled(promises).then((results) => {
         dataTypes.forEach((dataType, index) => {
-            relativeActivityResults[dataType.dailyDataType] = {};
```
chrisnowak, line 33: subtle change here to this - I think here the correct behavior wasn't to return an empty set of results for a data type if there is an error...but rather we should not add that data type key to the result dictionary at all.  Doesn't matter a ton but the micro trend treats an empty result as having "some data" so it was rendering empty components in some cases that should just be hidden

chrisnowak:
Definitely some tweaks or debouncing needed here; changing screen size seems to cause an infinite query loop under some circumstances (or at least a massive number of queries)

chrisnowak:
Going to also slightly adjust the design to be consistent with a couple additional places where we're doing similar things

src/components/container/RelativeActivityDayCoordinator/RelativeActivityDayCoordinator.tsx
```diff
@@ -47,19 +47,20 @@ export default function RelativeActivityDateRangeCoordinator(props: RelativeActi
         if (!props.dataTypes.length || !availableDataTypes?.length) return null;
 
         let bars: SparkBarChartBar[] = availableDataTypes.map(dataType => {
-            if (!relativeActivityData || !relativeActivityData[dataType.dailyDataType] || !relativeActivityData[dataType.dailyDataType][dayKey]) {
+            const relativePercent = relativeActivityData?.[dataType.dailyDataType]?.[dayKey]?.relativePercent;
```
bxh, line 50:
```suggestion
            const dataForDay = relativeActivityData?.[dataType.dailyDataType]?.[dayKey];
            const relativePercent = dataForDay?.relativePercent ?? 0;
```

src/components/container/RelativeActivityDayCoordinator/RelativeActivityDayCoordinator.tsx
```diff
@@ -47,19 +47,20 @@ export default function RelativeActivityDateRangeCoordinator(props: RelativeActi
...
+            if (relativePercent === undefined) {
                 return { color: 'var(--mdhui-color-primary)', barFillPercent: 0 };
             }
 
-            const value = relativeActivityData[dataType.dailyDataType][dayKey].value || 0;
+            const value = relativeActivityData![dataType.dailyDataType][dayKey].value || 0;
```
bxh, line 55:
```suggestion
            const value = dataForDay?.value ?? 0;
```

src/components/presentational/SparkBarChart/SparkBarChart.tsx
```diff
@@ -8,24 +8,46 @@ export interface SparkBarChartProps {
...
+	variant?: "square" | "rounded";
 }
 
 export interface SparkBarChartBar {
 	color: ColorDefinition;
 	barFillPercent: number;
+	opacity?: number;
 }
 
 export default function SparkBarChart(props: SparkBarChartProps) {
```
bxh, line 21: Minor: save `prop` reference below.
```suggestion
export default function SparkBarChart({averageFillPercent, bars, innerRef, style, gap, variant}: SparkBarChartProps) {
```

bxh (approved):
This is a solid and thoughtful refactor. Look good to me except for a test error to resolve.

src/components/container/MicroTrend/MicroTrend.tsx
```diff
@@ -1,30 +1,87 @@
...
+                if (!chartWidth) { return "bottom"; }
+                if (chartWidth < 150) {
+                    return "bottom";
+                }
+                return "right";
+            }
+
+            let chartWidth = widthReferenceRef.current?.clientWidth;
+            const approximateBarWidth = 24;
+            if (!chartWidth) { return 7; }
```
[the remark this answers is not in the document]
chrisnowak: Ah yes...got mixed up when I refactored/fixed this.  I think we can just have it return actually, no need to default it to 7 if there's no ref yet

src/components/container/MicroTrend/MicroTrend.tsx
```diff
@@ -1,30 +1,87 @@
...
 }
 
 export default function MicroTrend(props: MicroTrendProps) {
     const context = useContext(LayoutContext);
     const [results, setResults] = useState<{ [key: string]: RelativeActivityQueryResult } | undefined>(undefined);
     const dateRangeContext = useContext(DateRangeContext);
     const date = props.date ?? dateRangeContext?.intervalStart ?? startOfDay(new Date());
+    const widthReferenceRef = React.useRef<HTMLDivElement>(null);
+    const [barsToDisplay, setBarsToDisplay] = useState<number | null>(null);
+    const [chartPosition, setChartPosition] = useState<"right" | "bottom">("bottom");
```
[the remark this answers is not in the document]
chrisnowak: Makes sense for when "hideIfNoRecentData" is false, I'd only been thinking about that being set to true as it is for my use case

PR #464, opened 2025-04-15
Update the ViewEhr component so that it renders for Apple Health Records and Health Connect PHR as well.

greinard:
## Overview

This branch updates the `ViewEhr` component such that it also renders if data is available from Apple Health Records or Health Connect PHR.  It leverages the new data availability API to do so.

## Security

No new security risk.  Just adding additional conditions under which the `ViewEhr` component may render.

## Testing

I have updated the storybook entries, but it is worth testing this against live data to ensure the API queries and the handling of their responses functions as expected.  This has been tested using a snapshot version of MDH-UI and a local VB deployment pointed at QA.

### Documentation

### DIFF
--- a/src/components/container/ViewEhr/ViewEhr.tsx
+++ b/src/components/container/ViewEhr/ViewEhr.tsx
@@ -7,9 +7,9 @@ import { faRefresh } from '@fortawesome/free-solid-svg-icons'
 import { ColorDefinition, useInitializeView } from '../../../helpers'
 import './ViewEhr.css'
 import { ButtonVariant } from '../../presentational/Button/Button'
-import { noop } from "../../../helpers/functions";
+import { noop } from '../../../helpers/functions';
 
-export type ViewEhrPreviewState = 'no ehr connections' | 'some ehr connections' | 'fetching data';
+export type ViewEhrPreviewState = 'notConnected' | 'fetchComplete' | 'fetchingData';
 
 export interface ViewEhrProps {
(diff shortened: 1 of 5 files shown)


src/components/container/ViewEhr/ViewEhr.tsx
```diff
@@ -1,63 +1,69 @@
...
+import { ColorDefinition, useInitializeView } from '../../../helpers'
+import './ViewEhr.css'
 import { ButtonVariant } from '../../presentational/Button/Button'
+import { noop } from "../../../helpers/functions";
+
+export type ViewEhrPreviewState = 'no ehr connections' | 'some ehr connections' | 'fetching data';
 
 export interface ViewEhrProps {
-    onClick(): void;
+    previewState?: ViewEhrPreviewState;
```
chrisnowak, line 15: I guess this is vaguely a breaking change for the "fetchingData" previewState if someone was supplying that, but I think that would just be used by storybook and not view builder, so seems fine.  I think view builder supplying "fetchComplete" will result in TS complaining and possibly compilation error but wouldn't actually cause a bug or anything

chrisnowak (approved):

src/components/container/ViewEhr/ViewEhr.tsx
greinard: I agree, probably not a big deal, but I went ahead and reverted the values to avoid the breaking change.

PR #467, opened 2025-04-28
New Display Logic For Providers With Existing Connections (Direct, Through Same Endpoint, and Through Same Managing Organization)

boris-careev:
## Overview

This PR is related to the MyDataHelpsJS PR https://github.com/CareEvolution/MyDataHelps.js/pull/159.

Now the `ExternalAccountProvider` interface contains additional properties populated by the External Accounts API (see https://github.com/CareEvolution/Consumers/pull/13061): `enabled`, `message`, `relatedProvider`, `managingOrganization`. These properties are used in this PR to drive the UI behavior.

`ExternalAccountProvider` objects with `enabled=false` are displayed but can't be selected, and the `relatedProvider` + optionally `managingOrganization` are used with a localized template string to indicate why the provider cannot be selected. Here is what the UI looks like with non-selectable providers due to the user having an account with a provider with 1) the same endpoint and 2) in the same managing organization. 

_Tested locally using local deployments of MyDataHelpsUI (this PR + storybook), MyDataHelpsJS (https://github.com/CareEvolution/MyDataHelps.js/pull/159) and External Account API in Consumers (https://github.com/CareEvolution/Consumers/pull/13061 + participant access token)_

FYI @wmmoye @eevilninja @harithamenta @jillbee 

image

In this case:
 - The user has a connection to **Austin Regional Clinic** (usual behavior where we display "Connected" and the provider can no longer be selected) 
 - The **Carle Foundation Hospital** provider was imported in FHIR admin, and the managing organization of its endpoint was populated with the same code system + value as that of **Austin Regional Clinic** => meaning that both providers belong to the same managing organization, namely "HCA".
 - The **Bellin Health** provider was imported in FHIR admin, and linked to the same endpoint as **Austin Regional Clinic** => meaning that both providers use the exact same FHIR endpoint.

## Security

  - [x]There are no temporary testing changes committed such as API base URLs, access tokens, print/log statements, etc.

## Testing

- [x]  This change requires additional testing in the MDH iOS/Android/Web apps. (Create a pre-release tag/build and test in a ViewBuilder PR.)

Testing steps:

1) Make sure that a provider you have an account for exists in FHIR Admin, for instance: **Austin Regional Clinic**. 

2) In FHIR Admin, edit this provider's Endpoint to set the managing organization fields. The steps are: Providers tab => Click on the provider (in this case **Austin Regional Clinic**) => Associated FHIR Endpoint section => Edit Endpoint:
image
=> then set the fields as follow:
image
```
Managing Org. Identifier System
   https://fhir.epic.com/Developer/Management/OrganizationId

Managing Org. Identifier Value
   1882

Managing Org. Identifier Display
   HCA
```
3) In MDH, connect to that provider, so in this particular example, you would connect to **Austin Regional Clinic**. 

4) In FHIR Admin modify ANY provider, for example **Bellin Health** so it uses the same Endpoint as the provider you connected to (**Austin Regional Clinic** in this case). The steps are: Providers tab => select Bellin Health to access the edit page => Associated FHIR Endpoint section, delete the existing association if needed and find the **Austin Regional Clinic** endpoint:
image
image
This is what you should see after selecting the endpoint:
image
Make sure to click "Save" at the bottom of the page:
image

3) Now edit ANY OTHER provider, for example **Carle Foundation Hospital** => Associated FHIR Endpoint section => Edit Endpoint and set the managing organization fields to be the same as the provider you connected to (in this case **Austin Regional Clinic**)
image
```
Managing Org. Identifier System
   https://fhir.epic.com/Developer/Management/OrganizationId

Managing Org. Identifier Value
   1882

Managing Org. Identifier Display
   HCA
```

5) Now go to MyDataHelps => Provider Search and you should see the following messages under **Austin Regional Clinic**, **Bellin Health** and **Carle Foundation Hospital**. Also, these provider should not be selectable.
 
image

### Documentation

@chrisnowak, @lynnfaraday does this need a story book or doc update?

Also, what is the process to validate language + translation?

Side note: the current build error should go away once the MyDataHelps JS PR is merged.


lynnfaraday (changes requested):
IMHO we should have some test stories in the ProviderSearch.Stories to account for these display variations.

Testing in different languages should be part of the regular QA Plan, to ensure that all the prompts are in all the languages (which they currently aren't). Nathia or Basyl can help get the actual translations.

tanvdo (changes requested):
Once I linked this locally to the main branch of MyDataHelpsJS (unreleased), I had to make trivial modifications to the `ExternalAccountConnectionAlert.tsx` file in order get this to build.

package-lock.json
```diff
@@ -3558,14 +3557,14 @@
       }
     },
     "node_modules/@rollup/pluginutils": {
-      "version": "5.1.0",
-      "resolved": "https://registry.npmjs.org/@rollup/pluginutils/-/pluginutils-5.1.0.tgz",
-      "integrity": "sha512-XTIWOPPcpvyKI6L1NHo0lFlCyznUEyPmPY1mc3KpPVDYulHSTvyeLNVW00QTLIAFNhR3kYnJTQHeGqU4M3n09g==",
+      "version": "5.1.4",
```
greinard, line 3560: Was this update intentional? 

If not, we can reset the package-lock back to the one from main and re-run `npm install`.

src/components/container/ProviderSearch/ProviderSearch.tsx
```diff
@@ -239,6 +239,26 @@ export default function ProviderSearch(props: ProviderSearchProps) {
                             {linkedExternalAccounts[provider.id] && linkedExternalAccounts[provider.id].status != 'unauthorized' &&
                                 <div className="provider-status connected-status">{language("connected")}</div>
                             }
+                            {!linkedExternalAccounts[provider.id] && provider.enabled === false &&
```
greinard, line 242: Nitpick: You can simplify the second condition to `!provider.enabled`.

src/components/container/ProviderSearch/ProviderSearch.css
```diff
@@ -75,4 +75,5 @@
 
 .mdhui-provider-search .search-results .provider .provider-status.connected-status {
     color: var(--mdhui-color-success);
+    cursor: default;
```
greinard, line 78: I think you can remove this.  If we want to do something about the cursor when the provider is not enabled, we should probably adjust the style class for the entire provider `UnstyledButton` rather than just the status.

greinard: I still think we should handle this cursor adjustment at the provider level instead of just the status.

I suggest that we:
- Remove this CSS cursor adjustment.
- Add a new CSS rule:
```
.mdhui-provider-search .search-results .provider.disabled {
    cursor: default;
}
```
- Update the provider `UnstyledButton` so that class name is:
```
className={provider.enabled ? 'provider' : 'provider disabled'}
```

That should make it so the cursor is the default arrow over the entire disabled provider row.

greinard (changes requested):
Looks pretty good to me.  Just one more requested adjustment.

greinard (approved):

locales/de.json
```diff
@@ -442,6 +442,8 @@
     "procedure": "Verfahren",
     "procedures": "Verfahren",
     "procedures-title": "Verfahren",
+    "provider-disabled-reason-with-managing-organization": "Sie haben bereits eine Verbindung zu {provider} über {relatedProvider} ({managingOrganization}).",
```
lynnfaraday, line 445: I think  {{managingOrganization}} with two curlies is wrong - shouldn't it be like {provider}?

lynnfaraday (approved):

locales/de.json
boris-careev: oh, my eyes tricked me too for a second... but if you look closely, these are not double curlies but parenthesis + curly

tanvdo (approved):

PR #469, opened 2025-04-30
Fixed button placement was blocking content

ajmenca:
## Overview

- Button was moved to a fixed position in #100, to work properly with an infinite scroll so that the button wasn't pushed off the screen constantly by the new data.
- This broke the layout of the YouTube step, content would end up behind the button (the story was obviously broken and it wasn't noticed in review of #100):
image

- Also technically didn't work correctly in all cases anyway; if the list stopped loading, the last rows would be behind the button.
image

- Revert button to inline-block by default, customize only for ConnectEhrStep; fix ConnectEhrStep by adding margin-bottom so next button doesn't obscure content.
image
image

## Security

N/A; styling changes only.

## Testing

- Test that the YouTubeStep stories look correct.
- Test that the ConnectEhrStep stories look correct.
  - Verify using the Live story that infinite scroll/load works properly.
  - Verify using the Live story with a specific search that is more than one page, that scrolling to the end of the search allows you to see the last results.

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/step/ConnectEhrStep/ConnectEhrStep.css
+++ b/src/components/step/ConnectEhrStep/ConnectEhrStep.css
@@ -1,0 +1,9 @@
+.mdhui-step-container.mdhui-connect-ehr-step .mdhui-provider-search {
+    margin-bottom: 94px;
+}
+.mdhui-step-container.mdhui-connect-ehr-step .mdhui-step-next-button {
+    position: fixed;
+    bottom: 0;
+    left: 0;
+    right: 0;
+}
--- a/src/components/step/ConnectEhrStep/ConnectEhrStep.tsx
+++ b/src/components/step/ConnectEhrStep/ConnectEhrStep.tsx
@@ -7,6 +7,8 @@ import StepTitle from "../StepTitle";
 import { ExternalAccountProvider } from "@careevolution/mydatahelps-js";
 import { ProviderSearchPreviewState } from "../../container/ProviderSearch/ProviderSearch";
 
+import './ConnectEhrStep.css';
+
 export interface ConnectEhrStepProps {
     title?: string;
     text?: string;
@@ -23,7 +25,7 @@ export interface ConnectEhrStepProps {
  */
 export default function ConnectEhrStep (props: ConnectEhrStepProps) {
     return (
-        <StepLayout>
+        <StepLayout className="mdhui-connect-ehr-step">
             <StepTitle
                 text={props.title}
                 textAlign={props.styles.titleAlignment}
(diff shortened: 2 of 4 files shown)


greinard (approved):
I have tested via the storybook and this update fixes the reported issues.

I do have one suggested update for the `ConnectEhrStep` next button rendering.  I'm not really a fan of having the scrollable content be visible directly behind and against the button.  This suggested update adds a wrapper div with a transparent-to-white linear gradient such that the content going behind the button fades away just before reaching it.  https://github.com/CareEvolution/MyDataHelpsUI/compare/ajm/bug/survey-step-next-button-blocking-content...ConnectEhrSuggestion

image

ajmenca:
> I have tested via the storybook and this update fixes the reported issues.
> 
> I do have one suggested update for the `ConnectEhrStep` next button rendering. I'm not really a fan of having the scrollable content be visible directly behind and against the button. This suggested update adds a wrapper div with a transparent-to-white linear gradient such that the content going behind the button fades away just before reaching it. [ajm/bug/survey-step-next-button-blocking-content...ConnectEhrSuggestion](https://github.com/CareEvolution/MyDataHelpsUI/compare/ajm/bug/survey-step-next-button-blocking-content...ConnectEhrSuggestion)
> 
> image

So, this is one of of those places where I think both designs suffer from flaws.  I agree with you that the content showing up behind the button looks wonky.

However, take this example:
image

On windows, it's clear that you can scroll because there's a scroll bar.  But on Android (and I think iOS) scroll bars don't just *show up* any more.  They only show up once you *start scrolling*, which means that in these sorts of [perfectly cropped](https://tyler.io/2019/10/perfectly-cropped/) situations, there's no indication to a user that you can scroll.  I find this extremely annoying.  And I'm not alone in this; there's plenty of complaints and discussion around the internet on how to really indicate you can scroll.

Personally, my preference ends up being a drop shadow?  It helps hint that there's a layer involved:
image

The bad part here, is that without significantly re-working the implementation (because of which box is scrolling), it's hard to actually get feedback that the box is scrolled and therefore you can "remove" the dropshadow.  So when you get to the bottom (or don't have enough content to fill the screen), it does look a little weird:
image

ajmenca:
Or, alternatively, a much more subtle one:

image

greinard:
Yeah, I do like how that looks.

In any case, I don't think it is worth holding up this PR.  It was just some food for thought.  Maybe we can address it more globally/generically at some point with a first class fixed button option.  That way we won't have to copy this css tweak to another step type if one ends up needing similar behavior.

PR #500, opened 2025-07-11
Add caching rules for static assets

n8v:
This PR adds caching rules for various static assets to improve performance. See https://github.com/CareEvolution/SecurityAndChangeControl/issues/4041


kneumei (commented):
Same comment as over in ViewBuilder: Want to make sure this won't add friction to dev workflows.

kneumei (approved):

PR #510, opened 2025-08-19
Localize App Download

lynnfaraday:
## Overview

The app store download icons weren't localized

https://ui.mydatahelps.org/?path=/docs/container-appdownload--docs&globals=language:de

image

Now they use the central bade icons based on current locale (which required a new getCurrentLocale method that took into account our supported languages and fallbacks).

I also removed Samoan, which we decided not to support.

## Security

Minor UI update

## Testing

- [X] This change requires additional testing in the MDH iOS/Android/Web apps. (Create a pre-release tag/build and test in a ViewBuilder PR.)

Test the storybook scenarios.

Also test in ViewBuilder and the main app just to make sure there are no CSP concerns there blocking the images appearing.

### Documentation

- [X] This change does not impact documentation or Storybook.

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/container/AppDownload/AppDownload.tsx
```diff
@@ -5,10 +5,9 @@ import '@fortawesome/fontawesome-svg-core/styles.css';
 import PlatformSpecificContent from '../PlatformSpecificContent/PlatformSpecificContent';
 import { useInitializeView } from '../../../helpers/Initialization';
 import TextBlock from '../../presentational/TextBlock';
-import googlePlayDownload from '../../../assets/google-play-download.svg';
-import appStoreDownload from '../../../assets/app-store-download.svg';
 import language from "../../../helpers/language";
 import { CardTitle } from '../../presentational';
+import { getCurrentLocale } from '../../../helpers/language';
```
greinard, line 10: Minor:  This can be combined with the language import above.
```
import language, { getCurrentLocale } from '../../../helpers/language';
```

src/helpers/language.ts
```diff
@@ -78,6 +84,10 @@ export function getCountryCodeFromIso(locale: string) : string | undefined {
     return normalizedLocale.split("-")[1];
 }
 
+function isLocaleSupported(locale: string) {
+	return !!localeToStringsMap[normalizeLocaleString(locale)];
```
greinard, line 88: Minor: The `normalizeLocaleString(...)` call is not really an expensive operation, but given the current usage of this internal function, I think the `locale` will always be normalized already.

lynnfaraday: Yeah, I figured it was safer this way because we can't anticipate how it might be used in the future. If it were more expensive, we could just add a note, but it seemed pretty harmless to just put in the extra safeguard.

PR #526, opened 2025-09-27
Provider Successor Connection 

boris-careev:
## Overview

Early PR to test prototype feature nudging users towards creating a new connection with a more recent version of the provider they are currently connected to. For instance, to connect to the FHIR R4 version of the DSTU2 provider the user currently has a connection to. In the prototype I simply display a sentence with a link to initiate the new connection (see screenshots in the Testing section). This more recent version of the provider is referred to as "successor" below.

Improvements coming up:

- Verbiage (open to suggestions)
- Add missing languages
- Conditional display of the connection nudge message based on already connected providers

Side note: another way to do this could be to just replace the functionality of the Reconnect button with the same logic as the link I'm using in the prototype, to initiate the new connection. The problem with this is that we are not really reconnecting to the provider but creating a new connection, which, if the new connection doesn't yield as much data, could confuse users. Also, it would be up to us to delete the data from the previous provider, potentially deleting data and only have partial data from the new provider. And if we are not deleting data, then the user would see the old provider + the new provider in their Linked Accounts, creating more confusion as they thought they were just reconnecting the one account they had.

FYI @wmmoye, @chrisnowak 

## Security

## Testing

- Connect to a provider for which a successorID exist
- Go to the list of ExternamAccounts
- Make sure you see the provider you are connected to along with the following message + link to connect to the successor provider:

Example view 1:

image

image

Example view 2:

image

image

- Click on the "here" link and make sure you see the provider login page for the successor provider is displayed. 

- Ideally connect to the new provider if you have production credentials

- After logging in and going through consent on the provider side, the user should be redirected to the list of accounts

### Documentation

- [x] I have added relevant Storybook updates to this PR. (part of ExternalAccountList)
- [x] If this feature requires a developer doc update, I have tagged `@CareEvolution/api-docs`.


boris-careev:
Replaced with https://github.com/CareEvolution/MyDataHelpsUI/pull/604

PR #604, opened 2026-05-12
Redirect Reconnect Flow To Successor Provider When Configured

boris-careev:
## Overview

When a provider has a `successorID` configured, the existing Reconnect button now silently targets the successor provider instead of the legacy one - no additional UI or user-facing wording is introduced.

Changes:
- **`SingleExternalAccount`**: Only the Reconnect link is shown for accounts whose provider has a `successorID`, regardless of the account's sync status (except while the account is being deleted or refreshed). Previously it only appeared on `unauthorized` accounts.
- **`ExternalAccountList`**: `reconnectAccount` now resolves the provider ID as `successorID ?? id`, so the `connectExternalAccount` call targets the successor when one is configured. The backend is then responsible for detecting the predecessor account and cleaning it up after the successor connection succeeds.

## Security

These changes only alter which provider ID is passed to the existing `connectExternalAccount` API call. No new endpoints, no credentials, no data exposure. The successor ID comes from the provider model already returned by the API.

## Testing

To test, configure a provider with a `successorID` in the backend and verify:
1. A connected account for that provider shows the **Reconnect** button only (no Refresh link), regardless of its sync status.
2. Clicking **Reconnect** initiates the connection flow against the successor provider (confirm via network tab: `externalaccountproviders/{successorID}/connect`).
3. After successful connection, the backend deletes the predecessor account (backend change tracked separately).
4. Providers without a `successorID` retain existing reconnect behavior unchanged.

### Documentation

## Reviewers

- Subject-matter experts
- Style/editing reviewers


src/components/presentational/SingleExternalAccount/SingleExternalAccount.tsx
```diff
@@ -74,14 +74,14 @@ export default function SingleExternalAccount (props: SingleExternalAccountProps
...
-				{(getStatus() === "fetchComplete" || getStatus() === "error") &&
+				{(getStatus() === "fetchComplete" || getStatus() === "error") && !props.externalAccount.provider.successorID &&
 					<p>
 						<a href="javascript:{}" onClick={refresh}>
 							<FontAwesomeSvgIcon icon={faRepeat} /> {language("external-account-refresh")}
 						</a>
 					</p>
 				}
-				{getStatus() === "unauthorized" &&
+				{(getStatus() === "unauthorized" || (!!props.externalAccount.provider.successorID && getStatus() !== "deleting")) &&
```
greinard: Do we want to allow the reconnection while data is being fetched?  If not, this line should probably be:

```
{(getStatus() === "unauthorized" || (!!props.externalAccount.provider.successorID && getStatus() !== "fetchingData" && getStatus() !== "deleting")) &&
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@careevolution/mydatahelps-ui",
-  "version": "3.17.0",
+  "version": "3.17.0-boris-provider-successor-mdhui-v2.0",
```
greinard: Just an FYI that there is a `Snapshot.ps1` script you can run to automatically adjust the version number and tag/build snapshots.  You'd need to revert this updated version number to start using that, but then you just run that script each time you need a new snapshot built and it takes care of the rest.

rollup.config.mjs
```diff
@@ -9,7 +9,7 @@ import terser from "@rollup/plugin-terser";
 import peerDepsExternal from 'rollup-plugin-peer-deps-external';
 import analyze from 'rollup-plugin-analyzer';
 
-const limitBytes = 8.5e6;
+const limitBytes = 8.55e6;
```
greinard: You can revert this now.  It was bumped to `9.0e6` by a different PR.

greinard (approved):

PR #605, opened 2026-05-12
A few minor component updates.

greinard:
## Overview

This branch has a few minor updates to existing components.
- `Card` now supports having a click handler.  This is useful when you want to display a number child elements in a card, but have the entire thing function as a single CTA.  When a click handler is provided, the cursor gets set to `pointer`.
- `Action` now inherits the cursor from its parent rather than setting it to `default`.  This allows a clickable card's cursor to pass through to child actions when that child action does not have a click handler of its own.  If the action has a click handler of its own, its cursor still gets set to `pointer` regardless of what the parent cursor is.  I also cleaned up and modernized the `Action` Storybook a bit.
- `Goal` was updated such that it also inherits the cursor from its parent.  The `recharts-wrapper` class was setting the cursor to `default`, but now it inherits the cursor.  This ensures the cursor is uniform across the goal rather than swapping back and forth to `default` when pieces of the chart are hovered.
- `Goal` was also updated to prevent the outline from showing up on clicked pie chart segments in the progress ring.

## Security

No new security risk.  Just minor rendering and event handler updates.

## Testing

I have updated the Storybook stories to make it fairly simple to verify the above claims.

### Documentation

### DIFF
--- a/src/components/presentational/Action/Action.stories.tsx
+++ b/src/components/presentational/Action/Action.stories.tsx
@@ -1,49 +1,152 @@
-﻿import React from "react";
-import Action, { ActionProps } from "./Action";
-import Layout from "../Layout";
+﻿import React from 'react';
+import Action, { ActionProps } from './Action';
+import Layout from '../Layout';
 import './Action.stories.css';
-import { FontAwesomeSvgIcon } from "react-fontawesome-svg-icon";
-import { faFile } from "@fortawesome/free-solid-svg-icons";
-import { Meta, StoryObj } from "@storybook/react";
+import { FontAwesomeSvgIcon } from 'react-fontawesome-svg-icon';
+import { faArrowRight, faFile, faHeart } from '@fortawesome/free-solid-svg-icons';
+import { Meta, StoryObj } from '@storybook/react';
+import { argTypesToHide } from '../../../../.storybook/helpers';
 
-const meta: Meta<typeof Action> = {
-	title: "Presentational/Action",
-	component: Action,
-	parameters: {
-		layout: 'fullscreen'
-	}
+type ActionStoryArgs = React.ComponentProps<typeof Action> & {
+    colorScheme: 'auto' | 'light' | 'dark';
+    withClass: boolean;
+    withIcon: boolean;
+    withTitleIcon: boolean;
+    withIndicator: boolean;
+    withIndicatorIcon: boolean;
+    withIndicatorValue: boolean;
+    withClickHandler: boolean;
+    parentCursor: 'not set' | 'pointer' | 'crosshair';
 };
 
+const meta: Meta<ActionStoryArgs> = {
+    title: 'Presentational/Action',
+    component: Action,
+    parameters: {
+        layout: 'fullscreen'
+    },
+    render: (storyArgs: ActionStoryArgs) => {
+        const componentArgs: ActionProps = {
+            ...storyArgs,
+            ...(storyArgs.withClass && { className: 'action-story-primary', renderAs: 'div' }),
+            ...(storyArgs.withIcon && { icon: <FontAwesomeSvgIcon icon={faFile} color="#2e6e9e" /> }),
+            ...(storyArgs.withTitleIcon && { titleIcon: <FontAwesomeSvgIcon icon={faHeart} color="#df4747" style={{ marginRight: '4px' }} /> }),
+            ...(storyArgs.withIndicator && { indicator: <FontAwesomeSvgIcon icon={faArrowRight} color="#2e6e9e" /> }),
+            ...(storyArgs.withIndicatorIcon && { indicatorIcon: faArrowRight }),
+            ...(storyArgs.withIndicatorValue && { indicatorValue: '3' }),
+            ...(storyArgs.withClickHandler && { onClick: () => alert('Clicked') })
+        };
+
+        return <Layout colorScheme={storyArgs.colorScheme}>
+            <div style={{ cursor: storyArgs.parentCursor }}>
+                <Action {...componentArgs} />
+                {componentArgs.bottomBorder && <div style={{ padding: '16px', fontSize: '0.75em', color: '#888' }}>
+                    Note: Adding this here because the bottom border seen above only renders when the action is not the last child of its parent.
+                </div>}
+            </div>
+        </Layout>;
+    }
+};
 export default meta;
-type Story = StoryObj<typeof Action>;
 
-const render = (args: ActionProps) =>
-	<Layout colorScheme="auto">
-		<Action {...args} />
-	</Layout>;
-
-const baseArgs : ActionProps = {
-	title: "Baseline Survey",
-	subtitle: "Tap here to start your baseline survey",
-	onClick: () => alert("Clicked")
+export const Default: StoryObj<ActionStoryArgs> = {
+    args: {
+        colorScheme: 'auto',
+        title: 'Baseline Survey',
+        subtitle: 'Tap here to start your baseline survey',
+        withClass: false,
+        withIcon: false,
+        withTitleIcon: false,
+        withIndicator: false,
+        withIndicatorIcon: false,
+        withIndicatorValue: false,
+        indicatorPosition: 'default',
+        bottomBorder: false,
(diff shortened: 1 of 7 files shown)


src/components/presentational/Card/Card.stories.tsx
```diff
@@ -43,4 +44,18 @@ export const HighlightVariant = {
...
+export const ClickableCard = {
+	args: {
+		children: <>
+			<Action title="See more details" bottomBorder />
+			<div style={{ display: 'flex', alignItems: 'center', justifyContent: 'center' }}>
+				<Goal variant="compact" label="Goal 1" targetValue={5} maxValue={10} valueProvider={{ type: 'static integer', getValue: async () => 7 }} />
+				<Goal variant="compact" label="Goal 2" targetValue={5} maxValue={10} valueProvider={{ type: 'static integer', getValue: async () => 3 }} />
+			</div>
+		</>,
+		onClick: () => console.log("Card Clicked!")
```
lynnfaraday, line 73: "See more details" made me expect to see more details, not a console statement. I think it would be better if the click event actually did something visible (even if it was as simple as changing a color or something) with a prompt to match.

src/components/presentational/Card/Card.tsx
```diff
@@ -36,7 +37,7 @@ export default function (props: CardProps) {
 	let backgroundColor = resolveColor(layoutContext?.colorScheme, props.backgroundColor);
 
 	return (
-		<div style={{...props.style, backgroundColor:backgroundColor}} ref={props.innerRef} className={classes.join(" ")}>
+		<div style={{ ...props.style, backgroundColor: backgroundColor, ...(props.onClick && { cursor: "pointer" }) }} ref={props.innerRef} className={classes.join(" ")} onClick={props.onClick}>
```
lynnfaraday, line 40: In Chrome and Safari on Mac, I don't see any visible difference between the pointer on a clickable card vs a non-clickable card.

src/components/presentational/Action/Action.stories.tsx
```diff
@@ -1,49 +1,152 @@
...
-	args: {...baseArgs, titleIcon: <FontAwesomeSvgIcon icon={faFile} color="#2e6e9e" />},
-	render: render
-}
\ No newline at end of file
+export const Default: StoryObj<ActionStoryArgs> = {
+    args: {
+        colorScheme: 'auto',
+        title: 'Baseline Survey',
+        subtitle: 'Tap here to start your baseline survey',
+        withClass: false,
```
lynnfaraday, line 57: Maybe I'm missing something, but the story args here seem a little inconsistent compared to other stories. I'm used to seeing the actual _component_ properties reflected in the story arguments. For example, on `BasicPointsForBadges` you can control the `BasicPointsForBadgesProps` like pointsPerBadge and progressBarColor.  Here the story args seem to be activating specific scenarios, but I thought that was more normally done with separate stories?

src/components/presentational/Card/Card.stories.tsx
greinard: I think we pretty commonly use console logs for click handlers in the stories across the repo.  If you search for the text "console.log" you'll get a lot of hits in the stories.

We could add support for changing a color or something like that, but that means having a custom render function that manages additional state of its own, essentially creating another new test component within the story.  The idea here was just to see that the click handler gets called, regardless of what might happen as a result.

The "See more details" text is just arbitrary text for rendering.  We could modify it or maybe add a subtitle to the action that mentions checking for a console log upon click if you think that would be more user friendly.

src/components/presentational/Card/Card.tsx
greinard: I think the issue here was that all the previously existing `Card` stories had a clickable `Action` on them.  You were seeing the pointer from the clickable `Actions` on those existing `Cards` and from the `Card` itself on the new one.

I've added an additional story with a non-clickable `Card` so you can see the difference.  I've also updated the both of the new stories to have different title text and each includes a subtitle with instructions about what to verify when each has been clicked.

src/components/presentational/Action/Action.stories.tsx
greinard: It can be done either way.  I personally prefer to have a single story with all the configurable options because we rarely have individual stories to cover all the permutations.  With a single story and the ability to adjust multiple properties, we can exercise the component with any permutation to test things out.  That said, there are definitely times when I create a separate individual story to illustrate a specific scenario.  I usually hide almost all the controls so it can only show that one scenario, but the per-situation stories definitely have their place as well.

The abstracted settings, I think, also make it easier to configure the more advanced scenarios.  For example, the `indicator` property on `Action`.  I'm not sure the Storybook controls even support configuring that with a `<FontAwesomeSvgIcon icon={faArrowRight} color="#2e6e9e" />` React element.  Even if you can, you'd have to make sure you typed it all in there by hand perfectly.  Having a flag for "with indicator" makes it much easier for someone to see what that means with a real world example without needing to know the implementation details.

The same is true for many of the more complex model object properties too.  Sometimes, having a flag for a scenario is easier and less error prone to configure than trying to build the appropriate JSON config in a Storybook control.

Having the abstracted controls does limit a users ability to fully customize everything about the widget using the raw props, but I think having more user friendly ways to see how a component generally works is more likely what the Storybook should be about.  Devs who need to implement something specific with the component will look under the hood anyways to see how things are done and adjust as needed for their situation.

In other words, I don't think the Storybook controls should function as a spec document for devs.  I think it should function as a showcase of the available configuration options the component has, and should be usable by non-devs as well.  I think the abstracted settings can help with that.

lynnfaraday: > I don't think the Storybook controls should function as a spec document for devs

My only concern is that we don't have any other spec document for devs. I'm not a huge fan of leaving it to folks to sift through code to figure out how something works. Most of the other storybooks I've seen do a good job of surfacing both the available properties (via the control panel) and the use cases (via the specific stories).

Just my 2 cents. The way you have it is also valid, just a different philosophy.

lynnfaraday (approved):

src/components/presentational/Action/Action.stories.tsx
greinard: Yeah.  I guess I'm just not fully convinced of the value of seeing all the props as controls in a Storybook for a dev.  Perhaps if each prop was listed with a full description of what it does and how to use it, that would be useful, but I don't know that it requires the Storybook for that (and may not be the best form factor).  Maybe both can co-exist, though?

As a specific example, I don't think having an editable `className` control in the Storybook does much for a dev.  They wouldn't be able to create a style class to reference, so they'd only be able to use that prop if they knew of an existing style class that would be applicable to that exact component and the way the class name gets applied to it.  The only existing style class that does that is probably the default class that has already been applied to that component.

And then we have administrative type props like `innerRef` which are mostly irrelevant to general usage.  We use it for a specific purpose downstream, but I don't think it would ever need to be set (or do anything if it was) within the Storybook.  It would still be nice to know what it does via some documentation for devs, but again, maybe that doesn't have to be within Storybook.

We also have some props that are or can be set to React components, which I don't think is possible via existing Storybook controls.

Self-documenting code, especially in an open source library where devs have full access to see the inner workings can be pretty useful too.  That kind of documentation never gets out of date and devs can see exactly how things like the `className` and `innerRef` props are to be applied.

As you said, though, I don't think there is a right or wrong answer here, just philosophy and tooling.  🙂

greinard:
Thanks @lynnfaraday.

src/components/presentational/Action/Action.stories.tsx
lynnfaraday: Oh, I agree that it's not necessary for the storybook props to be a direct 1:1 map of the component props. Things like className, innerRef, or a click handler aren't really something you can meaningfully fiddle with via storybook.  Nor is something backed by data like ConditionsList going to be easily editable.  

But I do find it useful to be able to glance at and tweak properties like in DailyDataChart. It's an easier way to digest what it does than looking at the code. So my personal preference would be to do that where practical, with stories to highlight pre-built configurations.

image

(Self-documenting code is also important, in a parallel way.)

PR #613, opened 2026-05-20
Migrate some Apple Health daily data providers to V2.

greinard:
## Overview

This branch migrates a number of Apple Health daily data providers to use the V2 API.  I also added a few new ones.  To the extent possible, I used the V2 aggregate API.

New Providers:
- `Apple Health Blood Glucose` (V2 type: `Blood Glucose`, aggregate: `avg`)
- `Apple Health Min Blood Glucose` (V2 type: `Blood Glucose`, aggregate: `min`)
- `Apple Health Max Blood Glucose` (V2 type: `Blood Glucose`, aggregate: `max`)
- `Apple Health Min Heart Rate` (V2 type: `Hourly Minimum Heart Rate`, aggregate: `min`)
- `Health Connect Blood Glucose` (V2 type: `blood-glucose`, aggregate: `avg`)
- `Health Connect Min Blood Glucose` (V2 type: `blood-glucose`, aggregate: `min`)
- `Health Connect Max Blood Glucose` (V2 type: `blood-glucose`, aggregate: `max`)
- `Combined Blood Glucose` - Uses the `Apple Health Blood Glucose` and `Health Connect Blood Glucose` data providers, prioritizing the value from Apple Health when both are present for a given day.

Updated Providers:
- `Apple Health Max Heart Rate` (V2 type: `Hourly Maximum Heart Rate`, aggregate: `max`)
- `Apple Health Heart Rate Range` - Uses values from the V2 min and max heart rate providers to compute the range for each day.
- `Apple Health Mindful Minutes` (V2 type: `Mindful Sessions`, filter: `not SilverCloud CBT`)
- `Apple Health Resting Heart Rate` (V2 type: `Resting Heart Rate`, aggregate: `avg`)
- `Apple Health Sleep Time` (V2 type: `Sleep Analysis`, values: `['AsleepCore', 'AsleepREM', 'AsleepDeep', 'Asleep']`)
- `Apple Health Core Sleep Time` (V2 type: `Sleep Analysis`, value: `AsleepCore`)
- `Apple Health REM Sleep Time` (V2 type: `Sleep Analysis`, value: `AsleepREM`)
- `Apple Health Deep Sleep Time` (V2 type: `Sleep Analysis`, value: `AsleepDeep`)
- `Apple Health In Bed Time` (V2 type: `Sleep Analysis`, value: `InBed`)
- `Apple Health Steps` (V2 type: `Hourly Steps`, aggregate: `sum`)
- `Apple Health Steps While Wearing Device` (V2 type: `Hourly Steps`, aggregate: `sum`)
- `Combined Mindful Minutes` - Updated the criteria for using the V2 `Apple Health Mindful Minutes` provider.
- `Combined Resting Heart Rate` - Updated the criteria for using the V2 `Apple Health Resting Heart Rate` provider.
- `Combined Sleep Time` - Updated the criteria for using the V2 `Apple Health Sleep Time` provider.
- `Combined Steps` - Updated the criteria for using the V2 `Apple Health Steps` provider.

## Security

No new security risk.  Just migrating to use the delegated V2 device data API, which is already accessible to participants.

## Testing

This change requires a VB PR to fully test.
- Create a test view with a daily data chart for each of the affected daily data providers.
- Configure the view as a tab in QA and visit it with a test participant.
- Share device data with MDH as needed to verify the charts continue to render that data correctly.

### Documentation


locales/de.json
```diff
@@ -332,6 +332,9 @@
     "average-stress-level": "Durchschnittliche Stressstufe",
     "awake-time": "Wachzeit",
     "back": "Zurück",
+    "blood-glucose": "Blutzucker",
```
lynnfaraday, line 335: This change needs to go through the localization hub. Otherwise they're going to get out of sync and cause merge conflicts later. Happy to do that if you like, or you can try yourself. Instructions on slab.

__tests__/helpers/daily-data-providers/daily-data/daily-data-aggregate.test.ts
```diff
@@ -0,0 +1,82 @@
+import { describe, expect, it } from '@jest/globals';
+import { sampleEndDate, sampleStartDate } from '../../../fixtures/daily-data-providers';
```
lynnfaraday, line 2: I like the `setupXYX` functions in the fixture, but having the actual sample data separated from the tests makes them pretty hard to follow. And since most of them are just empty dates or really simple object hashes like `{ "SomeDate": 100 }`, I'm not sure it's worth the added obfuscation. Not a huge deal, just an observation.

src/helpers/daily-data-providers/apple-health-heart-rate-range.ts
```diff
@@ -1,20 +1,15 @@
...
-        return typeFilteredDataPoints.length > 0 ? aggregateFn(...typeFilteredDataPoints.map(getFloatValue)) : 0;
-    };
-
-    return Object.entries(dailyData).reduce((result, [dayKey, dataPoints]) => {
-        const maxHeartRate = getHeartRate(dataPoints, "HourlyMaximumHeartRate", Math.max);
-        const minHeartRate = getHeartRate(dataPoints, "HourlyMinimumHeartRate", Math.min);
-        if (maxHeartRate && minHeartRate) {
-            result[dayKey] = maxHeartRate - minHeartRate;
+    return Object.entries(minResult).reduce((result, [dayKey, minHeartRate]) => {
+        if (maxResult[dayKey] > minHeartRate) {
```
lynnfaraday, line 11: This works, but would it be better to have an explicit check that maxResult actually has an entry for that day?

src/helpers/daily-data-providers/apple-health-sleep-v2.ts
```diff
@@ -1,18 +0,0 @@
-﻿import { querySleepMinutesV2 } from './common-sleep-v2';
-import { DailyDataQueryResult } from '../query-daily-data';
-
-export function totalSleepMinutes(startDate: Date, endDate: Date): Promise<DailyDataQueryResult> {
```
lynnfaraday, line 4: This is going to be a major version bump, right? Since we're fundamentally changing the way things like inBedTime work and effectively deprecating the V1 queries...

src/helpers/daily-data-providers/health-connect-blood-glucose.ts
```diff
@@ -0,0 +1,8 @@
+﻿import { DailyDataQueryResult } from '../query-daily-data';
+import { queryAggregateDailyData } from './daily-data/daily-data-aggregate';
+
+const MMOL_TO_MGDL_SCALE_FACTOR = 18;
```
lynnfaraday, line 6: A comment for context would be nice on this scale constant.

Ditto in the other two files.

__tests__/helpers/daily-data-providers/apple-health-resting-heart-rate.test.ts
```diff
@@ -1,11 +1,10 @@
...
-import { sampleDailyData, sampleEndDate, sampleResult, sampleStartDate, setupAverageValueResult, setupDailyData, startDateFunctionEvaluator } from '../../fixtures/daily-data-providers';
+import { sampleEndDate, sampleResult, sampleStartDate, setupAggregateDailyData } from '../../fixtures/daily-data-providers';
 import appleHealthRestingHeartRate from '../../../src/helpers/daily-data-providers/apple-health-resting-heart-rate';
 
 describe('Daily Data Provider - Apple Health Resting Heart Rate', () => {
-    it('Should query for daily data and build an average value result keyed by start date.', async () => {
-        setupDailyData('AppleHealth', 'RestingHeartRate', sampleStartDate, sampleEndDate, startDateFunctionEvaluator, sampleDailyData);
-        setupAverageValueResult(sampleDailyData, sampleResult);
+    it('Should query for aggregate daily data and return the average values keyed by date.', async () => {
+        setupAggregateDailyData('AppleHealth', 'Resting Heart Rate', sampleStartDate, sampleEndDate, 'avg', sampleResult);
```
lynnfaraday, line 7: Throughout this are data type names like "Resting Heart Rate".  We have what seems to be a mapping for this in the tables like `RESTING_HEART_RATE_SOURCES`. It seems like it would be more clear if we used that, so instead of a literal string we have something like `RESTING_HEART_RATE_SOURCES[AppleHealth]`.  

This would apply both to the tests and to the providers themselves. So something like: 

`return queryAggregateDailyData('AppleHealth', 'Hourly Steps', startDate, endDate, 'sum');`

would become:

`return queryAggregateDailyData('AppleHealth', STEPS_SOURCES[AppleHealth], startDate, endDate, 'sum');`

lynnfaraday (changes requested):
Sorry it took so long.
Just a few little things and suggestions.

I'm not sure how to check if all of the device type names are correct, so it would be good to have someone more familiar with device data (Vlad maybe?) glance through as well.

__tests__/helpers/daily-data-providers/daily-data/daily-data-aggregate.test.ts
greinard: Yeah, I can see that.

I think the shared constants are useful, though, because we often don't need to know the specific values.  We just need to know there is a date, some daily data, or some data points, and that they get used in calls, returned from calls, and passed along to subsequent calls.  If we define them directly in each test, we also need to update all those instantiations if we want to adjust something about the typical testing values or if a new required field gets added to the shape of the type of some object.

For tests which care about the specific values, they should not use the shared constants.

The slight exception to that is for tests related to start/end date filtering.  For those, you may need to peek at the fact that the start date is `today - 6` and the end date is `today` (typical for our daily data charts) so you can place test dates within and outside that date range.  However, it saves a lot of boilerplate date instantiation by just reusing those, and the values are very discoverable.

src/helpers/daily-data-providers/apple-health-heart-rate-range.ts
greinard: I think this should be okay.  If there is no entry for that day, the comparison ends up being `undefined > minHeartRate`, which gets treated as `NaN > minHeartRate`, which is always `false`.  So for this condition to return `true`, `maxResult[dayKey]` has to be defined and has to be greater than `minHeartRate`.

src/helpers/daily-data-providers/apple-health-sleep-v2.ts
greinard: Possibly.  I'm not sure it necessarily has to be, but I am also not really opposed to that either.  Usually we determine that when it is time to cut the next release and we take a look to see if anything is a breaking change.

In this case, I don't think there will be any breaking changes as far as the code is concerned.  The data the updated providers return should be the same as they were before.  It'll just be coming from a different endpoint.  The V2 data types for these providers are all part of the "system" MyDataHelps project, so they are already available by default for all projects.  The V1 values they return get persisted to V2 as well, which is where we will source them from now on.

There is one non-code related difference on the back end, though, which is that the V2 data has a 90 day retention policy (currently), meaning data older than 90 days will not be available by default.  This should cover the vast majority of use cases we have for displaying device data, and the retention days can be lengthened via configuration for individual projects, when necessary.  Again, that is an API detail, though.

As we sunset V1 in favor of V2, I suspect we'll be purging the historic V1 data a bit at a time, so it wouldn't really benefit a user of this library to stay on an older version to continue using V1 data queries, at least not for very long.  If a given project needs more than 90 days worth of history, they should just request that so we can adjust the V2 retention period for that project.  The V2 API is generally more performant as well, so that is good for participants.

So I guess it comes down to a code contract vs behavior contract thing when it comes to SemVer on this one.  The code API itself is not changing, and I'm not sure we ever guaranteed any particular amount of historic data would be available to the daily data providers, but this will be a behavior change that could be unexpected.

I don't think we need to be shy about using major version bumps, so if we think that's a good way to at least draw attention to this change, even if nothing needs to be done as a result, then it is probably worth it.

src/helpers/daily-data-providers/health-connect-blood-glucose.ts
greinard: Updated all three.

__tests__/helpers/daily-data-providers/apple-health-resting-heart-rate.test.ts
greinard: There is some duplication of the type names across a few places, but here are my thoughts on a possible refactor:
- I would prefer that each individual provider defines it own set of availability check source(s) and that these definition declaration files just pull from those.  In other words, instead of the combined definitions file having a `RESTING_HEART_RATE_SOURCES` map, it would just get the sources from the provider(s).  Something like this so the provider itself, which knows best what types of data it needs and can return, would be responsible for defining those things:

```
    {
        dataSource: "Unified",
        type: DailyDataType.RestingHeartRate,
        dataProvider: combinedRestingHeartRateDataProvider,
        availabilityCheck: combinedAvailabilityCheck(combinedRestingHeartRateDataProvider.availabilitySources),
        labelKey: "resting-heart-rate",
        icon: <FontAwesomeSvgIcon icon={faHeartbeat} />,
        formatter: heartRateFormatter,
        previewDataRange: [40, 100]
    },
```

- Regardless of where the type codes end up being defined, I would still suggest that we use raw literal values in the unit tests.  Otherwise, we just test that the constant value was used, but not that the constant value was correct.

- This may be a worthwhile endeavor, but I'd prefer to defer that to a separate PR.  I'll create an issue on the backlog for it.

greinard: Issue for the possible refactor: https://github.com/CareEvolution/MyDataHelpsUI/issues/633

locales/de.json
greinard: PR submitted for the localization hub.

greinard:
Thanks for the review @lynnfaraday.  I think I have addressed or responded to each of your comments.

Please let me know if you have any further concerns.

src/helpers/daily-data-providers/apple-health-sleep-v2.ts
lynnfaraday: My concern is that some of the projects using this might not have V2 enabled at all. So if they upgrade, they  could go from having views that get data to having none without any sort of alert.

If all the data was fully ported over to V2 and we had V2 enabled for everybody by default, that'd be different, but my understanding is that's not currently the case.

__tests__/helpers/daily-data-providers/apple-health-resting-heart-rate.test.ts
lynnfaraday: Sounds good by me. And I agree that provider-specific data types are more useful than a combined definition.

src/helpers/daily-data-providers/apple-health-sleep-v2.ts
greinard: For these data types, it is the case.  The base MDH project (which all projects inherit from) has all of these V2 data types enabled, and they have been enabled for longer than the retention period (currently 90 days), so all current projects in the system have V2 enabled for these data types and all new projects will too, by default.

The only difference will be if any existing usage expects to be able to display data from these providers from more than 90 days in the past.

greinard:
@lynnfaraday - Any further concerns about this PR?  It doesn't want me to proceed with the changes requested marker.

> I'm not sure how to check if all of the device type names are correct, so it would be good to have someone more familiar with device data (Vlad maybe?) glance through as well.

For Apple Health single-type data types, which these all are, the V2 type is the same as the query rule name.  These can be found: https://designer.mydatahelps.dev/MiscAdmins/ServiceConfigurationEditor#/HealthKit/HealthKit

```
...
<QueryRule xsi:type="Quantity" Name="Resting Heart Rate"> ...
<QueryRule xsi:type="Quantity" Name="Blood Glucose"> ...
<QueryRule xsi:type="Category" Name="Sleep Analysis"> ...
<QueryRule xsi:type="Category" Name="Mindful Sessions"> ...
<QueryRule xsi:type="StatisticsCollection" Name="Hourly Steps"> ...
<QueryRule xsi:type="StatisticsCollection" Name="Hourly Maximum Heart Rate"> ...
<QueryRule xsi:type="StatisticsCollection" Name="Hourly Minimum Heart Rate"> ...
...
```

For Health Connect, you can look at the real time device data queries: https://github.com/CareEvolution/Consumers/blob/master/Mogul/Code/CfhrFramework/Framework/HealthConnectQueries.cs#L1128
- For `blood-glucose`: https://github.com/CareEvolution/Consumers/blob/master/Mogul/Code/CfhrFramework/Framework/HealthConnectQueries.cs#L1179

lynnfaraday (approved):

greinard:
Thanks @lynnfaraday.

PR #645, opened 2026-07-29
Meal Editor Logging

lynnfaraday:
## Overview

Add custom event logging to the meal editor for troubleshooting image uploads.

## Security

## Testing

- [X] This change requires additional testing in the MDH iOS/Android/Web apps. (Create a pre-release tag/build and test in a ViewBuilder PR.)

Test meal editor in web/mobile and ensure custom events are logged.

### Documentation

- [X] This change does not impact documentation or Storybook.

## Reviewers

- Subject-matter experts
- Style/editing reviewers

### DIFF
--- a/src/components/container/MealEditor/MealEditor.tsx
+++ b/src/components/container/MealEditor/MealEditor.tsx
@@ -1,6 +1,6 @@
 import React, { CSSProperties, useContext, useEffect, useRef, useState } from 'react';
 import './MealEditor.css';
-import { getMealImageUrls, getMeals, getMealToEdit, getMealTypeDisplayText, language, Meal, resolveColor, saveMeals, timestampSortAsc, uploadMealImageFile } from '../../../helpers';
+import { getMealImageUrls, getMeals, getMealToEdit, getMealTypeDisplayText, language, Meal, resolveColor, saveMeals, timestampSortAsc, uploadMealImageFile, logMealEvent } from '../../../helpers';
 import { Button, Card, LayoutContext, LoadingIndicator, UnstyledButton } from '../../presentational';
 import { format, parse, startOfDay } from 'date-fns';
 import { createPreviewData, MealEditorPreviewState } from './MealEditor.previewData';
@@ -9,6 +9,7 @@ import { faEdit, faPlus, faTrashCan } from '@fortawesome/free-solid-svg-icons';
 import { getFullDayAndDateString } from '../../../helpers/date-helpers';
 import { combineItemsWithAnalysisItems, itemSortByNameAsc } from '../../../helpers/glucose-and-meals/meals';
 import MealAnalysis from '../../presentational/MealAnalysis';
+import MyDataHelps, { DeviceInfo } from "@careevolution/mydatahelps-js";
 
 export interface MealEditorProps {
     previewState?: 'loading' | MealEditorPreviewState;
@@ -36,8 +37,14 @@ export default function MealEditor(props: MealEditorProps) {
     const [newImageFile, setNewImageFile] = useState<File>();
     const [imageTypeError, setImageTypeError] = useState<boolean>(false);
     const [imageUploadError, setImageUploadError] = useState<boolean>(false);
+    const [deviceInfo, setDeviceInfo] = useState<DeviceInfo>();
 
     const itemsToAddInputRef = useRef<HTMLInputElement>(null);
+
+    const reportError = (message: string) => {
+        logMealEvent("meal-error", deviceInfo, message);
+        props.onError();
+    }
 
     useEffect(() => {
         setLoading(true);
@@ -81,12 +88,16 @@ export default function MealEditor(props: MealEditorProps) {
                         setImageUrl(imageUrl);
                         setImageLoading(!!imageUrl);
                     } else {
-                        props.onError();
+                        reportError("Can't find meal reference.");
                     }
                 });
             } else {
-                props.onError();
+                reportError("No meal reference provided.");
             }
+        });
+
+        MyDataHelps.getDeviceInfo().then(info => {
+            setDeviceInfo(info);
         });
     }, [props.previewState]);
 
@@ -123,9 +134,11 @@ export default function MealEditor(props: MealEditorProps) {
         setImageUploadError(false);
 
         if (newImageFile) {
+            logMealEvent("uploading-image", deviceInfo);
             try {
                 await uploadMealImageFile(mealToEdit, newImageFile);
-            } catch {
+            } catch (err) {
+                logMealEvent("image-upload-error", deviceInfo, err);
                 setLoading(false);
                 setImageUploadError(true);
                 return;
@@ -142,6 +155,8 @@ export default function MealEditor(props: MealEditorProps) {
 
         const otherMeals = allMeals.filter(meal => meal.id !== mealToEdit.id);
         const updatedMeals = [...otherMeals, mealToEdit].sort(timestampSortAsc);
+
+        logMealEvent("saving-meals", deviceInfo);
 
         saveMeals(startOfDay(mealToEdit.timestamp), updatedMeals).then(props.onSave);
     };
@@ -170,7 +185,9 @@ export default function MealEditor(props: MealEditorProps) {
                 setImageUrl(URL.createObjectURL(file));
                 setImageTypeError(false);
                 setImageUploadError(false);
+                logMealEvent("image-added", deviceInfo, `Image type=${file.type} size=${file.size}`);
             } else {
+                logMealEvent("image-type-error", deviceInfo, `Invalid image type ${file.type}`);
                 setImageTypeError(true);
             }
         }
--- a/src/components/container/MealEditor/MealEditor.tsx
+++ b/src/components/container/MealEditor/MealEditor.tsx
@@ -41,8 +41,8 @@ export default function MealEditor(props: MealEditorProps) {
 
     const itemsToAddInputRef = useRef<HTMLInputElement>(null);
 
-    const reportError = (message: string) => {
-        logMealEvent("meal-error", deviceInfo, message);
+    const onError = (message: string) => {
+        logMealEvent("meal-error", mealToEdit, deviceInfo, message);
         props.onError();
     }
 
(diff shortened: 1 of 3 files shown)


src/components/container/MealEditor/MealEditor.tsx
```diff
@@ -36,9 +37,15 @@ export default function MealEditor(props: MealEditorProps) {
     const [newImageFile, setNewImageFile] = useState<File>();
     const [imageTypeError, setImageTypeError] = useState<boolean>(false);
     const [imageUploadError, setImageUploadError] = useState<boolean>(false);
+    const [deviceInfo, setDeviceInfo] = useState<DeviceInfo>();
 
     const itemsToAddInputRef = useRef<HTMLInputElement>(null);
 
+    const reportError = (message: string) => {
```
greinard: Maybe rename this to just `onError`, sort of like an override for `props.onError` that still calls the super after logging.

src/components/container/MealEditor/MealEditor.tsx
```diff
@@ -81,13 +88,17 @@ export default function MealEditor(props: MealEditorProps) {
                         setImageUrl(imageUrl);
                         setImageLoading(!!imageUrl);
                     } else {
-                        props.onError();
+                        reportError("Can't find meal reference.");
```
greinard: Maybe: `"Unable to find referenced meal: <meal_id>"`

src/components/container/MealEditor/MealEditor.tsx
```diff
@@ -143,6 +156,8 @@ export default function MealEditor(props: MealEditorProps) {
         const otherMeals = allMeals.filter(meal => meal.id !== mealToEdit.id);
         const updatedMeals = [...otherMeals, mealToEdit].sort(timestampSortAsc);
 
+        logMealEvent("saving-meals", deviceInfo);
```
greinard: I would maybe move this up to just after the preview state shortcut at the top.

src/components/container/MealEditor/MealEditor.tsx
```diff
@@ -123,9 +134,11 @@ export default function MealEditor(props: MealEditorProps) {
         setImageUploadError(false);
 
         if (newImageFile) {
+            logMealEvent("uploading-image", deviceInfo);
             try {
                 await uploadMealImageFile(mealToEdit, newImageFile);
```
greinard, line 143: Maybe log an `image-uploaded` event upon success.

src/components/container/MealEditor/MealEditor.tsx
```diff
@@ -123,9 +134,11 @@ export default function MealEditor(props: MealEditorProps) {
         setImageUploadError(false);
 
         if (newImageFile) {
+            logMealEvent("uploading-image", deviceInfo);
```
greinard: It is probably worth including the meal id in most of these log messages.  Might also be worth including the file name for the upload logs.

src/helpers/glucose-and-meals/meals.ts
```diff
@@ -135,4 +135,17 @@ export function itemSortByNameAsc(item1: MealItem, item2: MealItem): number {
 
 export function itemSortByConfidenceDesc(item1: MealItem, item2: MealItem): number {
     return (item2.confidenceScore ?? 0) - (item1.confidenceScore ?? 0);
+}
+
+export function logMealEvent(type: string, deviceInfo: DeviceInfo | undefined, data: any = {}) {
+    console.log(`Meal Editor: ${type} ${data}`);
```
greinard: You can remove this `console.log`.  It won't be helpful on the client, and a dev can just view the network request in Dev Tools to see that this is happening.

src/helpers/glucose-and-meals/meals.ts
```diff
@@ -135,4 +135,17 @@ export function itemSortByNameAsc(item1: MealItem, item2: MealItem): number {
...
+    MyDataHelps.trackCustomEvent({
+        eventType: "meal-editor",
+        properties: {
+            type: type,
+            body: {
+                errorData: data,
+                deviceInfo: deviceInfo
```
greinard: I'm a tad hesitant about logging the full device info with each of these.  I think it would be better to pull individual fields if we think they are important, like platform, model, or OS version.

I also think we already record the device information on the participant, so it may be enough to just log an identifier so we can link back to the device info we already have.

I think the device info can sometimes include properties like the person's name (e.g. "Bob's iPhone", etc.), so while it may be something we are okay with having in the logs, I usually like to err on the side of caution and just log the minimum we need to be able to link back to the other stored data.

greinard (changes requested):
It may be worth instrumenting a few of the other actions as well, such as on delete, on cancel, on remove image, etc.

You also probably want to target a branch off the `v3.21.2` tag so this can be deployed to the Prod ViewBuilder more quickly.  The `v4.0.0` stuff is going to take a little longer to test/validate for release to VB, so if this is semi-urgent, having it be part of a minor bump will help expedite its rollout.

src/components/container/MealEditor/MealEditor.tsx
lynnfaraday: I wanted it here as it's the final checkpoint before calling out to saveMeals. But I can add an extra one at the top where you suggested.

src/components/container/MealEditor/MealEditor.tsx
```diff
@@ -81,13 +88,17 @@ export default function MealEditor(props: MealEditorProps) {
                         setImageUrl(imageUrl);
                         setImageLoading(!!imageUrl);
                     } else {
-                        props.onError();
+                        onError(`Can't find meal reference. id=${mealReference.id}`);
```
greinard: Nitpick, but the thing we cant find here is the meal itself, not the reference.  So I'd use just "Can't find meal." or "Can't find referenced meal."

src/components/container/MealEditor/MealEditor.tsx
```diff
@@ -81,13 +88,17 @@ export default function MealEditor(props: MealEditorProps) {
...
                     }
                 });
             } else {
-                props.onError();
+                onError("No meal reference provided.");
             }
         });
+
+        MyDataHelps.getDeviceInfo().then(info => {
```
greinard: Since we would prefer to have this device info available for any of the logging, including when it loads the meal to edit, we can probably move the whole `getMealToEdit().then(...)` piece inside here after calling `setDeviceInfo(info);`.

The `getDeviceInfo()` call should be really fast as I believe it just runs in memory on the device or in the browser (no remote requests).

src/components/container/MealEditor/MealEditor.tsx
```diff
@@ -118,14 +131,19 @@ export default function MealEditor(props: MealEditorProps) {
...
         setLoading(true);
         setImageTypeError(false);
         setImageUploadError(false);
 
         if (newImageFile) {
+            logMealEvent("uploading-image", mealToEdit, deviceInfo, { name: newImageFile.name, size: newImageFile.size });
             try {
                 await uploadMealImageFile(mealToEdit, newImageFile);
-            } catch {
+                logMealEvent("image-uploaded", mealToEdit, deviceInfo);
```
greinard: Maybe include the `{ name: newImageFile.name, size: newImageFile.size }` here as well.

src/components/container/MealEditor/MealEditor.tsx
greinard, line 127: Probably worth adding an `onCancel` function in here that logs an event and calls `props.onCancel()`, and then update the cancel button at the bottom to call this `onCancel` instead of the prop function directly.

src/helpers/glucose-and-meals/meals.ts
```diff
@@ -135,4 +135,17 @@ export function itemSortByNameAsc(item1: MealItem, item2: MealItem): number {
 
 export function itemSortByConfidenceDesc(item1: MealItem, item2: MealItem): number {
     return (item2.confidenceScore ?? 0) - (item1.confidenceScore ?? 0);
+}
+
+export function logMealEvent(type: string, meal: Meal | undefined, deviceInfo: DeviceInfo | undefined, data: any = {}) {
```
greinard: Maybe use `unknown` for the type on `data` instead of `any`.

src/helpers/glucose-and-meals/meals.ts
```diff
@@ -135,4 +135,17 @@ export function itemSortByNameAsc(item1: MealItem, item2: MealItem): number {
...
+    MyDataHelps.trackCustomEvent({
+        eventType: "meal-editor",
+        properties: {
+            type: type,
+            body: data,
```
greinard: I like this better with the individual fields we are interested in.  Ties it back to the meal and the device info nicely without logging TMI.

Was thinking maybe this field could be named just `data` or `additionalData` rather than `body`, but I don't feel strongly about it if you prefer `body`.

greinard (changes requested):
Looking pretty good.  Just a few more suggestions.  Thanks for addressing the prior feedback.

src/components/container/MealEditor/MealEditor.tsx
greinard: Ah, yeah, that makes sense.

src/helpers/glucose-and-meals/meals.ts
```diff
@@ -135,4 +135,17 @@ export function itemSortByNameAsc(item1: MealItem, item2: MealItem): number {
 
 export function itemSortByConfidenceDesc(item1: MealItem, item2: MealItem): number {
     return (item2.confidenceScore ?? 0) - (item1.confidenceScore ?? 0);
+}
+
+export function logMealEvent(type: string, meal: Meal | undefined, deviceInfo: DeviceInfo | undefined, data: any = {}) {
+    MyDataHelps.trackCustomEvent({
+        eventType: "meal-editor",
```
greinard: Oh, one more thing.  I suppose at some point we might be able to reuse this function across the other meal related components, so maybe we just set the event type to `meal-event` here (kind of fits the function name better, too).

lynnfaraday:
Replaced by https://github.com/CareEvolution/MyDataHelpsUI/pull/648 due to rebase having a fit.

PR #646, opened 2026-07-29
Storybook: single dark/light theme toggle

n400:
## Summary
Replaces the three-item Theme dropdown and the backgrounds-addon color picker with a single
☀/☾ toolbar button that switches the whole preview — story canvas *and* Docs pages — between
light and dark.

- `.storybook/manager.js` (new) — the toolbar toggle button.
- `.storybook/preview.js` — `theme` global + a decorator that emulates `prefers-color-scheme`
  by patching `window.matchMedia` (components resolve `colorScheme="auto"` from it), paints the
  canvas via the semantic `--mdhui-background-color-1` token, and themes Docs pages through a
  custom `DocsContainer` fed by the decorator itself (which receives `context.globals` through
  the documented API) — no Storybook channel/event internals involved.
- `.storybook/main.js` — disables the essentials backgrounds addon (its canvas-color picker
  duplicated/conflicted with the theme button).
- `package.json` — `@storybook/blocks` and `@storybook/theming` promoted from transitive to
  explicit devDependencies (they are now imported directly).

## Caveats / things to know
- **Theme state is binary by design to simplify the UI.** "System" (follow the OS setting) applies only until the
  first click; after that the pinned light/dark choice sticks and persists in the URL
  (`&globals=theme:...`), so a shared link carries the pin. A URL without that param follows
  the OS again. There is deliberately no third "System" button state.
- **The `matchMedia` stub does not emit change events.** Components in this repo read the
  scheme at render (and the toggle re-renders every story), so nothing is affected — but a
  future component that *subscribes* via `matchMedia(...).addEventListener('change', …)` would
  not hear toolbar toggles.
- **Disabling backgrounds also removes the grid-overlay tool** (it ships inside the same
  addon). No story used `parameters.backgrounds`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


n400:
Superseded by #657, which now includes this commit and targets main directly.

